PR #10, opened 2025-07-18
Update region in memory client

sundargthb:
*Issue #, if available:*

*Description of changes:*

Updates MemoryClient initialization to follow standard AWS SDK region detection 
precedence (explicit param → env var → botocore session → default). Previously 
hardcoded region defaults and endpoints are now dynamic based on the detected region.

### DIFF
--- a/src/bedrock_agentcore/memory/client.py
+++ b/src/bedrock_agentcore/memory/client.py
@@ -41,11 +41,18 @@ class MemoryClient:
 
     def __init__(self, region_name: Optional[str] = None, environment: str = "prod"):
         """Initialize the Memory client."""
-
-        self.region_name = region_name or os.getenv("AWS_REGION") or boto3.Session().region_name or "us-west-2"
+        env_region = os.getenv("AWS_REGION")
+        if region_name is not None and env_region is not None and env_region != region_name:
+            warnings.warn(
+                f"AWS_REGION environment variable ({env_region}) differs from provided "
+                f"region_name ({region_name}). Using provided region_name.",
+                stacklevel=2,
+            )
+            self.region_name = region_name
+        else:
+            self.region_name = region_name or env_region or boto3.Session().region_name or "us-west-2"
         self.environment = environment
 
-        # Use region_name in endpoint URLs
         self.control_plane_endpoint = os.getenv(
             "AGENTCORE_CONTROL_ENDPOINT", f"https://bedrock-agentcore-control.{self.region_name}.amazonaws.com"
         )
--- a/src/bedrock_agentcore/memory/client.py
+++ b/src/bedrock_agentcore/memory/client.py
@@ -39,24 +39,18 @@ logger = logging.getLogger(__name__)
 class MemoryClient:
     """High-level Bedrock AgentCore Memory client with essential operations."""
 
-    def __init__(self, region_name: str = "us-west-2", environment: str = "prod"):
+    def __init__(self, region_name: Optional[str] = None, environment: str = "prod"):
         """Initialize the Memory client."""
-        self.region_name = region_name
-        if os.getenv("AWS_REGION"):
-            env_region = os.getenv("AWS_REGION")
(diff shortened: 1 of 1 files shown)


src/bedrock_agentcore/memory/client.py
```diff
@@ -39,24 +39,25 @@
 class MemoryClient:
     """High-level Bedrock AgentCore Memory client with essential operations."""
```
siwachabhi, line 40: we should be updating this where default region is defined near --region flag, so for all primitives, region derivation logic is consistent

src/bedrock_agentcore/memory/client.py
```diff
@@ -39,24 +39,25 @@
...
 
-    def __init__(self, region_name: str = "us-west-2", environment: str = "prod"):
+    def __init__(self, region_name: Optional[str] = None, environment: str = "prod"):
         """Initialize the Memory client."""
```
siwachabhi, line 42: We should be removing environment option

src/bedrock_agentcore/memory/client.py
```diff
@@ -39,24 +39,25 @@
...
+            self.region_name = region_name or env_region or boto3.Session().region_name or "us-west-2"
         self.environment = environment
 
         self.control_plane_endpoint = os.getenv(
-            "AGENTCORE_CONTROL_ENDPOINT", "https://bedrock-agentcore-control.us-west-2.amazonaws.com"
+            "AGENTCORE_CONTROL_ENDPOINT", f"https://bedrock-agentcore-control.{self.region_name}.amazonaws.com"
         )
         self.data_plane_endpoint = os.getenv(
-            "AGENTCORE_DATA_ENDPOINT", "https://bedrock-agentcore.us-west-2.amazonaws.com"
+            "AGENTCORE_DATA_ENDPOINT", f"https://bedrock-agentcore.{self.region_name}.amazonaws.com"
```
bergjaak, line 60: We should be able to delete all of this now that the AWS SDK is released (and so contains these endpoints)

src/bedrock_agentcore/memory/client.py
```diff
@@ -39,24 +39,25 @@
...
-        if os.getenv("AWS_REGION"):
-            env_region = os.getenv("AWS_REGION")
-            if env_region != region_name:
-                warnings.warn(
-                    f"AWS_REGION environment variable ({env_region}) differs from provided "
-                    f"region_name ({region_name}). Using provided region_name.",
-                    stacklevel=2,
-                )
+        env_region = os.getenv("AWS_REGION")
+        if region_name is not None and env_region is not None and env_region != region_name:
```
bergjaak: we should be able to remove all of this reading of the environment variable. And if the user does not pass the region argument to the __init__, then it should be possible for the boto client to use the default region or read the environment variable (but our code should not be reading the environment variable; we should let boto3 handle that)

src/bedrock_agentcore/memory/client.py
sundargthb: removed the environment option

src/bedrock_agentcore/memory/client.py
sundargthb: updated the region derivation logic in MemoryClient to ensure consistency across all primitives. Now uses the explicitly provided region_name when available, falling back to boto3.Session().region_name, and finally defaulting to 'us-west-2' if needed.

src/bedrock_agentcore/memory/client.py
sundargthb: I've removed the direct reading of environment variables for region configuration as suggested. The code now relies on boto3's built-in region resolution via boto3.Session().region_name

bergjaak (approved):

bergjaak (approved):

siwachabhi (approved):

bergjaak (approved):

PR #36, opened 2025-07-31
add automated release workflows (PyPI disabled for testing)

sundargthb:
*Issue #, if available:*

*Description of changes:*
add automated release workflows (PyPI disabled for testing)


.github/workflows/pr-automerge.yml
```diff
@@ -0,0 +1,80 @@
...
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  pull_request_review:
+    types: [submitted]
+  check_suite:
+    types: [completed]
+  workflow_run:
+    workflows: ["CI"]
+    types: [completed]
```
guoken1: not sure about this but I think this can result in the action running multiple times when doing a release

.github/workflows/pr-automerge.yml
```diff
@@ -0,0 +1,80 @@
...
+  pull-requests: write
+
+jobs:
+  auto-merge:
+    name: Auto-merge Release PRs
+    runs-on: ubuntu-latest
+    if: |
+      github.event.pull_request.user.login == 'github-actions[bot]' &&
+      startsWith(github.event.pull_request.head.ref, 'release/') &&
+      github.event.pull_request.base.ref == 'main'
```
guoken1, line 20: these checks may produce errors when the pull_request context is not available

.github/workflows/pr-automerge.yml
```diff
@@ -0,0 +1,80 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Check PR approval status
```
guoken1: this step is checking whether the pr is approved, so we probably don't need to perform this check immediately when the pr is opened

.github/workflows/pr-automerge.yml
```diff
@@ -0,0 +1,80 @@
...
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { data: checkRuns } = await github.rest.checks.listForRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: context.payload.pull_request.head.sha
+            });
+
+            const requiredChecks = ['Lint', 'Test Python 3.10'];
```
guoken1: I think we should include all our checks here

scripts/bump_version.py
```diff
@@ -0,0 +1,172 @@
...
+import sys
+from datetime import datetime
+from pathlib import Path
+from typing import Optional, Tuple
+
+
+def get_current_version() -> str:
+    """Get current version from pyproject.toml."""
+    content = Path("pyproject.toml").read_text()
+    match = re.search(r'version = "([^"]+)"', content)
```
guoken1: consider making this regex more robust

.github/workflows/release.yml
```diff
@@ -1,122 +1,212 @@
...
+          if [ "${{ github.event.inputs.dry_run }}" != "true" ]; then
+            echo "=== ACTUAL VERSION BUMP ==="
+            if [ -n "${{ github.event.inputs.changelog }}" ]; then
+              python scripts/bump_version.py ${{ github.event.inputs.bump_type }} \
+                --changelog "${{ github.event.inputs.changelog }}"
+            else
+              python scripts/bump_version.py ${{ github.event.inputs.bump_type }}
+            fi
+
+            uv lock --no-progress || echo "Note: uv.lock update failed"
```
guoken1: should this fail the workflow if it fails?

.github/workflows/release.yml
```diff
@@ -1,122 +1,212 @@
...
+            echo "- Previous version: ${{ steps.current.outputs.version }}" >> $GITHUB_STEP_SUMMARY
+            echo "- New version: ${{ steps.bump.outputs.version }}" >> $GITHUB_STEP_SUMMARY
+            echo "- Branch: release/v${{ steps.bump.outputs.version }}" >> $GITHUB_STEP_SUMMARY
+            echo "" >> $GITHUB_STEP_SUMMARY
+            echo "**Note:** PyPI publishing is currently disabled for testing" >> $GITHUB_STEP_SUMMARY
+          fi
+
+
+  test-and-build:
+    name: Test and Build
```
guoken1, line 151: I don't think there are tests in this job

scripts/bump_version.py
```diff
@@ -0,0 +1,172 @@
...
+    pyproject = Path("pyproject.toml")
+    content = pyproject.read_text()
+    content = re.sub(f'version = "{re.escape(old_version)}"', f'version = "{new_version}"', content)
+    pyproject.write_text(content)
+    print("✓ Updated pyproject.toml")
+
+    # Update __init__.py files
+    for init_file in Path("src").rglob("__init__.py"):
+        if update_version_in_file(init_file, old_version, new_version):
+            print(f"✓ Updated {init_file}")
```
guoken1: which files would require this update?

scripts/bump_version.py
```diff
@@ -0,0 +1,172 @@
...
+        return True
+    return False
+
+
+def update_all_versions(old_version: str, new_version: str):
+    """Update version in all relevant files."""
+    # Update pyproject.toml
+    pyproject = Path("pyproject.toml")
+    content = pyproject.read_text()
+    content = re.sub(f'version = "{re.escape(old_version)}"', f'version = "{new_version}"', content)
```
guoken1: be careful about making unintended updates here

scripts/bump_version.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # Generate entry
+    date = datetime.now().strftime("%Y-%m-%d")
+    entry = f"\n## [{new_version}] - {date}\n\n"
+
+    if changes:
+        entry += changes + "\n"
+    else:
+        # Auto-generate from git log
+        git_log = get_git_log()
```
guoken1, line 206: I'm not sure if our git logs are good enough to be dropped straight into the changelog

.github/workflows/pr-automerge.yml
```diff
@@ -0,0 +1,66 @@
...
+        with:
+          script: |
+            const { data: checkRuns } = await github.rest.checks.listForRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: context.payload.pull_request.head.sha
+            });
+
+            // Include ALL required checks
+            const requiredChecks = [
```
guoken1, line 35: consider adding integration tests too

.github/workflows/release.yml
```diff
@@ -76,47 +224,92 @@ jobs:
...
+    steps:
+      - name: Approval checkpoint
+        run: |
+          echo "✅ TestPyPI deployment successful"
+          echo "📦 Package available at: https://test.pypi.org/project/bedrock-agentcore/"
+          echo ""
+          echo "⚠️  MANUAL APPROVAL REQUIRED FOR PRODUCTION"
+          echo ""
+          echo "Before approving:"
+          echo "1. Test package: pip install -i https://test.pypi.org/simple/ bedrock-agentcore"
```
guoken1, line 245: extra space

scripts/bump_version.py
```diff
@@ -0,0 +1,259 @@
...
+from pathlib import Path
+from typing import Optional, Tuple
+
+
+def get_current_version() -> str:
+    """Get current version from pyproject.toml."""
+    content = Path("pyproject.toml").read_text()
+    # More robust regex that matches the project version specifically
+    # Look for version under [project] or [tool.poetry] sections
+    pattern = r'(?:^\[project\]|\[tool\.poetry\])[\s\S]*?^version\s*=\s*"([^"]+)"'
```
guoken1, line 14: we are not using poetry so we can remove it

guoken1 (approved):

.github/workflows/pr-automerge.yml
sundargthb: Updated to only trigger on pull_request_review when approved, preventing multiple runs

.github/workflows/pr-automerge.yml
sundargthb: The workflow now only runs when review is approved, ensuring PR context is available

.github/workflows/pr-automerge.yml
sundargthb: Since we're only checking PR status, no checkout needed

.github/workflows/pr-automerge.yml
sundargthb: now includes all Python version tests and build checks

scripts/bump_version.py
sundargthb: thanks.. Version regex now specifically targets project section to avoid false matches

.github/workflows/release.yml
sundargthb: updated to fail the workflow if lock update fails

scripts/bump_version.py
sundargthb: Currently only src/bedrock_agentcore/__init__.py contains version

scripts/bump_version.py
sundargthb: Added commit categorization and formatting for cleaner changelogs

The updated files ensure more reliable releases with better error handling and cleaner output.

scripts/bump_version.py
sundargthb: Version replacement now only occurs in the [project] section

PR #46, opened 2025-08-07
chore: remove concurrency checks and simplify thread pool handling

siwachabhi:
Tested using an agent like:

```
import asyncio
import time
from strands import Agent

from bedrock_agentcore import BedrockAgentCoreApp

app = BedrockAgentCoreApp()
agent = Agent()


@app.entrypoint
async def agent_invocation(payload):
    """Handler for agent invocation"""
    user_message = payload.get(
        "prompt", "No way prompt found in input, please guide customer to create a json payload with prompt key"
    )
    
    # Wait for 60 seconds before starting the stream (blocks the thread)
    print("Waiting for 45 seconds before starting stream...")
    time.sleep(45)
    print("45 second wait complete. Starting stream...")
    
    stream = agent.stream_async(user_message)
    async for event in stream:
        print(event)
        yield (event)


if __name__ == "__main__":
    app.run()

```

which was earlier failing due to ping failures, but now with platform changes, separate thread pool isn't required


Vivekbhadauria1 (approved):

Abishek10 (approved):
LGTM !

PR #61, opened 2025-08-27
feat(memory): add functionality to memory client

akshseh:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/bedrock_agentcore/memory/client.py
+++ b/src/bedrock_agentcore/memory/client.py
@@ -115,19 +115,15 @@ class MemoryClient:
             )
             return memory
         except ClientError as e:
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/memory/client.py
```diff
@@ -92,6 +92,47 @@ def create_memory(
             logger.error("Failed to create memory: %s", e)
             raise
 
+    def create_or_get_memory_id(
```
jona62: I think we should return the entire memory object here, we can always extract the id from the returned memory object and rename the function appropriately to `get_or_create_memory`

siwachabhi (approved):

PR #66, opened 2025-09-05
feat(memory): Add passthrough for gmdp and gmcp operations for Memory

jgordley:
Many of the LangChain/Crew/Llamaindex integrations require individual event management and listing of memory records in addition to just searching.

## Features
  - Added `__get_attr__` method to the Memory Client to dynamically forward the request to either the data plane or the control plane depending on the operation
  - Added a filter for methods that are supported on the memory client

## Testing
Added unit tests to cover GetEvent, DeleteEvent, ListMemoryRecords, and a few other operations.

### DIFF
--- a/src/bedrock_agentcore/memory/client.py
+++ b/src/bedrock_agentcore/memory/client.py
@@ -435,6 +435,236 @@ class MemoryClient:
             logger.error("Failed to create blob event: %s", e)
             raise
 
+    def get_event(
+        self,
+        memory_id: str,
+        actor_id: str,
+        session_id: str,
+        event_id: str,
+    ) -> Dict[str, Any]:
+        """Retrieves information about a specific event in an AgentCore Memory resource.
+
+        Args:
+            memory_id: Memory resource ID
+            actor_id: Actor identifier
+            session_id: Session identifier
+            event_id: Event identifier to retrieve
+
+        Returns:
+            Dictionary containing the event information
+
+        Example:
+            event = client.get_event(
+                memory_id="mem-xyz",
+                actor_id="user-123",
+                session_id="session-456",
+                event_id="123#abc123"
+            )
+        """
+        try:
+            response = self.gmdp_client.get_event(
+                memoryId=memory_id,
+                actorId=actor_id,
+                sessionId=session_id,
+                eventId=event_id,
+            )
+
+            logger.info("Retrieved event: %s", event_id)
+            return response["event"]
+
+        except ClientError as e:
+            logger.error("Failed to get event: %s", e)
+            raise
+
+    def delete_event(
+        self,
+        memory_id: str,
+        actor_id: str,
+        session_id: str,
+        event_id: str,
+    ) -> None:
+        """Deletes an event from an AgentCore Memory resource.
+
+        When you delete an event, it is permanently removed.
+
+        Args:
+            memory_id: Memory resource ID
+            actor_id: Actor identifier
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/memory/client.py
```diff
@@ -435,6 +435,236 @@ def create_blob_event(
             logger.error("Failed to create blob event: %s", e)
             raise
 
+    def get_event(
```
kandoin: Instead of adding these pass-through methods, a better way will be to directly have the attributes copied from the gmdp and gmcp_client into this client and definining the set of apis you want to import here. Something along these lines will give you all the methods. 

You can then define a filter for the one you'd like to import in this client so that you aren't copying over methods for runtime/gateway etc. This will allow you to one shot import all the methods and not have this translation logic.

```
+    def __getattr__(self, name: str):
+        """Dynamically forward method calls to the appropriate boto3 client.
+
+        This method enables access to all boto3 client methods without explicitly
+        defining them. Methods are looked up in the following order:
+        1. gmdp_client (bedrock-agentcore) - for data plane operations
+        2. gmcp_client (bedrock-agentcore-control) - for control plane operations
+
+        Args:
+            name: The method name being accessed
+
+        Returns:
+            A callable method from the appropriate boto3 client
+
+        Raises:
+            AttributeError: If the method doesn't exist on either client
+
+        Example:
+            # Access any boto3 method directly
+            client = MemoryClient()
+
+            # These calls are forwarded to the appropriate boto3 client
+            response = client.list_memory_records(memoryId="mem-123", namespace="test")
+            metadata = client.get_memory_metadata(memoryId="mem-123")
+        """
+        # First try gmdp_client (data plane) as it's more commonly used
+        if hasattr(self.gmdp_client, name):
+            method = getattr(self.gmdp_client, name)
+            logger.debug("Forwarding method '%s' to gmdp_client", name)
+            return method
+
+        # Then try gmcp_client (control plane)
+        if hasattr(self.gmcp_client, name):
+            method = getattr(self.gmcp_client, name)
+            logger.debug("Forwarding method '%s' to gmcp_client", name)
+            return method
+
+        # Method not found on either client
+            f"Method not found on gmdp_client or gmcp_client. "
+            f"'bedrock-agentcore' and 'bedrock-agentcore-control' services."
+        )
```

jgordley: This is a good point. Although we do lose some of the advantages of doing things like pagination under the hood for list event methods and extracting the response through returning `return response['event']`.

Although we can't just delete the old methods since users may have already implemented agents using those methods that now require a new `.get('event')` call to get the object that they previously would have needed, I agree with implementing this for all the new methods going forward that are not covered. I'll update the implementation.

src/bedrock_agentcore/memory/client.py
```diff
@@ -38,6 +38,28 @@
...
+        "delete_memory_record",
+        "list_memory_records",
+        "create_event",
+        "get_event",
+        "delete_event",
+        "list_events",
+    }
+
+    # AgentCore Memory control plane methods
+    _ALLOWED_GMCP_METHODS = {
```
bergjaak, line 54: would it make sense to put these directly in the __getattr__ function so that the customer would have no confusion that these might be used somewhere else? eg I could see a customer interpreting this as meaning that they are not allowed to call other endpoints

jona62:
LGTM!

PR #87, opened 2025-09-18
feat: Initial commit for Session Manager, Session and Actor constructs

jona62:
# feat: Initial commit for Session Manager, Session and Actor constructs

## Overview

This PR introduces a comprehensive session management system for the Bedrock AgentCore SDK, providing high-level abstractions for managing conversational AI agents with memory capabilities. The implementation includes three core constructs: `MemorySessionManager`, `MemorySession`, and `Actor`, along with supporting data models and example notebooks.

### 🏗️ Core Architecture

- **MemorySessionManager**: Central orchestrator for memory operations with automatic boto3 client management
- **MemorySession**: Stateful conversation wrapper providing scoped operations for a specific actor-session pair  
- **Actor**: Actor-scoped operations for managing multiple sessions per user

### 🧠 Memory Management Capabilities

- **Short-term Memory**: Event-based conversation storage with branching support
- **Long-term Memory**: Semantic search and retrieval of persistent knowledge
- **Conversation Branching**: Fork conversations from any point to explore alternative paths
- **Turn Management**: Intelligent grouping of messages into conversational turns

### 🔧 Developer Experience

- **LLM Integration**: Built-in callback pattern for seamless LLM integration with memory context
- **Automatic Pagination**: Handles large datasets transparently
- **Type Safety**: Comprehensive type hints and structured data models
- **Error Handling**: Robust error handling with detailed logging

### Core Implementation
- `src/bedrock_agentcore/memory/session.py` - Main session management classes (900+ lines)
- `src/bedrock_agentcore/memory/model/__init__.py` - data models
- `src/bedrock_agentcore/memory/__init__.py` - Updated module exports

### Data Models
- `ActorSummary` - Actor metadata wrapper
- `Branch` - Conversation branch representation
- `BranchContext` - Branch message context
- `ConversationTree` - Hierarchical conversation structure
- `DictWrapper` - Base class for data model wrappers
- `Event` - Event data wrapper
- `EventMessage` - Message within events
- `MemoryRecord` - Long-term memory record
- `SessionSummary` - Session metadata wrapper

### Testing
- `tests/bedrock_agentcore/memory/test_session.py` - Comprehensive unit tests

### Simple Session Creation
```python
manager = MemorySessionManager(memory_id="my-memory", region="us-east-1")
session = manager.create_memory_session(actor_id="user-123")
```

### LLM Integration with Memory
```python
def my_llm(user_input: str, memories: List[Dict]) -> str:
    # Your LLM logic here
    return response

retrieval_config = {
    "support/user/{actorId}/{sessionId}": RetrievalConfig(
        top_k=5,
        relevance_score=0.3
    ),
    "user/preferences/{actorId}": RetrievalConfig(
        top_k=3,
        relevance_score=0.5
    )
}

memories, response, event = session.process_turn_with_llm(
    user_input="What did we discuss yesterday?",
    llm_callback=my_llm,
    retrieval_config=retrieval_config
)
```

### Conversation Branching
```python
# Fork conversation from any point
branch_event = session.fork_conversation(
    root_event_id="event-123",
    branch_name="alternative-path",
    messages=[ConversationalMessage("Let's try a different approach", MessageRole.USER)]
)
```

### Memory Search & Retrieval
```python
# Semantic search in long-term memory
relevant_memories = session.search_long_term_memories(
    query="customer preferences",
    namespace_prefix="customer/profile/{actorId}",
    top_k=5
)
```

### Dynamic Method Forwarding
The `MemorySessionManager` uses `__getattr__` to dynamically forward boto3 client methods, providing direct access to all Bedrock AgentCore data plane operations while maintaining type safety and error handling.

### Intelligent Event Sorting
Custom sorting algorithm handles chronological event ordering using millisecond prefixes from event IDs, ensuring deterministic ordering even with same-second timestamps.

### Conversation Tree Building
Transforms flat event lists into hierarchical tree structures for visualization and analysis, handling complex branching scenarios automatically.

### Robust Pagination
All list operations handle pagination transparently, ensuring complete data retrieval regardless of dataset size.

## Testing Coverage

- Unit tests for all core classes and methods (Coverage report: 95%)
- Mock-based testing to avoid external dependencies
- Error handling validation
- Type checking and parameter validation

## Breaking Changes

None - this is a new feature addition that extends the existing memory module without modifying existing APIs.

## Migration Guide

Existing code using the memory module will continue to work unchanged. New session management features are opt-in and can be adopted incrementally.

## Future Enhancements

This foundation enables future features like:
- Session persistence and restoration
- Advanced branching strategies
- Memory optimization and cleanup
- Real-time collaboration support

### DIFF
--- a/src/bedrock_agentcore/memory/session.py
+++ b/src/bedrock_agentcore/memory/session.py
@@ -12,8 +12,6 @@ from .constants import BlobMessage, ConversationalMessage, MessageRole
 from .models import (
     ActorSummary,
     Branch,
-    BranchContext,
-    ConversationTree,
     DictWrapper,
     Event,
     EventMessage,
@@ -266,7 +264,7 @@ class SessionManager:
         retrieved_memories = []
         if retrieval_namespace:
             search_query = retrieval_query or user_input
-            retrieved_memories = self.search_long_term_memory(
+            retrieved_memories = self.search_long_term_memories(
                 query=search_query, namespace_prefix=retrieval_namespace, top_k=top_k
             )
             logger.info("Retrieved %d memories for LLM context", len(retrieved_memories))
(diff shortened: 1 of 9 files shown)


src/bedrock_agentcore/memory/session.py
```diff
@@ -0,0 +1,1015 @@
...
+
+        # Method not found on client
+        raise AttributeError(
+            f"'{self.__class__.__name__}' object has no attribute '{name}'. "
+            f"Method not found on _data_plane_client. "
+            f"Available methods can be found in the boto3 documentation for "
+            f"'bedrock-agentcore' services."
+        )
+
+    def _event_sort_key(self, event: Dict[str, Any]) -> Tuple[int, str]:
```
bergjaak: what does this do? How does it work since boto3 cant support milliseconds? We have a randum hash in the event id which does guarantee table order, if the idea is just to guarantee stable order

src/bedrock_agentcore/memory/session.py
```diff
@@ -0,0 +1,1015 @@
...
+        Args:
+            event: Event dictionary containing eventId and eventTimestamp
+
+        Returns:
+            Tuple of (milliseconds_since_epoch, suffix) for stable sorting
+        """
+        event_id = event.get("eventId", "")
+
+        # Handle None or empty eventId
+        if not event_id:
```
bergjaak: this should never be the case, right? Like something has gone seriously wrong if this is the case

src/bedrock_agentcore/memory/session.py
```diff
@@ -0,0 +1,1015 @@
...
+            else:
+                ms = 0
+            return (ms, "")
+
+        # eventId format: "<epoch_ms>#<suffix>"
+        head, _, tail = event_id.partition("#")
+
+        try:
+            # Extract milliseconds from eventId prefix
+            ms = int(head)
```
bergjaak: boto3 datetime doesn't have milliseconds

bergjaak: one way you can hack around that is by doing this
```python
import boto3
from datetime import datetime

# Create bedrock-agentcore client
client = boto3.client('bedrock-agentcore')

# Override the timestamp serialization method
original_serialize_timestamp = client._serializer._serializer._serialize_type_timestamp

def serialize_timestamp_with_microseconds(serialized, value, shape, name):
    if isinstance(value, datetime):
        serialized[name] = value.timestamp()  # Float with microseconds
    else:
        original_serialize_timestamp(serialized, value, shape, name)

client._serializer._serializer._serialize_type_timestamp = serialize_timestamp_with_microseconds

# Create event
response = client.create_event(
    memoryId="ToolsTestMemory-Xcj2m46Tsi",
    sessionId="testtest",
    actorId="testtest",
    eventTimestamp=datetime.now(),
    payload=[
        {
            'conversational': {
                'content': {
                    'text': 'Test event'
                },
                'role': 'USER'
            }
        }
    ]
)

# List events
events_response = client.list_events(
    memoryId="ToolsTestMemory-Xcj2m46Tsi",
    sessionId="testtest",
    actorId="testtest"
)
print(events_response)
```

... which is obviously hacky...

bergjaak: but if we _did_ do this, then we could make it stable by adding unit tests that ensure that this is still working. So then minor and major version bumps of boto3 would catch if those broke because of changes to boto3 code

src/bedrock_agentcore/memory/session.py
```diff
@@ -0,0 +1,1015 @@
...
+            actor_id=actor_id,
+            session_id=session_id,
+            messages=[(user_input, Role.USER.name), (agent_response, Role.ASSISTANT.name)],
+            event_timestamp=event_timestamp,
+        )
+
+        logger.info("Completed full conversation turn with LLM")
+        return retrieved_memories, agent_response, event
+
+    def create_blob_event(
```
bergjaak: I am not sure I like the idea of having an API that is specificalyl for blob event. Because a single event can contain a mixture of Blob payload type and Conversational payloadType. So I think this method tricks the user into thinking that they cannot mix blob and payload types.

src/bedrock_agentcore/memory/session.py
```diff
@@ -0,0 +1,1015 @@
...
+            event = response["event"]
+            logger.info("Created blob event: %s", event["eventId"])
+
+            return event
+
+        except ClientError as e:
+            logger.error("Failed to create blob event: %s", e)
+            raise
+
+    def add_turns(
```
bergjaak: I don't like this method either, since it has the same problems as create_blob_event

src/bedrock_agentcore/memory/session.py
```diff
@@ -0,0 +1,1015 @@
...
+        return [BranchContext(message) for message in messages]
+
+    def get_last_k_turns(
+        self,
+        actor_id: str,
+        sesssion_id: str,
+        k: int = 5,
+        branch_name: Optional[str] = None,
+        include_branches: bool = False,
+        max_results: int = 100,
```
bergjaak, line 559: what if max_results < k? Also dont we need to paginate to support branch and other filters that may be added in the future since max_results is a target (not a guarantee)

jona62: If the `max_results < k` then last_k turn is evaluated based on the events fetched with that max_results.
`self.list_events` handles the pagination logic.

src/bedrock_agentcore/memory/session.py
```diff
@@ -0,0 +1,966 @@
...
+
+        # Method not found on client
+        raise AttributeError(
+            f"'{self.__class__.__name__}' object has no attribute '{name}'. "
+            f"Method not found on _data_plane_client. "
+            f"Available methods can be found in the boto3 documentation for "
+            f"'bedrock-agentcore' services."
+        )
+
+    def _event_sort_key(self, event: Dict[str, Any]) -> Tuple[int, str]:
```
bergjaak: let's replace this by making the events accurate at millisecond level

```
 ➜ cat bedrock_agentcore_client.py
import boto3
from datetime import datetime

# Create bedrock-agentcore client
client = boto3.client('bedrock-agentcore')

# Override the timestamp serialization method
original_serialize_timestamp = client._serializer._serializer._serialize_type_timestamp

def serialize_timestamp_with_microseconds(serialized, value, shape, name):
    if isinstance(value, datetime):
        serialized[name] = value.timestamp()  # Float with microseconds
    else:
        original_serialize_timestamp(serialized, value, shape, name)

client._serializer._serializer._serialize_type_timestamp = serialize_timestamp_with_microseconds

```

boto3 is working on a fix, so this workaround is the best we can do for now (and then remove the workaround once it's officially supported by boto3

bergjaak: (it says microseconds but it's actually milliseconds)

jona62: Removed this function, will add this in an upcoming PR

piradeepk (commented):
LGTN

piradeepk (approved):
LGTM

PR #89, opened 2025-09-19
Remove TestPyPI publishing step from release workflow

sundargthb:
# Simplify Release Workflow

## Description
This change modifies the GitHub Actions release workflow to publish directly to PyPI without first publishing to TestPyPI. After some successful releases, we are confident in our release process and can simplify it by removing the TestPyPI step.

## Changes
- Removed TestPyPI publishing step from the GitHub Actions release workflow
- Retained the pull request process for approving release changes (pyproject.toml, uv.lock, changelog.md)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -142,9 +142,8 @@ jobs:
           ### 🔄 Release Process
           After merging this PR:
           1. Package will be built and tested
-          2. Published to Test PyPI automatically
-          3. **Manual approval required** before production PyPI
-          4. GitHub release and tag created after production
+          2. **Manual approval required** before publishing to PyPI
+          3. GitHub release and tag created after PyPI publication
 
           ### 🚨 Running from: ${{ github.ref }}
           ${{ github.ref != 'refs/heads/main' && '**WARNING**: Not running from main branch!' || '✅ Running from main branch' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -289,12 +266,41 @@ jobs:
...
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          # MUST specify password to avoid Trusted Publishing
+          # MUST specify password to avoid Trusted Publishing issues
           password: ${{ secrets.PYPI_API_TOKEN }}
           skip-existing: false
+          verbose: true
+
+      - name: Wait for PyPI availability
```
guoken1, line 291: will this move on to the next step after 5 mins? what happens if the publish fails? would the previous step fail?

guoken1 (approved):

.github/workflows/release.yml
sundargthb: this step was checking for CDN propagation, not whether the publish succeeded. When we publish to PyPI, the package is immediately stored on their servers (and the publish step would fail if this didn’t work), but it takes time to propagate across PyPI’s global CDN network.  

The wait step gives the CDN time to update, but if it’s still not available after 5 minutes, we continue anyway since we know the upload succeeded and the package will eventually be accessible everywhere. The GitHub release and tag should still be created to match what we’ve successfully published.

PR #100, opened 2025-09-30
chore: Add README for MemorySessionManager

jona62:
chore: Add README for MemorySessionManager

### DIFF
--- a/src/bedrock_agentcore/memory/README.md
+++ b/src/bedrock_agentcore/memory/README.md
@@ -1,44 +1,147 @@
 # Bedrock AgentCore Memory SDK
 
-High-level Python SDK for AWS Bedrock AgentCore Memory service with streamlined session management and flexible conversation handling.
+High-level Python SDK for AWS Bedrock AgentCore Memory service with streamlined session management and flexible
+conversation handling.
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Setup](#setup)
+  - [Installation](#installation)
+  - [Authentication](#authentication)
+  - [Environment Variables](#environment-variables)
+- [Recommended Classes](#recommended-classes)
+- [Key Features](#key-features)
+- [Quick Start](#quick-start)
+- [Usage](#usage)
+  - [Enhanced LLM Integration with Memory Context](#enhanced-llm-integration-with-memory-context)
+  - [Natural Conversation Flow](#natural-conversation-flow)
+  - [Branch Management](#branch-management)
+  - [Session and Actor Management](#session-and-actor-management)
+  - [Memory Record Management](#memory-record-management)
+  - [Alternative Pattern: Separated Operations](#alternative-pattern-separated-operations)
+- [Error Handling](#error-handling)
+  - [Common Exceptions](#common-exceptions)
+  - [Best Practices for Error Handling](#best-practices-for-error-handling)
+- [Migration from MemoryClient](#migration-from-memoryclient)
+- [Best Practices](#best-practices)
+- [API Reference](#api-reference)
+
+## Overview
+
+The Bedrock AgentCore Memory SDK provides a comprehensive solution for managing conversational AI memory with both short-term (conversational events) and long-term (semantic memory) storage capabilities. The SDK is designed around three main components:
+
+### Core Components
+
+1. **MemorySessionManager** - The primary interface for managing multiple sessions and actors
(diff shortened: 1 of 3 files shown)


src/bedrock_agentcore/memory/README.md
```diff
@@ -1,179 +1,293 @@
...
+
+## Recommended Classes
+
+### MemorySessionManager (Recommended)
+The primary interface for managing conversational AI sessions with both short-term (conversational events) and long-term (semantic memory) storage. Provides a clean, session-oriented API for memory operations.
+
+### MemorySession (Recommended)
+Session-scoped interface that simplifies operations by automatically handling memory_id, actor_id, and session_id parameters.
+
+### MemoryClient (Legacy)
```
bergjaak, line 110: Deprecated?

src/bedrock_agentcore/memory/README.md
```diff
@@ -1,179 +1,293 @@
...
+- Session-scoped operations with automatic parameter handling
+- Create MemorySession instances for simplified API calls
+- Built-in actor and session tracking
+
 ### Flexible Conversation API
-- Save any number of messages in a single call
-- Support for USER, ASSISTANT, TOOL, OTHER roles
+- Save any number of messages in a single call with `add_turns()`
+- Support for USER, ASSISTANT, TOOL, OTHER roles via `ConversationalMessage`
+- Support for binary data via `BlobMessage`
```
bergjaak, line 126: it is not binary but rather Document (see Smithy Document)

src/bedrock_agentcore/memory/README.md
```diff
@@ -1,179 +1,293 @@
 # Bedrock AgentCore Memory SDK
```
bergjaak, line 1: I think it would be worth it to have an Overview section like the starter repo

src/bedrock_agentcore/memory/README.md
jona62: We've not deprecated it yet

src/bedrock_agentcore/memory/README.md
jona62: Technically you can write binary data if you encode it in base64

src/bedrock_agentcore/memory/README.md
```diff
@@ -1,179 +1,587 @@
...
+3. **MemoryClient** - Legacy client interface (still supported but not recommended for new projects)
+
+### Architecture
+
+The memory system operates on a hierarchical structure:
+- **Memory** - Top-level container for all data
+- **Actor** - Represents individual users or entities
+- **Session** - Conversation contexts within an actor
+- **Events** - Individual conversation turns or actions
+- **Branches** - Alternative conversation paths for A/B testing or exploration
```
bergjaak, line 48: also for editing

src/bedrock_agentcore/memory/README.md
```diff
@@ -1,179 +1,587 @@
...
+### Retry Strategies
+
+Implement retry logic for transient failures:
+
+```python
+import time
+import random
+from botocore.exceptions import ClientError
+
+def exponential_backoff_retry(func, max_retries=3, base_delay=1):
```
bergjaak: what's this here for? Doesn't boto3 do all of this for us? I don't think we should have this or implement this in our code base since the boto3 retry config handles all of this

src/bedrock_agentcore/memory/README.md
```diff
@@ -1,179 +1,587 @@
...
+1. **Always handle authentication errors**:
+   ```python
+   try:
+       manager = MemorySessionManager(memory_id="test")
+   except NoCredentialsError:
+       # Guide user to configure credentials
+       print("Please run 'aws configure' or set AWS environment variables")
+   ```
+
+2. **Validate inputs before API calls**:
```
bergjaak, line 391: this seems like bad advice. Should let the service do it so that you can get validation updates for free (if validation changes, then you need to change the client as well)

jona62: Validation of input fields especially from users is always recommended, maybe this should be a better example shown with the `create_event` API but input validation is never a bad advice

src/bedrock_agentcore/memory/README.md
jona62: Removed

PR #106, opened 2025-10-06
Add support for Identity OAuth 2.0 federation enhancements

brianlaoaws:
*Description of changes:*

* Add support for Identity OAuth 2.0 federation enhancements


tests/bedrock_agentcore/services/test_identity.py
```diff
@@ -260,6 +264,7 @@ async def test_get_token_with_optional_parameters(self):
                 workloadIdentityToken=agent_identity_token,
                 resourceOauth2ReturnUrl=callback_url,
                 forceAuthentication=force_authentication,
+                includedState=session_state,
```
fei-yuan: Can we call it something else other than "included_state" or "session_state"?

To avoid conflating it with the session that our service generates and tracks, can we call this customer-supplied param as "custom_state" instead?  It will also be more consistent with "custom_parameters" on the underlying API.

src/bedrock_agentcore/identity/auth.py
```diff
@@ -40,6 +41,7 @@ def requires_access_token(
         callback_url: OAuth2 callback URL
         force_authentication: Force re-authentication
         token_poller: Custom token poller implementation
+        custom_state: A state that allows applications to verify the validity of call backs to callback_url
```
fei-yuan: callbacks

src/bedrock_agentcore/services/identity.py
```diff
@@ -132,6 +133,7 @@ async def get_token(
             callback_url: OAuth2 callback URL (must be pre-registered)
             force_authentication: Force re-authentication even if token exists in the token vault
             token_poller: Custom token poller implementation
+            custom_state: A state that allows applications to verify the validity of call backs to callback_url
```
fei-yuan: callbacks

PR #107, opened 2025-10-06
fix: add the include_parent_events parameter to the get_last_k_turns method

jona62:
fix: add the include_parent_events parameter to the get_last_k_turns method


src/bedrock_agentcore/memory/session.py
```diff
@@ -921,10 +921,16 @@ def process_turn_with_llm(
         )
 
     def get_last_k_turns(
-        self, k: int = 5, branch_name: Optional[str] = None, max_results: int = 100
+        self,
+        k: int = 5,
+        branch_name: Optional[str] = None,
+        include_branches: Optional[bool] = None,
```
bergjaak: if a branch name is included but include branches is not, what do we do?

I think we should name it includeParentBranches, like the API

jona62: Although I agree that it should be include_parent_branches, We have to be careful to not introduce backwards incompatible changes now that this is public

src/bedrock_agentcore/memory/session.py
```diff
@@ -625,7 +625,7 @@ def get_last_k_turns(
         session_id: str,
         k: int = 5,
         branch_name: Optional[str] = None,
-        include_branches: bool = False,
```
bergjaak, line 628: we can do a major version bump maybe?

bergjaak (approved):

src/bedrock_agentcore/memory/session.py
jona62: Possibly, I'll run this by the SDK folks

PR #116, opened 2025-10-14
Add OAuth2 3LO Identity APIs and enable setting 3LO callback URL via HTTP headers

SkyWalker-CMD:
**Overview**

Add OAuth2 3LO Identity APIs and enable setting 3LO callback URL for the `GetResourceOauth2Token` API via HTTP headers.

### DIFF
--- a/src/bedrock_agentcore/identity/auth.py
+++ b/src/bedrock_agentcore/identity/auth.py
@@ -56,7 +56,7 @@ def requires_access_token(
                 scopes=scopes,
                 on_auth_url=on_auth_url,
                 auth_flow=auth_flow,
-                callback_url=_get_workload_callback_url(callback_url),
+                callback_url=_get_oauth2_callback_url(callback_url),
                 force_authentication=force_authentication,
                 token_poller=token_poller,
             )
@@ -144,11 +144,11 @@ def requires_api_key(*, provider_name: str, into: str = "api_key") -> Callable:
     return decorator
 
 
-def _get_workload_callback_url(user_provided_callback_url: Optional[str]):
-    if user_provided_callback_url:
-        return user_provided_callback_url
+def _get_oauth2_callback_url(user_provided_oauth2_callback_url: Optional[str]):
+    if user_provided_oauth2_callback_url:
+        return user_provided_oauth2_callback_url
 
-    return BedrockAgentCoreContext.get_workload_callback_url()
+    return BedrockAgentCoreContext.get_oauth2_callback_url()
 
 
 async def _get_workload_access_token(client: IdentityClient) -> str:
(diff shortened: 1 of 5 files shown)


src/bedrock_agentcore/identity/auth.py
```diff
@@ -143,6 +143,11 @@ def sync_wrapper(*args: Any, **kwargs: Any) -> Any:
 
     return decorator
 
+def _get_workload_callback_url(user_provider_callback_url: Optional[str]):
+    if user_provider_callback_url:
+        return user_provider_callback_url
+    
+    return BedrockAgentCoreContext.get_workload_callback_url()
```
brianlaoaws: Should we check for existence here?

src/bedrock_agentcore/runtime/app.py
```diff
@@ -280,6 +281,10 @@ def _build_request_context(self, request) -> RequestContext:
             agent_identity_token = headers.get(ACCESS_TOKEN_HEADER)
             if agent_identity_token:
                 BedrockAgentCoreContext.set_workload_access_token(agent_identity_token)
+                
+            agent_workload_callback_url = headers.get(CALLBACK_URL_HEADER)
+            if agent_workload_callback_url:
+                BedrockAgentCoreContext.set_workload_callback_url(agent_workload_callback_url)
```
brianlaoaws: Why do we need to set the callback url as a header?

SkyWalker-CMD: This is because the callback url is required as a parameter to the `GetResourceOauth2Token` API call made in the SDK.

* Ref: https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_GetResourceOauth2Token.html#BedrockAgentCore-GetResourceOauth2Token-request-resourceOauth2ReturnUrl

src/bedrock_agentcore/identity/auth.py
SkyWalker-CMD: Good point, the default behavior for the `callback_url` is `None`, so the existence check would just return `None` if not present, so can be skipped.

fei-yuan (approved):

brianlaoaws (dismissed):

src/bedrock_agentcore/runtime/models.py
```diff
@@ -17,6 +17,7 @@ class PingStatus(str, Enum):
 SESSION_HEADER = "X-Amzn-Bedrock-AgentCore-Runtime-Session-Id"
 REQUEST_ID_HEADER = "X-Amzn-Bedrock-AgentCore-Runtime-Request-Id"
 ACCESS_TOKEN_HEADER = "WorkloadAccessToken"  # nosec
+CALLBACK_URL_HEADER = "WorkloadCallbackUrl"
```
siwachabhi: How does customer pass this url during invoke?

siwachabhi (approved):

fei-yuan (approved):

PR #119, opened 2025-10-16
Update Outbound Oauth error message

RushikeshJoshi4:
Update Outbound Oauth error message

### DIFF
--- a/src/bedrock_agentcore/identity/auth.py
+++ b/src/bedrock_agentcore/identity/auth.py
@@ -151,7 +151,11 @@ async def _get_workload_access_token(client: IdentityClient) -> str:
     else:
         # workload access token context var was not set, so we should be running in a local dev environment
         if os.getenv("DOCKER_CONTAINER") == "1":
-            raise ValueError("Workload access token has not been set.")
+            raise ValueError(
+                "Workload access token has not been set. If invoking agent runtime via SIGV4 inbound auth, "
+                "please specify the X-Amzn-Bedrock-AgentCore-Runtime-User-Id header and retry. "
+                "For details, see - https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-oauth.html"
+            )
 
         return await _set_up_local_auth(client)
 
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/identity/auth.py
```diff
@@ -151,7 +151,11 @@ async def _get_workload_access_token(client: IdentityClient) -> str:
     else:
         # workload access token context var was not set, so we should be running in a local dev environment
         if os.getenv("DOCKER_CONTAINER") == "1":
-            raise ValueError("Workload access token has not been set.")
+            raise ValueError(
+                "Workload access token has not been set. If invoking agent runtime via SIGV4 inbound auth, "
```
siwachabhi, line 155: Thanks for adding this!, just to confirm if invoke path is not sigv4, how does this work? is this header still required during invoke?

src/bedrock_agentcore/identity/auth.py
```diff
@@ -151,7 +151,11 @@ async def _get_workload_access_token(client: IdentityClient) -> str:
     else:
         # workload access token context var was not set, so we should be running in a local dev environment
         if os.getenv("DOCKER_CONTAINER") == "1":
-            raise ValueError("Workload access token has not been set.")
+            raise ValueError(
```
aws-bviswa, line 154: Can we determine if the request is using SigV4 or IAM auth type and provide more accurate error message in case of missing workload access token?
- In SigV4 case, we need the header passed as the error message says
- But in Oauth case, we don't need this header, so don't want to distract customers who are already using OAuth path and are not passing this header (because JWT token is already available)?

src/bedrock_agentcore/identity/auth.py
RushikeshJoshi4: For the inbound oauth path, this header is not required, but bearer-token is required. If bearer-token is not passed, you'll get an AccessDenied with the appropriate error message.

Now, if both the bearer-token and the user-id header are passed, the user-id will be ignored. We also get a warning log like this - 
```
Both bearer token and user id are specified, ignoring user id
```

siwachabhi (approved):

src/bedrock_agentcore/identity/auth.py
RushikeshJoshi4: My understanding is, for Oauth case, it should never end up here. As you said, if JWT token is not there, then that is a 403 Access denied even before this code path is reached. 

There is no clean way to check if its IAM Sigv4 vs Oauth inbound request.

PR #131, opened 2025-10-22
feat: support for async llm callback

OpenGaet:
*Issue #, if available:*

*Description of changes:*
### Added Async LLM Callback Support

New Features:
• Added process_turn_with_llm_async() method to MemorySessionManager and MemorySession classes
• Supports async LLM callbacks that return Awaitable[str]
• Maintains full backwards compatibility with existing sync callbacks

Code Improvements:
• Extracted common logic into helper methods (_retrieve_memories_for_llm() and _save_conversation_turn())
• Added test coverage for async functionality

Files Modified:
• src/bedrock_agentcore/memory/session.py (+122/-16 lines)
• tests/bedrock_agentcore/memory/test_session.py (+35 lines)

Usage:
```python
# Async callback support
await manager.process_turn_with_llm_async(                                                                                                                                                                         
    actor_id="user-123",                                                                                                                                                                                           
    session_id="session-456",                                                                                                                                                                                      
    user_input="Hello",                                                                                                                                                                                            
    llm_callback=async_callback,  # async function                                                                                                                                                                 
    retrieval_config=config                                                                                                                                                                                        
)
```

### DIFF
--- a/src/bedrock_agentcore/memory/session.py
+++ b/src/bedrock_agentcore/memory/session.py
@@ -1,9 +1,10 @@
 """Module containing session management classes for AgentCore Memory interactions."""
 
+import asyncio
 import logging
 import uuid
 from datetime import datetime, timezone
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Awaitable, Callable, Dict, List, Optional, Tuple, Union
 
 import boto3
 from botocore.config import Config as BotocoreConfig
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/memory/session.py
```diff
@@ -302,6 +303,78 @@ def my_llm(user_input: str, memories: List[Dict]) -> str:
...
+    def _retrieve_memories_for_llm(
+        self,
+        actor_id: str,
+        session_id: str,
+        user_input: str,
+        retrieval_config: Optional[Dict[str, RetrievalConfig]],
+    ) -> List[Dict[str, Any]]:
+        """Helper method to retrieve memories for LLM context."""
         retrieved_memories = []
         if retrieval_config:
```
bergjaak, line 379: What if instead, we make it so `_retrieve_memories_for_llm` requires the `retrieval_config` but then this `if retrieval_config` check happens where we call _retrieve_memories_for_llm? I think that would be a less accident prone API since then there is no way the user tries to call _retrieve_memories_for_llm without a retrieval config and is surprised that they get no memory records

bergjaak (approved):

src/bedrock_agentcore/memory/session.py
OpenGaet: This was the existing logic, just moved to a separate function to reduce code duplication. On top of this, it is a "private" function, so the user should not be calling `_retrieve_memories_for_llm` directly. But feel free to make changes as desired.

PR #163, opened 2025-11-19
feat: parallelize retrieve memories API calls for multiple namespaces to improve latency

colifran:
*Description of changes:*
- Refactored `retrieve_customer_context` to retrieve memories from multiple namespaces concurrently using `ThreadPoolExecutor` instead of sequentially
- Extracted namespace retrieval logic into a helper function `retrieve_for_namespace` for cleaner concurrent execution
- Added per-namespace error handling so failures in one namespace don't block retrieval from others (graceful degradation)
- Removed unneeded braces around `len(all_context)`

This improves response time for support queries when multiple namespaces are configured, reducing total retrieval time from the sum of all calls to the maximum of any single call.

*Testing:*
- I performed manual testing with tracing enabled this screenshot shows the current implementation is making `retrieve_memories` API calls sequentially:
Screenshot 2025-11-19 at 2 24 53 PM

- This screenshot shows that `retrieve_memories` API calls are made in parallel with the updated code:
Screenshot 2025-11-10 at 3 59 36 PM

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -3,6 +3,7 @@
 import json
 import logging
 import threading
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime, timedelta, timezone
 from typing import TYPE_CHECKING, Any, Optional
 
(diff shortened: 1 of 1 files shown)


src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -494,30 +495,50 @@ def retrieve_customer_context(self, event: MessageAddedEvent) -> None:
...
-
-                for memory in memories:
-                    if isinstance(memory, dict):
-                        content = memory.get("content", {})
-                        if isinstance(content, dict):
-                            text = content.get("text", "").strip()
-                            if text:
-                                all_context.append(text)
+            
+            with ThreadPoolExecutor() as executor:
```
bergjaak, line 527: im guessing it's fine since we do this `with` syntax, but do we know if this spawns a thread or something that could lead to thread exhaustion? It would be great to just confirm that's not an issue

bergjaak: https://docs.python.org/3/library/concurrent.futures.html yeah seems fine

bergjaak: https://docs.python.org/3/library/concurrent.futures.html yeah seems fine

colifran: The `ThreadPoolExecutor` does spawn threads but it uses a thread pool and doesn't spawn unlimited threads - instead it reuses threads from the pool. The default is `min(32, os.cpu_count() + 4)`. The `ThreadPoolExecutor` lifecycle is managed within the context manager which is defined using `with`. In Python `with` is a context manager and does the following:
- Ensures all threads are properly shut down when the block exits
- Waits for any in-flight tasks to complete before closing
- Ensures resources are cleans up even if an exception occurs

bergjaak (dismissed):

Issue #168, opened 2025-11-24
Add Pyupgrade (UP) to ruff config


maxrabin:
Much of the code uses older Python syntax. Use Ruff to auto-upgrade to more modern syntax.

maxrabin:
Hey team, I'd love to hear if there's anything I can do to push this forward. The core change is just adding UP to Ruff, the rest of the changes are the results of running pre-commit with the new Ruff config.

PR #169, opened 2025-11-24
Upgrade Python syntax by using Pyupgrade, adding it to Ruff (UP).

maxrabin:
*Issue #, if available:* #168

*Description of changes:* Upgrade Python syntax by using Pyupgrade, adding it to Ruff (UP).

Also fix reference in ruff config from "tests-integ" to "tests_integ"

### DIFF
--- a/scripts/bump_version.py
+++ b/scripts/bump_version.py
@@ -3,7 +3,6 @@ import subprocess
 import sys
 from datetime import datetime
 from pathlib import Path
-from typing import Optional, Tuple
 
 
 def get_current_version() -> str:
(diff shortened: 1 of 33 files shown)


sundargthb:
Thanks for the contribution and patience! We're going to close this for now:

- Massive conflict surface — this PR touches many core files across the SDK and is heavily conflicted with main. Resolving these conflicts carries real risk of introducing regressions in critical paths (runtime, memory, etc.).

- Cosmetic, not functional — Optional[str] vs str | None and Dict vs dict are pure syntax preferences with zero behavioral impact. The codebase is internally consistent as-is.

- We'd want to own this internally — if we decide to adopt UP rules, we'd add it to our ruff config and run it ourselves in a single controlled commit against current main, with full CI validation. Taking it through an external PR that's stale and conflicted across dozens of files is the wrong path.

That said, the tests-integ → tests_integ typo fix in the ruff config is a valid catch — if you'd like to open a tiny PR for just that one-line fix, we'd happily review it.

We'll track the UP rule addition internally. Appreciate the suggestion!

PR #179, opened 2025-12-01
feat(identity):  Add @requires_iam_access_token decorator for AWS STS JWT tokens

sundargthb:
- Extend auth_flow parameter to accept AWS_JWT in addition to M2M/USER_FEDERATION
- Add get_aws_jwt_token() and get_aws_jwt_token_sync() methods to IdentityClient

Extends the `@requires_access_token` decorator to support AWS JWT federation as a third authentication flow alongside OAuth M2M and USER_FEDERATION.

## Motivation

AWS IAM Outbound Web Identity Federation allows agents to obtain signed JWTs directly from AWS STS without:
- Credential providers
- Client secrets
- AgentCore Identity service involvement

This simplifies M2M authentication for external services that support OIDC token validation.

### IdentityClient (`identity.py`)

Added new methods:
- `get_aws_jwt_token()` - Async method to call STS:GetWebIdentityToken
- `get_aws_jwt_token_sync()` - Synchronous version

### Decorator (`auth.py`)

Extended `@requires_access_token`:
- Added `"AWS_JWT"` to `auth_flow` Literal type
- Added new parameters: `audience`, `signing_algorithm`, `duration_seconds`, `tags`
- Routes to STS when `auth_flow="AWS_JWT"`
- Validates required parameters based on flow type

## Usage

```python
# AWS JWT flow (no secrets required)
@tool
@requires_access_token(
    auth_flow="AWS_JWT",
    audience=["https://api.example.com"],
)
def call_external_api(*, access_token: str) -> str:
    # access_token is an AWS-signed JWT
    return requests.get(url, headers={"Authorization": f"Bearer {access_token}"}).text
```

## Parameter Reference

| Parameter | Required For | Description |
|-----------|--------------|-------------|
| `provider_name` | M2M, USER_FEDERATION | Credential provider name |
| `scopes` | M2M, USER_FEDERATION | OAuth scopes |
| `audience` | AWS_JWT | Token recipient URLs |
| `signing_algorithm` | AWS_JWT (optional) | ES384 (default) or RS256 |
| `duration_seconds` | AWS_JWT (optional) | Token lifetime (60-3600) |

## Breaking Changes

None. New `auth_flow` value is additive. Existing code unchanged.


tests_integ/identity/test_auth_flows.py
```diff
@@ -30,7 +30,83 @@ async def need_api_key(*, api_key: str):
...
+
+
 if __name__ == "__main__":
-    asyncio.run(need_api_key(api_key=""))
-    asyncio.run(need_token_2LO_async(access_token=""))
-    asyncio.run(need_token_3LO_async(access_token=""))
+    # OAuth flows (require credential providers to be set up)
+    # asyncio.run(need_api_key(api_key=""))
+    # asyncio.run(need_token_2LO_async(access_token=""))
+    # asyncio.run(need_token_3LO_async(access_token=""))
```
SkyWalker-CMD: Just to confirm, is this expected to comment out these tests?

src/bedrock_agentcore/services/identity.py
```diff
@@ -249,3 +249,97 @@ async def get_api_key(self, *, provider_name: str, agent_identity_token: str) ->
...
+            return {
+                "token": response["WebIdentityToken"],
+                "expiration": response["Expiration"],
+            }
+
+        except Exception as e:
+            self.logger.error("Failed to get AWS JWT token: %s", str(e))
+            raise
+
+    def get_aws_jwt_token_sync(
```
SkyWalker-CMD: I think this may cause some confusion, because the AWS JWT token is an STS API, and not related to AgentCore Identity. 

Just curious, why not use the STS client directly in the `@requires_access_token` annotation? (instead of indirectly via the AgentCore Identity client)

src/bedrock_agentcore/services/identity.py
```diff
@@ -249,3 +249,97 @@ async def get_api_key(self, *, provider_name: str, agent_identity_token: str) ->
...
+            "Audience": audience,
+            "SigningAlgorithm": signing_algorithm,
+            "DurationSeconds": duration_seconds,
+        }
+
+        if tags:
+            params["Tags"] = tags
+
+        try:
+            response = sts_client.get_web_identity_token(**params)
```
SkyWalker-CMD: Just to confirm, should there by any version bump for the `boto`/`boto3` client in the `uv.lock` file? 

This is a new STS API which is not available in the current version of `boto`, so will most likely fail at runtime when executed.

src/bedrock_agentcore/identity/auth.py
```diff
@@ -20,41 +20,94 @@
...
+    scopes: Optional[List[str]] = None,
     on_auth_url: Optional[Callable[[str], Any]] = None,
-    auth_flow: Literal["M2M", "USER_FEDERATION"],
     callback_url: Optional[str] = None,
     force_authentication: bool = False,
     token_poller: Optional[TokenPoller] = None,
     custom_state: Optional[str] = None,
     custom_parameters: Optional[Dict[str, str]] = None,
+    # AWS JWT parameters (required for AWS_JWT)
+    audience: Optional[List[str]] = None,
```
SkyWalker-CMD: I think it may make more sense to create a completely brand new annotation for the STS JWT token, as otherwise this `@requires_access_token` annotation becomes too overloaded.

Also, because the new JWT token is not related to AgentCore Identity, and actually calls an API on STS, so for better separation, might be better to create a new annotation.

src/bedrock_agentcore/services/identity.py
```diff
@@ -249,3 +249,97 @@ async def get_api_key(self, *, provider_name: str, agent_identity_token: str) ->
         req = {"resourceCredentialProviderName": provider_name, "workloadIdentityToken": agent_identity_token}
 
         return self.dp_client.get_resource_api_key(**req)["apiKey"]
+
+    async def get_aws_jwt_token(
```
SkyWalker-CMD: Just curious, what's the difference between this function and `get_aws_jwt_token_sync`?

The name suggests the difference is `async` vs `sync`, but the implementation looks the same, as far as I can tell.

src/bedrock_agentcore/identity/auth.py
SkyWalker-CMD: I was thinking of an annotation named `@requires_iam_access_token` or a name along those lines, but curious on your thoughts.

tests_integ/identity/test_auth_flows.py
sundargthb: removed the comments..

src/bedrock_agentcore/services/identity.py
sundargthb: agree, the STS call has nothing to do with AgentCore Identity. moved the STS call directly into the decorator

src/bedrock_agentcore/services/identity.py
sundargthb: we will be updating this as part of reinvent release..

src/bedrock_agentcore/identity/auth.py
sundargthb: created a separate `@requires_iam_access_token` decorator, specifically for STS JWT tokens.

src/bedrock_agentcore/services/identity.py
sundargthb: With the new `@requires_iam_access_token` decorator approach, the STS call will be inline in the decorator anyway.

src/bedrock_agentcore/services/identity.py
SkyWalker-CMD: Perfect, sounds good. Thanks!

Issue #189, opened 2025-12-12
Runtime cannot load agent when using latest 1.1.1 release


Loupi:
**Describe the bug**
Using the latest 1.1.1 release with direct code deployment (zip), when using "test endpoint" in the AWS console, 
getting this error: Unable to invoke endpoint successfully: Runtime initialization time exceeded. Please make sure that initialization completes in 30s..

**To Reproduce**
Upgrate to 1.1.1, build zip, deploy, test endpoint.

**Expected behavior**
Agent runs

**Additional context**
Downgrading to 1.0.7 everything runs fine.

Agent built with Strands 1.19.0.

The bedrock-agentcore 1.1.1 package added pre-commit>=4.2.0 as a hard runtime dependency, which is unusual and likely a packaging bug. This brings in 10+ unnecessary transitive dependencies (virtualenv, nodeenv, cfgv, identify, etc.) that may cause compatibility issues in the production environment.
The build zip file was 8Mb larger.

baburyx:
got the same error, checked the log, and it says.
2025-12-15T16:12:19.062000+00:00 2025/12/15/[runtime-logs-b2a1eede-1c70-4d9f-b501-d86386dc26c5]fa2e2f3e-c6aa-4f9c-8cfa-941853a048d5 /opt/aws/agentcore-runtime/python/versions/3.11.14/bin/python: can't open file '/var/task/main.py': [Errno 13] Permission denied
even though the entypoint is set correctly. I need this fixed asap

doronneumann:
same here

PR #194, opened 2025-12-22
fix: pre-commit causes unnecessary dependenties

kentakozuka:
Removes unnecessary dependencies for this package. pre-commit should be included in dev-dependencies.

fixes https://github.com/aws/bedrock-agentcore-sdk-python/issues/189


uv.lock
```diff
@@ -1,14 +1,14 @@
 version = 1
```
jariy17, line 1: Why is there so many changes in the uv.lock. We are just removing a dependency.

PR #202, opened 2026-01-01
feat(code-interpreter): Add convenience methods for file operations and package management

sundargthb:
## Summary

Adds developer-friendly convenience methods to `CodeInterpreter` class, reducing boilerplate for common operations while maintaining full backward compatibility.

## Motivation

Current usage requires verbose `invoke()` calls with manual parameter construction:
```python
# Current - verbose
client.invoke('writeFiles', {'content': [{'path': 'data.csv', 'text': content}]})
client.invoke('executeCommand', {'command': 'pip install pandas numpy'})
```

This PR enables cleaner patterns that match developer expectations:
```python
# New - intuitive
client.upload_file('data.csv', content, description='Sales data')
client.install_packages(['pandas', 'numpy'])
```

### New Methods
| Method | Purpose |
|--------|---------|
| `upload_file(path, content, description)` | Upload file with semantic context |
| `upload_files(files)` | Batch file upload |
| `install_packages(packages, upgrade)` | Install pip packages safely |
| `download_file(path)` | Read file content |
| `download_files(paths)` | Batch file read |
| `execute_code(code, language, clear_context)` | Typed code execution |
| `execute_shell(command)` | Shell command wrapper |

### Security
- Rejects absolute paths (`/etc/passwd` → `ValueError`)
- Blocks shell injection in package names (`;`, `&`, `|`, `` ` ``, `$`)
- Validates language parameter

## Backward Compatibility

✅ **No breaking changes** - All existing `invoke()` patterns continue to work unchanged.

## Testing

- Unit tests with mocked `invoke()` calls
- Integration tests against live AgentCore environment
- Session reuse verification

### DIFF
--- a/src/bedrock_agentcore/tools/code_interpreter_client.py
+++ b/src/bedrock_agentcore/tools/code_interpreter_client.py
@@ -11,6 +11,7 @@ from contextlib import contextmanager
 from typing import Any, Dict, Generator, List, Optional, Union
 
 import boto3
+from botocore.config import Config
 
 from bedrock_agentcore._utils.endpoints import get_control_plane_endpoint, get_data_plane_endpoint
 
@@ -83,6 +84,7 @@ class CodeInterpreter:
             "bedrock-agentcore",
             region_name=region,
             endpoint_url=get_data_plane_endpoint(region),
+            config=Config(read_timeout=300),
         )
 
         self._identifier = None
@@ -490,7 +492,7 @@ class CodeInterpreter:
         result = self.invoke("writeFiles", {"content": [file_content]})
 
         # Store description as metadata (available for future LLM context)
-        if description and hasattr(self, "_file_descriptions"):
+        if description:
             self._file_descriptions[path] = description
 
         return result
@@ -500,6 +502,9 @@ class CodeInterpreter:
         files: List[Dict[str, str]],
     ) -> Dict[str, Any]:
         """Upload multiple files to the code interpreter environment.
+
+        This operation is atomic - either all files are written or none are.
+        If any file fails, the entire operation fails.
 
         Args:
             files: List of file specifications, each containing:
@@ -666,6 +671,7 @@ class CodeInterpreter:
                     Default is 'python'.
             clear_context: If True, clears all previous variable state before execution.
                         Default is False (variables persist across calls).
+                        Note: Only supported for Python. Ignored for JavaScript/TypeScript.
 
         Returns:
             Dict containing execution results including stdout, stderr, exit_code.
@@ -696,7 +702,7 @@ class CodeInterpreter:
             },
         )
 
-    def execute_shell(
+    def execute_command(
         self,
         command: str,
     ) -> Dict[str, Any]:
@@ -712,13 +718,41 @@ class CodeInterpreter:
 
         Example:
             >>> # List files
-            >>> result = client.execute_shell('ls -la')
+            >>> result = client.execute_command('ls -la')
 
             >>> # Check Python version
-            >>> result = client.execute_shell('python --version')
+            >>> result = client.execute_command('python --version')
         """
         self.logger.info("Executing shell command: %s...", command[:50])
         return self.invoke("executeCommand", {"command": command})
+
+    def clear_context(self) -> Dict[str, Any]:
+        """Clear all variable state in the Python execution context.
+
+        This resets the interpreter to a fresh state, removing all
+        previously defined variables, imports, and function definitions.
+
+        Note: Only affects Python context. JavaScript/TypeScript contexts
+        are not affected.
+
+        Returns:
+            Dict containing the result of the clear operation.
+
+        Example:
+            >>> client.execute_code('x = 10')
+            >>> client.execute_code('print(x)')  # prints 10
+            >>> client.clear_context()
+            >>> client.execute_code('print(x)')  # NameError: x is not defined
+        """
+        self.logger.info("Clearing Python execution context")
+        return self.invoke(
+            "executeCode",
+            {
+                "code": "# Context cleared",
+                "language": "python",
+                "clearContext": True,
+            },
+        )
 
 
 @contextmanager
--- a/tests/bedrock_agentcore/tools/test_code_interpreter_client.py
+++ b/tests/bedrock_agentcore/tools/test_code_interpreter_client.py
@@ -1,6 +1,6 @@
 import base64
 import datetime
-from unittest.mock import MagicMock, patch
+from unittest.mock import ANY, MagicMock, patch
 
 import pytest
 
@@ -34,7 +34,10 @@ class TestCodeInterpreterClient:
             endpoint_url="https://mock-control-endpoint.com",
         )
         mock_session.client.assert_any_call(
-            "bedrock-agentcore", region_name=region, endpoint_url="https://mock-data-endpoint.com"
+            "bedrock-agentcore",
+            region_name=region,
+            endpoint_url="https://mock-data-endpoint.com",
+            config=ANY,
         )
         assert client.control_plane_client == mock_control_client
         assert client.data_plane_client == mock_data_client
@@ -1273,7 +1276,7 @@ class TestCodeInterpreterClient:
         client.data_plane_client.invoke_code_interpreter.return_value = mock_response
 
         # Act
-        result = client.execute_shell("ls -la")
+        result = client.execute_command("ls -la")
 
         # Assert
         client.data_plane_client.invoke_code_interpreter.assert_called_once_with(
@@ -1300,7 +1303,7 @@ class TestCodeInterpreterClient:
         client.data_plane_client.invoke_code_interpreter.return_value = mock_response
 
         # Act
-        result = client.execute_shell("python --version")
+        result = client.execute_command("python --version")
 
         # Assert
         client.data_plane_client.invoke_code_interpreter.assert_called_once_with(
--- a/tests_integ/tools/test_code.py
+++ b/tests_integ/tools/test_code.py
@@ -88,16 +88,16 @@ print("✅ Test 5 passed\n")
 
 
 # Test 6: Execute shell command
-print("Test 6: Execute shell command using execute_shell()")
+print("Test 6: Execute shell command using execute_command()")
 with code_session("us-west-2") as client:
     # Check Python version
-    result = client.execute_shell("python --version")
+    result = client.execute_command("python --version")
     print("Shell command result:")
     for event in result["stream"]:
         print(event["result"]["content"])
 
     # List directory
-    result = client.execute_shell("ls -la")
+    result = client.execute_command("ls -la")
     print("Directory listing:")
     for event in result["stream"]:
         print(event["result"]["content"])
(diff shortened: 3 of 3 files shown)


src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -404,6 +430,296 @@ def invoke(self, method: str, params: Optional[Dict] = None):
...
+
+        if description:
+            self.logger.info("Uploading file: %s (%s)", path, description)
+        else:
+            self.logger.info("Uploading file: %s", path)
+
+        result = self.invoke("writeFiles", {"content": [file_content]})
+
+        # Store description as metadata (available for future LLM context)
+        if description and hasattr(self, "_file_descriptions"):
```
aws-bviswa: do we need the hasattr check here? I see _file_descriptions is defined above as {}?

src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -404,6 +430,296 @@ def invoke(self, method: str, params: Optional[Dict] = None):
...
+            if path.startswith("/"):
+                raise ValueError(f"Path must be relative, not absolute. Got: {path}")
+
+            if isinstance(content, bytes):
+                file_contents.append({"path": path, "blob": base64.b64encode(content).decode("utf-8")})
+            else:
+                file_contents.append({"path": path, "text": content})
+
+        self.logger.info("Uploading %d files", len(files))
+        return self.invoke("writeFiles", {"content": file_contents})
```
aws-bviswa, line 538: didn't follow this. we are taking list of files and there is no batch file upload API in code interpreter? so we should just call upload_file utility we have above?

Also need to see how to handle partial failure in this case. Throwing an exception even if one file upload files should be ok and users will have to retry entire batch. Some advanced support would be to instead return list of responses back and let clients handle which ones to retry with

src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -404,6 +430,296 @@ def invoke(self, method: str, params: Optional[Dict] = None):
             arguments=params or {},
         )
 
+    def upload_file(
+        self,
+        path: str,
+        content: Union[str, bytes],
+        description: str = "",
+    ) -> Dict[str, Any]:
```
aws-bviswa, line 440: are we not able to strengthen the return type further here? what's the expected shape we plan to return here?

src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -404,6 +430,296 @@ def invoke(self, method: str, params: Optional[Dict] = None):
...
+            Dict mapping file paths to their contents.
+
+        Example:
+            >>> files = client.download_files(['data.csv', 'results.json'])
+            >>> print(files['data.csv'])
+        """
+        self.logger.info("Downloading %d files", len(paths))
+        result = self.invoke("readFiles", {"paths": paths})
+
+        files = {}
```
aws-bviswa, line 640: can we re_use the utility method above to download_file?

src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -404,6 +430,296 @@ def invoke(self, method: str, params: Optional[Dict] = None):
...
+                            elif "blob" in resource:
+                                files[file_path] = base64.b64decode(resource["blob"]).decode("utf-8")
+
+        return files
+
+    def execute_code(
+        self,
+        code: str,
+        language: str = "python",
+        clear_context: bool = False,
```
aws-bviswa, line 661: minor note: this clear_context is only application for python and is not supported for other languages yet.

src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -404,6 +430,296 @@ def invoke(self, method: str, params: Optional[Dict] = None):
...
+        """Execute code in the interpreter environment.
+
+        This is a convenience wrapper around the executeCode method with
+        typed parameters for better IDE support and validation.
+
+        Args:
+            code: The code to execute.
+            language: Programming language - 'python', 'javascript', or 'typescript'.
+                    Default is 'python'.
+            clear_context: If True, clears all previous variable state before execution.
```
aws-bviswa, line 672: should we consider adding a first class method to clear context? so that users can write code such as:
```
client.executeCode(`x = 10`)
client.executeCode(`x += 1`)
client.clearContext() // this can be a dummy call we run with clearContext = True
client.executeCode(`x = 1`)
```

so it's easy and clean to clear context instead of always needing to set it along with another execute code call.

src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -404,6 +430,296 @@ def invoke(self, method: str, params: Optional[Dict] = None):
...
+        return self.invoke(
+            "executeCode",
+            {
+                "code": code,
+                "language": language,
+                "clearContext": clear_context,
+            },
+        )
+
+    def execute_shell(
```
aws-bviswa: should we call this exec() or execute_command() to keep consistency with API?

src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -404,6 +430,296 @@ def invoke(self, method: str, params: Optional[Dict] = None):
...
+                raise ValueError(f"Invalid characters in package name: {pkg}")
+
+        packages_str = " ".join(packages)
+        upgrade_flag = "--upgrade " if upgrade else ""
+        command = f"pip install {upgrade_flag}{packages_str}"
+
+        self.logger.info("Installing packages: %s", packages_str)
+        return self.invoke("executeCommand", {"command": command})
+
+    def download_file(
```
aws-bviswa, line 583: should we call this read/write file to keep it mapped to API input names? (we may introduce download/upload for some future operations related to blob storage upload/download)

src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -31,6 +32,30 @@ class CodeInterpreter:
         client: The boto3 client for interacting with the service.
         identifier (str, optional): The code interpreter identifier.
         session_id (str, optional): The active session ID.
+
+    Basic Usage:
```
aws-bviswa, line 37: what if we could have subclasses to organize the utility methods based on:
- file system controls
- code execution controls
- command line controls

so we would end up with:
```
session.file.read()
session.file.write()

session.code.execute()

session.cmd.execute()
session.cmd.startTask()
session.cmd.getTask()
session.cmd.stopTask()
```

src/bedrock_agentcore/tools/code_interpreter_client.py
```diff
@@ -404,6 +430,296 @@ def invoke(self, method: str, params: Optional[Dict] = None):
...
+                "language": language,
+                "clearContext": clear_context,
+            },
+        )
+
+    def execute_shell(
+        self,
+        command: str,
+    ) -> Dict[str, Any]:
+        """Execute a shell command in the interpreter environment.
```
aws-bviswa, line 709: we also have few command task related API's we could have wrappers for

https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/code-interpreter-api-reference-examples.html

this can also be a follow-up

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: will remove this..

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: The writeFiles API does accept a list, so it is a batch operation at the API level. But your point about reusing upload_file and handling partial failures is valid. 

Two options:
- Keep as-is since writeFiles API handles the batch natively (simpler, atomic)
- Loop through and call upload_file for each, collect results, and let caller handle partial failures

I'd lean toward Option A for now since the API handles it atomically - either all succeed or all fail. We can add advanced partial-failure handling as a follow-up if users request it.

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: The return type is Dict[str, Any] because invoke() returns that. We could define a TypedDict for the response shape, but it couples us to the API response structure. For now, keeping Dict[str, Any] gives flexibility. Can revisit if we want stricter typing across the SDK.

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: Yes, makes sense. Will refactor to reuse download_file

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: Good point. Will update the docstring to clarify this limitation.

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: cleaner API. Will add.

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: sure, execute_command matches the API name better. Will rename.

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: i think upload_file/download_file are more intuitive from a user perspective than write_file/read_file. The existing invoke('writeFiles') and invoke('readFiles') are the low-level API - these convenience methods add some semantic clarity.

Suggest we keep upload_file/download_file for now, and if we add blob storage later, we can name those upload_to_s3/download_from_s3 or similar.

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: can we keep this as a follow-up enhancement. Current flat structure works and matches how other SDKs typically expose methods. We can layer subclasses on top in a future version without breaking changes.

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: Agree these would be useful. Already have invoke() paths for these. Can add as follow-up since the core file/code operations are the priority.

src/bedrock_agentcore/tools/code_interpreter_client.py
sundargthb: on second thought, the readFiles API accepts multiple paths in one call, so download_files is more efficient as a batch operation (1 API call) rather than looping through download_file (N API calls). The parsing logic is duplicated but minor - can extract to a helper if you feel strongly.

src/bedrock_agentcore/tools/code_interpreter_client.py
aws-bviswa: - it will be hard to change structure as a follow-up as then customers will need to change code to support this though right? 
- if this matches other SDK's and structures what customers are already familiar with then it's okay to leave it as is.

PR #208, opened 2026-01-08
feat(memory): add episodic memory strategy support

aneeshsabarad:
- Add EPISODIC to StrategyType, MemoryStrategyTypeEnum, OverrideType enums
- Add EPISODIC to DEFAULT_NAMESPACES
- Add episodic wrapper keys for extraction, consolidation, reflection
- Add add_episodic_strategy() and add_episodic_strategy_and_wait()
- Add add_custom_episodic_strategy() and add_custom_episodic_strategy_and_wait()
- Update _wrap_configuration() to handle EPISODIC_OVERRIDE
- Add unit tests for all new episodic functionality

*Issue #, if available:*

*Description of changes:*

Enable support for using Episodic Strategies with AgentCore Memory

### DIFF
--- a/src/bedrock_agentcore/memory/client.py
+++ b/src/bedrock_agentcore/memory/client.py
@@ -22,6 +22,7 @@ from botocore.exceptions import ClientError
 from .constants import (
     CUSTOM_CONSOLIDATION_WRAPPER_KEYS,
     CUSTOM_EXTRACTION_WRAPPER_KEYS,
+    CUSTOM_REFLECTION_WRAPPER_KEYS,
     DEFAULT_NAMESPACES,
     EXTRACTION_WRAPPER_KEYS,
     MemoryStatus,
@@ -1419,6 +1420,51 @@ class MemoryClient:
         self.add_user_preference_strategy(memory_id, name, description, namespaces)
         return self._wait_for_memory_active(memory_id, max_wait, poll_interval)
 
+    def add_episodic_strategy(
+        self,
+        memory_id: str,
+        name: str,
+        reflection_namespaces: List[str],
+        description: Optional[str] = None,
+        namespaces: Optional[List[str]] = None,
+    ) -> Dict[str, Any]:
+        """Add an episodic memory strategy.
+
+        Args:
+            memory_id: Memory resource ID
+            name: Strategy name
+            reflection_namespaces: Namespaces for reflections (can be less nested than episode namespaces)
+            description: Optional description
+            namespaces: Optional namespaces for episodes
+        """
+        strategy: Dict = {
+            StrategyType.EPISODIC.value: {
+                "name": name,
+                "reflectionConfiguration": {"namespaces": reflection_namespaces},
+            }
+        }
+
+        if description:
+            strategy[StrategyType.EPISODIC.value]["description"] = description
+        if namespaces:
+            strategy[StrategyType.EPISODIC.value]["namespaces"] = namespaces
+
+        return self._add_strategy(memory_id, strategy)
+
+    def add_episodic_strategy_and_wait(
+        self,
+        memory_id: str,
+        name: str,
+        reflection_namespaces: List[str],
+        description: Optional[str] = None,
+        namespaces: Optional[List[str]] = None,
+        max_wait: int = 300,
+        poll_interval: int = 10,
+    ) -> Dict[str, Any]:
+        """Add an episodic strategy and wait for memory to return to ACTIVE state."""
+        self.add_episodic_strategy(memory_id, name, reflection_namespaces, description, namespaces)
+        return self._wait_for_memory_active(memory_id, max_wait, poll_interval)
+
     def add_custom_semantic_strategy(
         self,
         memory_id: str,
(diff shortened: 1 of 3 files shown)


src/bedrock_agentcore/memory/client.py
```diff
@@ -1419,6 +1420,52 @@ def add_user_preference_strategy_and_wait(
         self.add_user_preference_strategy(memory_id, name, description, namespaces)
         return self._wait_for_memory_active(memory_id, max_wait, poll_interval)
 
+    def add_episodic_strategy(
+        self,
+        memory_id: str,
+        name: str,
+        description: Optional[str] = None,
+        namespaces: Optional[List[str]] = None,
+        reflection_namespaces: Optional[List[str]] = None,
```
abhivr-amazon: should this be optional if we require a reflection namespace?

src/bedrock_agentcore/memory/client.py
```diff
@@ -1419,6 +1420,52 @@ def add_user_preference_strategy_and_wait(
...
+
+        return self._add_strategy(memory_id, strategy)
+
+    def add_episodic_strategy_and_wait(
+        self,
+        memory_id: str,
+        name: str,
+        description: Optional[str] = None,
+        namespaces: Optional[List[str]] = None,
+        reflection_namespaces: Optional[List[str]] = None,
```
abhivr-amazon: same comment as above. This should be required.

src/bedrock_agentcore/memory/client.py
aneeshsabarad: We should leave it to match the same behavior as what is defined in the cli and model, as validation of this will come from the service side, not the client side

src/bedrock_agentcore/memory/client.py
aneeshsabarad: Updated

tests/bedrock_agentcore/memory/test_client.py
```diff
@@ -2851,3 +2851,197 @@ def test_create_or_get_memory_general_exception():
                 raise AssertionError("Exception was not raised")
             except Exception as e:
                 assert "Unexpected error" in str(e)
+
+
```
abhivr-amazon, line 2855: nit: We should also add negative test cases to check for errors being thrown

src/bedrock_agentcore/memory/client.py
aneeshsabarad: Discussed offline and it is changed now

src/bedrock_agentcore/memory/client.py
```diff
@@ -1419,6 +1420,51 @@ def add_user_preference_strategy_and_wait(
         self.add_user_preference_strategy(memory_id, name, description, namespaces)
         return self._wait_for_memory_active(memory_id, max_wait, poll_interval)
 
+    def add_episodic_strategy(
```
aidandaly24, line 1423: nit: The parameter order differs from other strategy methods like `add_user_preference_strategy(memory_id, name, description, namespaces)`. Having
reflection_namespaces (required) between name and description breaks the pattern. The only impact is that the inconsistent API may confuse users familiar with other strategy methods.

aneeshsabarad: Yeah, decided to move reflection_namespaces before description since the required parameters should all come before the optional ones ideally

src/bedrock_agentcore/memory/client.py
```diff
@@ -1482,6 +1528,78 @@ def add_custom_semantic_strategy_and_wait(
...
+            name: Strategy name
+            extraction_config: {"prompt": "...", "modelId": "..."}
+            consolidation_config: {"prompt": "...", "modelId": "..."}
+            reflection_config: {"prompt": "...", "modelId": "...", "namespaces": [...]}
+            description: Optional description
+            namespaces: Optional namespaces list
+        """
+        reflection = {
+            "appendToPrompt": reflection_config["prompt"],
+            "modelId": reflection_config["modelId"],
```
aidandaly24: suggestion: Should we consider adding input validation on `reflection_config["prompt"]` and `reflection_config["modelId"]`? I think this may raise a `KeyError` with an unclear error message if they are missing. We could instead raise a ValueError like:

```
if key not in reflection_config:
        raise ValueError(f"reflection_config missing required key: {key}")
```

aneeshsabarad: Yeah this should be there. Prompt and modelId are required for all extraction, consolidation, and reflection config also so we can add a validation to check all of them like:

```
for config, config_name in [
            (extraction_config, "extraction_config"),
            (consolidation_config, "consolidation_config"),
            (reflection_config, "reflection_config"),
        ]:
            for key in ("prompt", "modelId"):
                if key not in config:
                    raise ValueError(f"{config_name} missing required key: {key}")
```

src/bedrock_agentcore/memory/client.py
aidandaly24: Sounds good to me, I also thought the same but then noticed that odd difference.

src/bedrock_agentcore/memory/client.py
aneeshsabarad: Also will move reflection to be in the strategy block to read cleaner like: 

```
strategy = {
            StrategyType.CUSTOM.value: {
                "name": name,
                "configuration": {
                    "episodicOverride": {
                        "extraction": {
                            "appendToPrompt": extraction_config["prompt"],
                            "modelId": extraction_config["modelId"],
                        },
                        "consolidation": {
                            "appendToPrompt": consolidation_config["prompt"],
                            "modelId": consolidation_config["modelId"],
                        },
                        "reflection": {
                            "appendToPrompt": reflection_config["prompt"],
                            "modelId": reflection_config["modelId"],
                            **(
                                {"namespaces": reflection_config["namespaces"]}
                                if "namespaces" in reflection_config
                                else {}
                            ),
                        },
                    }
                },
            }
        }
```

tests/bedrock_agentcore/memory/test_client.py
jariy17: We would like negative tests too

src/bedrock_agentcore/memory/client.py
```diff
@@ -1820,10 +1951,16 @@ def _wrap_configuration(
                     wrapped_config["extraction"] = {EXTRACTION_WRAPPER_KEYS[strategy_type_enum]: extraction}
                 elif strategy_type == "CUSTOM" and override_type:
                     override_enum = OverrideType(override_type)
-                    if override_type in ["SEMANTIC_OVERRIDE", "USER_PREFERENCE_OVERRIDE"]:
+                    if override_enum in CUSTOM_EXTRACTION_WRAPPER_KEYS:
                         wrapped_config["extraction"] = {
                             "customExtractionConfiguration": {CUSTOM_EXTRACTION_WRAPPER_KEYS[override_enum]: extraction}
                         }
+            elif strategy_type == "CUSTOM" and override_type and any(key in extraction for key in custom_override_keys):
```
jariy17: It's a little messy here to understand. What does this part of code achieve?

aneeshsabarad: 1) None of the other memory strategies have negative tests, so I just followed the same structure for the strategy related tests in this one. Should we still add negative tests just for episodic?

2) Refactoring this to be simpler, the wrapping was off before itself but it should be fine now. Tested it to match the behavior of the bedrock-agentcore-control cli.

src/bedrock_agentcore/memory/constants.py
```diff
@@ -86,12 +90,18 @@ class MessageRole(Enum):
 CUSTOM_EXTRACTION_WRAPPER_KEYS: Dict[OverrideType, str] = {
```
jariy17, line 90: nit: OVERRIDE_EXTRACTION_WRAPPERS_KEYS

src/bedrock_agentcore/memory/client.py
```diff
@@ -1811,19 +1939,22 @@ def _wrap_configuration(
...
+                override_enum = OverrideType(override_type)
+                if override_enum in CUSTOM_EXTRACTION_WRAPPER_KEYS:
+                    wrapped_config["extraction"] = {
+                        "customExtractionConfiguration": {CUSTOM_EXTRACTION_WRAPPER_KEYS[override_enum]: extraction}
+                    }
+                else:
+                    wrapped_config["extraction"] = extraction
+            elif any(key in extraction for key in builtin_config_keys):
+                strategy_type_enum = MemoryStrategyTypeEnum(strategy_type)
+                if strategy_type in ("SEMANTIC", "USER_PREFERENCE"):
```
jariy17, line 1954: Does episodic strategy has a special wrapped_config dictionary like SEMANTIC and USER_PREFERENCES

jariy17 (commented):
Just need to address non-nit comments.

src/bedrock_agentcore/memory/client.py
aneeshsabarad: Ran local integration tests to verify the behavior on update memory operations match what is expected from the service, and the wrapped values are accepted

jariy17 (approved):

PR #209, opened 2026-01-09
fix(memory): Improve pagination behavior in get_last_k_turns() and list_messages()

### LINKED ISSUE #206: Improve pagination behavior in get_last_k_turns() and list_messages()
**Is your feature request related to a problem? Please describe.**
Two related pagination issues affect developer experience:

1. `get_last_k_turns(k=200)` - Requesting 200 turns only fetches 100 events (default `max_results`), so users don't get the expected number of turns. `k` and `max_results` are independent parameters, which is unintuitive.

```
# User expects 200 turns, but only gets turns found within 100 events
turns = client.get_last_k_turns(k=200)  # max_results=100 by default
```

2. `list_messages(limit=None)` - Returns 100 messages instead of all messages. Users must guess a "large enough" number to fetch complete history.

```
# Expected: all messages, Actual: 100 messages
messages = session_repository.list_messages()

# Current workaround - arbitrary large number
messages = session_repository.list_messages(limit=10000)
```

**Describe the solution you'd like**
1. `get_last_k_turns(k)` in `memory/session.py`, `memory/client.py`: Auto-calculate `max_results` based on `k` if not explicitly specified

```
def get_last_k_turns(self, ..., k: int = 5, max_results: Optional[int] = None):
    if max_results is None:
        max_results = max(100, k * 3)  # estimate ~2-3 events per turn
    events = self.list_events(max_results=max_results)
      ...
```

2. `list_messages()` in `memory/integrations/strands/session_manager.py`: Add `fetch_all` parameter for explicit intent.
```
messages = list_messages()                # 100 (default, backward compatible)
messages = list_messages(limit=500)       # 500
messages = list_messages(fetch_all=True)  # all messages
```

**Describe alternatives you've considered**
1. Change `limit=None` to mean "fetch all" - Follows common API conventions, but may break existing code that relies on current default behavior (max 100).
2. Increase default `max_results` to a larger value (e.g., 1000) - Simple but doesn't solve the core issue and may cause performance problems.
3. Document current behavior more clearly - Doesn't improve DevX, just shifts burden to users.

**Additional context**
Use cases affected:
 - Session restoration requiring complete conversation history
 - Context compaction / summarization
 - Conversation export
 - Analytics

kevmyung:
## Summary

- **`get_last_k_turns()`** in both `MemoryClient` and `MemorySessionManager` now automatically paginates until `k` turns are found when `max_results` is not specified
- **`list_messages()`** in `AgentCoreMemorySessionManager` adds `fetch_all` parameter to retrieve all messages instead of being limited to 100
- Maintains backward compatibility: existing code with explicit `max_results` parameter behaves the same

## Problem

Previously, `get_last_k_turns(k=200)` would only fetch 100 events (the default `max_results`) regardless of the `k` value requested. This caused issues when users needed more than 100 events worth of conversation turns.

Similarly, `list_messages(limit=None)` was hardcoded to return only 100 messages, making it impossible to retrieve all messages in a session.

### `client.py` & `session.py`
- Changed `max_results` default from `100` to `None`
- When `max_results=None`, automatically paginates until `k` turns are found
- When `max_results` is explicitly provided, respects that limit (backward compatible)

### `strands/session_manager.py`
- Added `fetch_all: bool = False` parameter to `list_messages()`
- When `fetch_all=True`, fetches up to 10,000 messages

### Tests
- Added `test_get_last_k_turns_auto_pagination` - verifies pagination continues until k turns found
- Added `test_get_last_k_turns_explicit_max_results` - verifies backward compatibility
- Added `test_list_messages_fetch_all` - verifies fetch_all parameter works correctly

## Test plan

- [x] Unit tests pass (`pytest tests/bedrock_agentcore/memory/`)
- [ ] Manual integration test with actual Amazon AgentCore Memory

Closes #206

### DIFF
--- a/src/bedrock_agentcore/memory/client.py
+++ b/src/bedrock_agentcore/memory/client.py
@@ -1109,9 +1109,7 @@ class MemoryClient:
         }
 
         if branch_name and branch_name != "main":
-            base_params["filter"] = {
-                "branch": {"name": branch_name, "includeParentBranches": include_branches}
-            }
+            base_params["filter"] = {"branch": {"name": branch_name, "includeParentBranches": include_branches}}
 
         try:
             turns: List[List[Dict[str, Any]]] = []
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -519,7 +519,8 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
             )
             if retrieval_config.relevance_score:
                 memories = [
-                    m for m in memories
+                    m
+                    for m in memories
                     if m.get("relevanceScore", retrieval_config.relevance_score) >= retrieval_config.relevance_score
                 ]
             context_items = []
--- a/src/bedrock_agentcore/memory/session.py
+++ b/src/bedrock_agentcore/memory/session.py
@@ -807,9 +807,7 @@ class MemorySessionManager:
         }
 
         if branch_name and branch_name != "main":
-            base_params["filter"] = {
-                "branch": {"name": branch_name, "includeParentBranches": include_parent_branches}
-            }
+            base_params["filter"] = {"branch": {"name": branch_name, "includeParentBranches": include_parent_branches}}
 
         try:
             turns: List[List[EventMessage]] = []
--- a/tests/bedrock_agentcore/memory/test_session.py
+++ b/tests/bedrock_agentcore/memory/test_session.py
@@ -2283,11 +2283,13 @@ class TestEdgeCases:
             # Mock _data_plane_client.list_events with many events
             events = []
             for i in range(10):  # Create 10 events, each with a USER message (new turn)
-                events.append({
-                    "eventId": f"event-{i}",
-                    "eventTimestamp": f"2023-01-01T10:{i:02d}:00Z",
-                    "payload": [{"conversational": {"role": "USER", "content": {"text": f"Message {i}"}}}],
-                })
+                events.append(
+                    {
+                        "eventId": f"event-{i}",
+                        "eventTimestamp": f"2023-01-01T10:{i:02d}:00Z",
+                        "payload": [{"conversational": {"role": "USER", "content": {"text": f"Message {i}"}}}],
+                    }
+                )
 
             manager._data_plane_client.list_events.return_value = {"events": events}
 
--- a/tests/bedrock_agentcore/memory/test_client.py
+++ b/tests/bedrock_agentcore/memory/test_client.py
@@ -3075,9 +3075,7 @@ def test_get_last_k_turns_auto_pagination():
         ]
 
         # Request 2 turns without max_results - should paginate automatically
-        turns = client.get_last_k_turns(
-            memory_id="mem-123", actor_id="user-123", session_id="session-456", k=2
-        )
+        turns = client.get_last_k_turns(memory_id="mem-123", actor_id="user-123", session_id="session-456", k=2)
 
         assert len(turns) == 2
         assert mock_gmdp.list_events.call_count == 2


src/bedrock_agentcore/memory/session.py
```diff
@@ -786,52 +786,82 @@ def get_last_k_turns(
...
-                max_results=max_results,
-            )
-
-            if not events:
-                return []
-
-            # Process events to group into turns
             turns = []
             current_turn = []
+            next_token = None
```
aidandaly24, line 815: suggestion (non-blocking): This pagination logic (lines 805-857) is nearly identical to `client.py` lines 1108-1157. Consider extracting into a shared helper to reduce duplication and maintenance burden. Future bug fixes would need to be applied to both locations otherwise.

tests/bedrock_agentcore/memory/test_session.py
```diff
@@ -3854,3 +3854,47 @@ def test_list_branches_with_pagination(self):
             # Verify nextToken was passed in second call
             second_call_args = mock_client_instance.list_events.call_args_list[1][1]
             assert second_call_args["nextToken"] == "token-123"
+
+    def test_get_last_k_turns_auto_max_results(self):
+        """Test get_last_k_turns auto-calculates max_results based on k."""
+        with patch("boto3.Session") as mock_boto_client:
+            mock_client_instance = MagicMock()
+            mock_boto_client.return_value = mock_client_instance
```
aidandaly24, line 3824: issue: This test mocks `manager.list_events`, but the new implementation in `session.py` no longer calls `list_events()` - it directly calls `self._data_plane_client.list_events()`. This mock won't intercept the actual calls.

Should mock `manager._data_plane_client.list_events` instead, similar to how `test_client.py` mocks `client.gmdp_client.list_events`.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -448,7 +455,13 @@ def list_messages(
             raise SessionException(f"Session ID mismatch: expected {self.config.session_id}, got {session_id}")
 
         try:
-            max_results = (limit + offset) if limit else 100
+            if fetch_all:
+                max_results = 10000
```
aidandaly24: nit: The value 10000 is a magic number. Consider extracting to a named constant (e.g., `MAX_FETCH_ALL_RESULTS = 10000`) for clarity and maintainability.

src/bedrock_agentcore/memory/session.py
```diff
@@ -786,52 +786,82 @@ def get_last_k_turns(
         k: int = 5,
         branch_name: Optional[str] = None,
         include_parent_branches: bool = False,
```
aidandaly24, line 788: thought: client.py uses `include_branches` while session.py uses `include_parent_branches`. This inconsistency predates this PR but is worth noting for future cleanup.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -427,7 +428,13 @@ def update_message(self, session_id: str, agent_id: str, session_message: Sessio
         )
 
     def list_messages(
-        self, session_id: str, agent_id: str, limit: Optional[int] = None, offset: int = 0, **kwargs: Any
+        self,
+        session_id: str,
+        agent_id: str,
+        limit: Optional[int] = None,
+        offset: int = 0,
+        fetch_all: bool = False,
```
jariy17: For `fetch_all`, we aren't really fetching all issue, just whatever MAX_FETCH_ALL_RESULTS is. Therefore, we should just remove this field.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -448,7 +456,13 @@ def list_messages(
             raise SessionException(f"Session ID mismatch: expected {self.config.session_id}, got {session_id}")
 
         try:
-            max_results = (limit + offset) if limit else 100
```
jariy17, line 451: Let's keep this line but have MAX_FETCH_ALL_RESULTS which is 10000

src/bedrock_agentcore/memory/pagination.py
```diff
@@ -0,0 +1,77 @@
+"""Pagination utilities for AgentCore Memory SDK."""
+
+import logging
+from typing import Any, Callable, Dict, List, Optional
+
+logger = logging.getLogger(__name__)
+
+
+def paginate_turns(
```
jariy17: I don't really seeing anyone using this function besides MemoryClient and Session class. I would either move into either class or both classes.

src/bedrock_agentcore/memory/client.py
```diff
@@ -1087,56 +1087,43 @@ def get_last_k_turns(
         session_id: str,
         k: int = 5,
         branch_name: Optional[str] = None,
-        include_branches: bool = False,
-        max_results: int = 100,
+        include_parent_branches: bool = False,
```
jariy17: If include_branches is already commited in the code, renaming this will introduce a breaking change

PR #210, opened 2026-01-12
Add integration_source parameter for framework attribution telemetry

sundargthb:
## Summary

Adds optional `integration_source` parameter to SDK clients (CodeInterpreter, BrowserClient, MemoryClient) to enable attribution telemetry for tracking customer acquisition from framework integrations.

Currently, there's no way to distinguish whether a customer discovered AgentCore through LangChain docs, CrewAI, or direct AWS documentation.

## Solution

Use **User-Agent header modification** via botocore's `Config(user_agent_extra=...)` to append integration source information to all HTTP requests:
```
User-Agent: ... bedrock-agentcore-python/1.0.0 (integration_source=langchain)
```

This approach:
- ✅ Avoids API contract changes
- ✅ Leverages existing AWS request logging infrastructure
- ✅ Is 100% backward compatible (parameter defaults to `None`)

## Changes

| File | Change |
|------|--------|
| `_utils/user_agent.py` | **NEW** - Utility function `build_user_agent_suffix()` |
| `tools/browser_client.py` | Add `integration_source` param to `__init__` |
| `tools/code_interpreter_client.py` | Add `integration_source` param to `__init__` |
| `memory/client.py` | Add `integration_source` param to `__init__` |

## Usage
```python
# Direct SDK usage (optional - for custom integrations)
from bedrock_agentcore.tools.code_interpreter_client import CodeInterpreter

client = CodeInterpreter(
    region="us-west-2",
    integration_source="langchain"  # or "crewai", "strands", etc.
)

# Existing code continues to work unchanged (backward compatible)
client = CodeInterpreter(region="us-west-2")
```

## Backward Compatibility

✅ **No breaking changes.** The `integration_source` parameter is optional with a default value of `None`. All existing code works without modification.

### DIFF
--- a/src/bedrock_agentcore/_utils/user_agent.py
+++ b/src/bedrock_agentcore/_utils/user_agent.py
@@ -34,7 +34,7 @@ def build_user_agent_suffix(integration_source: Optional[str] = None) -> str:
 
     if integration_source:
         # Sanitize to prevent header injection
-        sanitized = "".join(c for c in integration_source if c.isalnum() or c in "-_")
+        sanitized = "".join(c for c in integration_source.lower() if c.isalnum() or c in "-_")
         return f"{base} (integration_source={sanitized})"
 
     return base


src/bedrock_agentcore/_utils/user_agent.py
```diff
@@ -0,0 +1,40 @@
...
+        >>> build_user_agent_suffix("langchain")
+        'bedrock-agentcore/1.0.0 (integration_source=langchain)'
+        >>> build_user_agent_suffix()
+        'bedrock-agentcore/1.0.0'
+    """
+    base = f"bedrock-agentcore/{SDK_VERSION}"
+
+    if integration_source:
+        # Sanitize to prevent header injection
+        sanitized = "".join(c for c in integration_source if c.isalnum() or c in "-_")
```
notgitika: nit: maybe we can also add a `.lower()` here? that would prevent LangChain and Langchain as separate values.
I assume it would almost always be the same though.

notgitika (dismissed):

notgitika:
LGTM

src/bedrock_agentcore/_utils/user_agent.py
sundargthb: Good catch!!  Added `.lower()`

Issue #190, opened 2025-12-13
Strands integration ignores RetrievalConfig.relevance_score when injecting AgentCore Memory context


nakedible:
### Summary
In the Strands integration, `RetrievalConfig.relevance_score` is defined/configurable but appears to be ignored when retrieving and injecting AgentCore Memory results into `agent.messages`. This leads to low-relevance memories being injected even when a high threshold is configured.

### Where
`src/bedrock_agentcore/memory/integrations/strands/session_manager.py`
- `AgentCoreMemorySessionManager.retrieve_customer_context(...)` (and its inner retrieval/injection logic)

### Expected behavior
If `RetrievalConfig.relevance_score` is set (e.g. `0.9`), only memory records with `relevanceScore >= relevance_score` should be injected into the `<user_context>...</user_context>` block, mirroring the behavior in:
- `src/bedrock_agentcore/memory/session.py` → `MemorySessionManager._retrieve_memories_for_llm(...)` (client-side filtering by `relevanceScore`)

### Actual behavior
The Strands integration injects whatever the service returns in the `top_k` results (and concatenates their `content.text`), without applying the configured minimum relevance threshold.

### Why this matters
- Injects irrelevant context → worse model outputs
- Increased context size / token usage
- Confusing UX because the option exists (and has a default) but does not affect behavior

### Steps to reproduce
1. Configure Strands with AgentCore Memory session manager and a high threshold:
   ```py
   RetrievalConfig(top_k=10, relevance_score=0.9, ...)
   ```
2. Ensure retrieval returns at least one record with a low `relevanceScore` (e.g. 0.1–0.3) and at least one with a high score (e.g. 0.9+). (This can be done by mocking `memory_client.retrieve_memories(...)` in a unit test.)
3. Send a user message that triggers `retrieve_customer_context`.
4. Observe that the injected `<user_context>` includes the low-score record(s) despite the threshold.

### Suggested fix
Apply the same client-side filtering used by `MemorySessionManager._retrieve_memories_for_llm(...)` inside the Strands retrieval path before building the injected context:
- If `relevance_score` is non-zero, filter memory records by `record.get("relevanceScore", ...) >= relevance_score`.

### Environment
- bedrock-agentcore-sdk-python: v1.1.1 (also reproducible on `main` as of Dec 2025)
- strands-agents: (please fill)
- Python: (please fill)

AndreaSottana:
Is there any update on this issue?
I would be very keen to know

PR #211, opened 2026-01-12
fix: apply relevance_score filtering in Strands integration (#190)

aidandaly24:
**Issue #, if available:** #190

**Description of changes:**

Fixes the Strands integration ignoring `RetrievalConfig.relevance_score` when injecting AgentCore Memory context.

*Problem*: The `retrieve_customer_context()` method was injecting all retrieved memories into `<user_context> `regardless of their relevance score, even when users configured a high threshold.

*Solution*: Added the same client-side filtering logic that already exists in `MemorySessionManager._retrieve_memories_for_llm()`:
- Filter memories where `relevanceScore >= relevance_score` before building injected context
- Fixed incorrect type hint (`AgentCoreMemoryConfig` → `RetrievalConfig`)d

Testing: Added `test_retrieve_customer_context_filters_by_relevance_score` to verify filtering behavior. I also conducted manual testing through debug logs with my own memory instance to make sure that the relevance filtering was working.

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -21,7 +21,7 @@ from typing_extensions import override
 from bedrock_agentcore.memory.client import MemoryClient
 
 from .bedrock_converter import AgentCoreMemoryConverter
-from .config import AgentCoreMemoryConfig
+from .config import AgentCoreMemoryConfig, RetrievalConfig
 
 if TYPE_CHECKING:
     from strands.agent.agent import Agent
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -510,6 +510,11 @@ def retrieve_for_namespace(namespace: str, retrieval_config: AgentCoreMemoryConf
                 query=user_query,
                 top_k=retrieval_config.top_k,
             )
```
jariy17, line 512: Have we explored if Memory's RetrieveMemoryRecords [API](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_RetrieveMemoryRecords.html#API_RetrieveMemoryRecords_ResponseSyntax) can filter out based on relevance_score?

jariy17: RetrieveMemoryRecord doesn't support server side filter. So client-side filtering is fine.

PR #236, opened 2026-01-30
feat(memory): add metadata support to MemoryClient events

jariy17:
## Summary
- Add `metadata` parameter to `create_event`, `create_blob_event`, and `fork_conversation` methods
- Add `event_metadata` filter parameter to `list_events` method for querying events by metadata
- Metadata supports up to 15 key-value pairs with keys 1-128 characters
- Filtering supports `EQUALS_TO`, `EXISTS`, and `NOT_EXISTS` operators

## Test plan
- [x] Added unit tests for `create_event` with metadata
- [x] Added unit tests for `create_blob_event` with metadata  
- [x] Added unit tests for `list_events` with event_metadata filter
- [x] Added unit tests for `list_events` with combined branch and metadata filters
- [x] Added unit tests for `fork_conversation` with metadata
- [x] All 108 unit tests pass
- [x] Verified with real API calls against AgentCore Memory service

### DIFF
--- a/src/bedrock_agentcore/memory/client.py
+++ b/src/bedrock_agentcore/memory/client.py
@@ -35,6 +35,7 @@ from .constants import (
     Role,
     StrategyType,
 )
+from .models.filters import EventMetadataFilter, MetadataValue
 
 logger = logging.getLogger(__name__)
 
@@ -344,6 +345,7 @@ class MemoryClient:
         messages: List[Tuple[str, str]],
         event_timestamp: Optional[datetime] = None,
         branch: Optional[Dict[str, str]] = None,
+        metadata: Optional[Dict[str, MetadataValue]] = None,
     ) -> Dict[str, Any]:
         """Save an event of an agent interaction or conversation with a user.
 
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/memory/client.py
```diff
@@ -771,6 +786,7 @@ def list_events(
         session_id: str,
         branch_name: Optional[str] = None,
         include_parent_branches: bool = False,
+        event_metadata: Optional[List[EventMetadataFilter]] = None,
```
aidandaly24, line 789: nit: The `EventMetadataFilter` TypedDict in `filters.py` defines `operator: OperatorType (enum)`, but the docstring examples correctly show passing the string `"EQUALS_TO"` directly, which is what boto3 actually needs.

This works fine at runtime since Python doesn't enforce TypedDict, and the examples are correct. The root issue is in `filters.py` where operator should probably be typed as `str` rather than `OperatorType`. Worth a follow-up cleanup but def not a blocker for this PR. Should I create an issue for this? Or can we have strands do it?

jariy17: Already created an [issue](https://github.com/aws/bedrock-agentcore-sdk-python/issues/240)

Issue #149, opened 2025-11-03
AgentCoreMemorySessionManager doesn't support Branches, multi-agent, or metadata features


akshseh:
AgentCoreMemorySessionManager doesn't support Branches, multi-agent applications, or "metadata" features from AgentCore Memory.

PR #244, opened 2026-01-31
feat(memory): event metadata state identification, message batching, and redundant sync elimination

### LINKED ISSUE #196: AgentCoreMemorySessionManager Requires access to all actor_id prefixes "agent_*" and "session_*"
**Describe the bug**
AgentCore Memory session manager uses prefixed actor IDs instead of configured actor_id for certain operations, causing IAM access denied errors when using restrictive policies based on the provided actor_id.

**To Reproduce**
1. Configure AgentCoreMemorySessionManager with a specific actor_id (e.g., "tenant123/user456")
2. Create IAM policy that only allows memory operations for that exact actor_id using bedrock-agentcore:actorId condition
3. Initialize the session manager in your application
4. Attempt to create or read session/agent data
5. See AccessDeniedException because the library internally uses "session_*" and "agent_*" prefixed actor IDs

**Expected behavior**
The session manager should use the configured actor_id for all memory operations, or clearly document that additional actor_id patterns need to be allowed in IAM policies when using bedrock-agentcore:actorId conditions.

**Screenshots**

**Smartphone (please complete the following information):**

**Additional context**
  The AgentCoreMemorySessionManager class internally uses three different actor_id patterns:
1. Configured actor_id (e.g., "tenant123/user456") - used for message operations
2. "session_" prefix (e.g., "session_abc123") - used in _get_full_session_id() for session operations
3. "agent_" prefix (e.g., "agent_claude") - used in _get_full_agent_id() for agent operations

  Lines of interest in session_manager.py:
  - Line 136: return f"{SESSION_PREFIX}{session_id}"
  - Line 152: return f"{AGENT_PREFIX}{agent_id}"
  - Line 181: Uses session-prefixed actor_id for create_session()
  - Line 252: Uses agent-prefixed actor_id for create_agent()
  - Line 364: Uses configured actor_id for create_message()

  IAM Condition Key Behavior
  The bedrock-agentcore:actorId condition key compares against the actual actor_id sent in each API call:
  - When the library calls create_session(), it sends actorId: "session_<session_id>"
  - When the library calls create_agent(), it sends actorId: "agent_<agent_id>"
  - When the library calls create_message(), it sends actorId: "<your_configured_actor_id>"

IAM policies that only allow your configured actor_id will deny the prefixed calls, causing failures during session manager initialization and agent operations.

Workaround
  Use IAM condition with ForAnyValue:StringLike to allow multiple actor_id patterns:
```
  {
    "Condition": {
      "ForAnyValue:StringLike": {
        "bedrock-agentcore:actorId": [
          "session_*",
          "agent_*",
          "your-configured-actor-id"
        ]
      }
    }
  }
```

### LINKED ISSUE #220: Unclear usage of actor_id in AgentCoreMemorySessionManager implementation.
## Issue
Session manager implementation -> https://github.com/aws/bedrock-agentcore-sdk-python/blob/main/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
states 
```
 event = self.memory_client.gmdp_client.create_event(
            memoryId=self.config.memory_id,
            actorId=self._get_full_session_id(session.session_id),
            sessionId=self.session_id,
            payload=[
                {"blob": json.dumps(session.to_dict())},
            ],
            eventTimestamp=self._get_monotonic_timestamp(),
        )
```
Here the actorId is assigned Session information whereas the actorId should be defined by the AI Developer and not updated/changed by the SessionManager. 

## Potential Solution 
We could potentially make use of AgentCore Memory metadata feature to add the session metadata.

jariy17:
## Summary

- Replaces prefixed actorId approach with event metadata for identifying session and agent state events
- Adds automatic migration for backwards compatibility with existing sessions
- Implements true message batching with configurable `batch_size`, and fixes `events_to_messages()` ordering for batched events
- Skips redundant `sync_agent` API calls via agent state hash tracking, eliminating ~6+ unnecessary API calls per conversation turn (code credit: @kevmyung)

Fixes #220
Fixes #196

### Event Metadata for State Identification
Session and agent state events are now identified using event metadata instead of encoding state type into the actorId, enabling cleaner actor identification.

**Before:**
```python
# Session state used a prefixed actorId
create_event(actorId="session_my-session-id", ...)

# Agent state used a prefixed actorId
create_event(actorId="agent_my-agent-id", ...)
```

**After:**
```python
# Session state uses shared actorId + metadata
create_event(
    actorId="shared-actor-id",
    metadata={"stateType": {"stringValue": "SESSION"}},
    ...
)

# Agent state uses shared actorId + metadata
create_event(
    actorId="shared-actor-id",
    metadata={
        "stateType": {"stringValue": "AGENT"},
        "agentId": {"stringValue": "my-agent-id"},
    },
    ...
)
```

### Backwards Compatibility
Existing sessions created with the old prefixed actorId format continue to work without any code changes.
- Auto-migration detects legacy events on read, creates new events with metadata, and deletes old prefixed events

### Message Batching
Multiple messages can now be grouped into single API calls, reducing the number of requests.
- Add `batch_size` configuration option (1-100, default 1)
- Group conversational messages by session_id into batched API calls
- Send blob messages (>9KB) individually via separate API path
- Clear buffer only after all API calls succeed to prevent data loss
- Fix `events_to_messages()` ordering for batched events by iterating in reverse instead of reversing the flat message list
- Add 6 unit tests for multi-event and batched payload ordering

### Redundant Sync Elimination
The strands framework's `register_hooks` fires `sync_agent` on both `MessageAddedEvent` and `AfterInvocationEvent`, causing 2 API calls (`read_agent` + `create_agent`) per sync — ~6+ redundant calls per turn when agent state hasn't changed.
- Compute a hash of the serialized agent state (excluding `created_at`/`updated_at` timestamps) after each sync
- Compare against the last synced hash before each `sync_agent` call; skip entirely when unchanged
- Seed the hash during `initialize()` so the first hook-triggered sync correctly detects unchanged state
- Hash is NOT updated if the parent sync raises, ensuring the next call retries

## Integration Tests 
Added integration tests. 

## Test plan

- [x] All 129 tests pass (`pytest tests/bedrock_agentcore/memory/integrations/strands/ -v`)
- [x] End-to-end demo with real API completed successfully
- [x] Legacy migration tested with existing sessions
- [x] Batching tests pass: `pytest tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py -k "Batching" -v`
- [x] Converter ordering tests pass: `pytest tests/bedrock_agentcore/memory/integrations/strands/test_bedrock_converter.py -v` (24 tests)
- [x] State hash tracking tests pass: 7 tests covering hash computation, skip/call behavior, failure resilience, and initialization

## Future Work

- **Branching support**: Add support for conversation branching to enable forking sessions (related: #149)

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -3,11 +3,10 @@
 import json
 import logging
 import threading
-import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime, timedelta, timezone
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, Optional, TypeVar
+from typing import TYPE_CHECKING, Any, Optional
 
 import boto3
 from botocore.config import Config as BotocoreConfig


src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -250,13 +316,19 @@ def create_agent(self, session_id: str, session_agent: SessionAgent, **kwargs: A
...
                 {"blob": json.dumps(session_agent.to_dict())},
             ],
             eventTimestamp=self._get_monotonic_timestamp(),
+            metadata={
+                STATE_TYPE_KEY: {"stringValue": StateType.AGENT.value},
+                AGENT_ID_KEY: {"stringValue": session_agent.agent_id},
+            },
         )
+        # Track created agent for eventual consistency handling
+        self._created_agent_ids.add(session_agent.agent_id)
```
akshseh: This is to say that we support multiple agents per session ? Does Strands support that for session managers ? And also, this means we're using metadata to store and retrieve events from one `agent_ID` and not branches.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -280,18 +352,58 @@ def read_agent(self, session_id: str, agent_id: str, **kwargs: Any) -> Optional[
...
+                ),
+                EventMetadataFilter.build_expression(
+                    left_operand=LeftExpression.build(AGENT_ID_KEY),
+                    operator=OperatorType.EQUALS_TO,
+                    right_operand=RightExpression.build(agent_id),
+                ),
+            ]
+
+            # Use retry with backoff to handle eventual consistency
+            events = self._retry_with_backoff(
```
akshseh: Should we do retries for functions like read_agent ? It's causing unnecessary costs, maybe it's just a new agent that's empty.

jariy17: Retries are needed because newly created events aren't immediately visible when filtering by event_metadata. Without retries, we may miss agent state updates.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
jariy17: `_created_agent_ids` is a fallback for when `list_events_by_filter` fails to find an agent due to eventual consistency. Newly created events aren't immediately visible when filtering by metadata - there's a short propagation delay. If the metadata query returns nothing but the `agent_id` is in `_created_agent_ids`, we know the agent was created and proceed accordingly rather than failing in `update_agent`

jariy17: Strands supports multiple agents per session. However, we strongly advise against this to the customer by a log message in 582.

PR #256, opened 2026-02-06
feat(memory): implement true message batching for flush_messages()

jariy17:
## Summary

- Add `batch_size` configuration option for message buffering (1-100, default 1)
- Implement true message batching in `flush_messages()` - groups conversational messages by session_id into single API calls
- Preserve message order within each session's payload (earliest first)
- Use latest timestamp from grouped messages for combined events
- Send blob messages (>9KB) individually via different API path
- Clear buffer only after ALL API calls succeed to prevent data loss
- Fix `events_to_messages()` ordering bug: iterate `reversed(events)` instead of reversing the flat message list after iteration, which scrambled intra-event payload order for batched events
- Add 6 unit tests covering multi-event ordering, batched payload ordering, mixed blob/conversational ordering, and malformed payload resilience

## Test plan

- [x] Run batching tests: `pytest tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py -k "Batching" -v`
- [x] Verify multi-session grouping works correctly
- [x] Verify timestamp selection uses latest timestamp
- [x] Verify partial failure preserves entire buffer
- [x] Verify blob messages sent individually
- [x] Run converter ordering tests: `pytest tests/bedrock_agentcore/memory/integrations/strands/test_bedrock_converter.py -v`
- [x] Verify multi-event chronological ordering
- [x] Verify batched event intra-payload ordering preserved
- [x] Verify malformed payload does not break batch
- [x] All 24 converter tests pass
- [x] All 87 session manager tests pass

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/bedrock_converter.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/bedrock_converter.py
@@ -72,7 +72,7 @@ class AgentCoreMemoryConverter:
             list[SessionMessage]: list of SessionMessage objects.
         """
         messages = []
-        for event in events:
+        for event in reversed(events):
             for payload_item in event.get("payload", []):
                 if "conversational" in payload_item:
                     conv = payload_item["conversational"]
@@ -93,7 +93,7 @@ class AgentCoreMemoryConverter:
                                 logger.error("This is not a SessionMessage but just a blob message. Ignoring")
                     except (json.JSONDecodeError, ValueError):
                         logger.error("Failed to parse blob content: %s", payload_item)
-        return list(reversed(messages))
+        return messages
 
     @staticmethod
     def total_length(message: tuple[str, str]) -> int:
--- a/tests/bedrock_agentcore/memory/integrations/strands/test_bedrock_converter.py
+++ b/tests/bedrock_agentcore/memory/integrations/strands/test_bedrock_converter.py
@@ -8,6 +8,19 @@ from strands.types.session import SessionMessage
 from bedrock_agentcore.memory.integrations.strands.bedrock_converter import AgentCoreMemoryConverter
 
 
+def _make_conversational_event(session_messages):
+    """Build one event with multiple conversational payloads."""
+    payloads = []
+    for sm in session_messages:
+        payloads.append({
+            "conversational": {
+                "content": {"text": json.dumps(sm.to_dict())},
+                "role": sm.message["role"].upper(),
+            }
+        })
+    return {"payload": payloads}
+
+
 class TestAgentCoreMemoryConverter:
     """Test cases for AgentCoreMemoryConverter."""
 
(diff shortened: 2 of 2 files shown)


src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -645,3 +670,94 @@ def initialize(self, agent: "Agent", **kwargs: Any) -> None:
         RepositorySessionManager.initialize(self, agent, **kwargs)
 
     # endregion RepositorySessionManager overrides
+
+    # region Batching support
+
+    def flush_messages(self) -> list[dict[str, Any]]:
```
jariy17: this should be private method.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -645,3 +670,94 @@ def initialize(self, agent: "Agent", **kwargs: Any) -> None:
...
+        """Exit the context manager and flush any pending messages.
+
+        Args:
+            exc_type: Exception type if an exception occurred.
+            exc_val: Exception value if an exception occurred.
+            exc_tb: Exception traceback if an exception occurred.
+        """
+        self.flush_messages()
+
+    # endregion Batching support
```
jariy17, line 800: Why is this a comment?

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -645,3 +670,94 @@ def initialize(self, agent: "Agent", **kwargs: Any) -> None:
...
+        logger.info("Flushed %d messages to AgentCore Memory", len(results))
+        return results
+
+    def pending_message_count(self) -> int:
+        """Return the number of messages pending in the buffer.
+
+        Returns:
+            int: Number of buffered messages waiting to be sent.
+        """
+        with self._buffer_lock:
```
jariy17, line 764: There must be a some atomic library that would incorporate the lock whenever you do operations in _message_buffer.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -645,3 +670,94 @@ def initialize(self, agent: "Agent", **kwargs: Any) -> None:
...
+            messages_to_send = list(self._message_buffer)
+            self._message_buffer.clear()
+
+        if not messages_to_send:
+            return []
+
+        results = []
+        for session_id, messages, is_blob, monotonic_timestamp in messages_to_send:
```
jariy17, line 705: this is not actually buffering right now.

jariy17: you can take all the messages_to_send and put it under messages, if it's not a blob. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-agentcore/client/create_event.html. Make sure the messages are ordered in the earilers to latest when storing in one create_event.

PR #273, opened 2026-02-21
feat: isolate async handler execution on dedicated worker event loop

siwachabhi:
## Summary

- Introduces a dedicated persistent worker event loop in a background daemon thread to isolate all async handler execution from the main uvicorn event loop
- Async handlers containing blocking calls (e.g. `time.sleep`, synchronous HTTP) previously froze `/ping` health checks, causing container termination — this fix ensures `/ping` always responds promptly
- Implements three-way handler dispatch: async generators bridged via `queue.Queue`, regular async via `run_coroutine_threadsafe`, sync via `run_in_threadpool`
- Propagates `contextvars` across event loop boundaries using `copy_context()` + Django asgiref `_restore_context` pattern (Python 3.10+ compatible)

## Test plan

- [x] 18 new tests in `TestWorkerLoopInvocation` covering:
  - Async handler runs on worker thread (not main)
  - Sync handler runs in thread pool (worker loop NOT created)
  - Blocking async handler does NOT block `/ping`
  - Context propagation to async and sync handlers
  - Async generator bridging and streaming
  - Fire-and-forget `create_task` survives handler return
  - Exception propagation from worker loop
  - Lazy initialization and loop reuse
  - Concurrent async invocations
  - `functools.partial` and callable class dispatch
  - `HEALTHY_BUSY` ping status during background tasks
  - End-to-end HTTP tests for sync streaming, async, and async gen streaming
- [x] Updated existing `test_async_handler_runs_on_worker_loop` to assert worker thread isolation
- [x] All 105 runtime tests pass
- [x] Full test suite: 950 passed, 0 failures
- [x] Pre-commit hooks (ruff lint, ruff format, trailing whitespace) all pass

### DIFF
--- a/src/bedrock_agentcore/runtime/app.py
+++ b/src/bedrock_agentcore/runtime/app.py
@@ -5,9 +5,11 @@ Provides a Starlette-based web server that wraps user functions as HTTP endpoint
 
 import asyncio
 import contextvars
+import functools
 import inspect
 import json
 import logging
+import queue
 import threading
 import time
 import uuid
@@ -15,6 +17,7 @@ from collections.abc import Sequence
 from typing import Any, Callable, Dict, Optional
 
 from starlette.applications import Starlette
+from starlette.concurrency import run_in_threadpool
 from starlette.middleware import Middleware
 from starlette.responses import JSONResponse, Response, StreamingResponse
 from starlette.routing import Route, WebSocketRoute
@@ -37,6 +40,30 @@ from .models import (
     PingStatus,
 )
 from .utils import convert_complex_objects
+
+
+def _is_async_callable(obj: Any) -> bool:
+    """Check if obj is async-callable, unwrapping functools.partial."""
+    while isinstance(obj, functools.partial):
+        obj = obj.func
+    return asyncio.iscoroutinefunction(obj) or (callable(obj) and asyncio.iscoroutinefunction(obj.__call__))
+
+
+def _is_async_gen_callable(obj: Any) -> bool:
+    """Check if obj is an async generator function, unwrapping functools.partial."""
+    while isinstance(obj, functools.partial):
+        obj = obj.func
+    return inspect.isasyncgenfunction(obj) or (callable(obj) and inspect.isasyncgenfunction(obj.__call__))
+
+
+def _restore_context(ctx: contextvars.Context) -> None:
+    """Restore context variables from a snapshot (Django asgiref pattern)."""
+    for var, value in ctx.items():
+        try:
+            if var.get() != value:
+                var.set(value)
+        except LookupError:
+            var.set(value)
 
 
 class RequestContextFormatter(logging.Formatter):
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/runtime/app.py
```diff
@@ -39,6 +42,30 @@
 from .utils import convert_complex_objects
 
 
+def _is_async_callable(obj: Any) -> bool:
+    """Check if obj is async-callable, unwrapping functools.partial."""
+    while isinstance(obj, functools.partial):
+        obj = obj.func
+    return asyncio.iscoroutinefunction(obj) or (callable(obj) and asyncio.iscoroutinefunction(obj.__call__))
```
jariy17, line 49: **nit:** This only unwraps `functools.partial` but doesn't follow `__wrapped__` chains set by `functools.wraps`. A decorated async function would be silently misclassified as sync and dispatched to the thread pool instead of the worker loop:

```python
def my_decorator(func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        return func(*args, **kwargs)
    return wrapper

@my_decorator
async def handler(payload):
    return {"ok": True}

# _is_async_callable(handler) → False (checks wrapper, a regular def)
```

**Unverified suggestion** — use `inspect.unwrap()` to follow `__wrapped__` chains (has a built-in 200-hop safety limit to prevent cycles):

```python
def _unwrap(obj: Any) -> Any:
    """Unwrap functools.partial and decorator chains."""
    while isinstance(obj, functools.partial):
        obj = obj.func
    try:
        obj = inspect.unwrap(obj)
    except StopIteration:
        pass  # cycle detected, use as-is
    return obj


def _is_async_callable(obj: Any) -> bool:
    obj = _unwrap(obj)
    return asyncio.iscoroutinefunction(obj) or (
        callable(obj) and asyncio.iscoroutinefunction(obj.__call__)
    )


def _is_async_gen_callable(obj: Any) -> bool:
    obj = _unwrap(obj)
    return inspect.isasyncgenfunction(obj) or (
        callable(obj) and inspect.isasyncgenfunction(obj.__call__)
    )
```

src/bedrock_agentcore/runtime/app.py
```diff
@@ -463,16 +493,92 @@ def run(self, port: int = 8080, host: Optional[str] = None, **kwargs):
 
         uvicorn.run(self, **uvicorn_params)
 
-    async def _invoke_handler(self, handler, request_context, takes_context, payload):
+    def _ensure_worker_loop(self) -> asyncio.AbstractEventLoop:
```
jariy17, line 496: The sync generator returned by this method is iterated by Starlette via `iterate_in_threadpool`, which holds a thread for the entire duration of the stream. The thread spends most of its time blocked on `q.get()` waiting for the next chunk from the worker loop.

Under concurrent streaming workloads (e.g., LLM streaming responses each taking ~20 seconds), this can exhaust the default thread pool (~40 threads) [ref](https://starlette.dev/threadpool/). Once exhausted, all new requests  are queued waiting for a thread to free up, even though the occupied threads are mostly idle waiting on `q.get()`.

siwachabhi: this one is expected though, customers at max have 2 concurrent connections to runtime servers

PR #278, opened 2026-02-23
revert: rename retrieved_memory tag back to user_context

notgitika:
## Summary

Reverts the `<user_context>` → `<retrieved_memory>` tag rename introduced in #271. This tag rename is a breaking change for users who reference `<user_context>` in system prompts or parsing logic.

The core prefill fix from #271 (inserting memory before the last user message) is **not affected** — only the tag name is reverted.

A configurable `context_tag` parameter is proposed in #277 so users can opt in to a custom tag name without breaking existing setups.

## Test plan

- [x] All `<retrieved_memory>` references replaced with `<user_context>` in source + tests
- [x] Unit tests pass
- [ ] Integration tests pass


jariy17 (approved):
LGTM, just check what failed the workflow linter.

Issue #254, opened 2026-02-06
Support injectable message converter in AgentCoreMemorySessionManager


afarntrog:
**Is your feature request related to a problem? Please describe.**
The `AgentCoreMemorySessionManager` currently uses a hardcoded `AgentCoreMemoryConverter` for converting between Strands `SessionMessage` format and Bedrock AgentCore Memory event format. This creates an interoperability gap when users have messages stored in STM using non-Bedrock formats (e.g., OpenAI format).

__Current behavior:__

- Users can store messages in STM in any format (as long as it adheres to the STM payload structure)
- When using `AgentCoreMemorySessionManager` with pre-existing OpenAI-formatted messages, `list_messages()` fails because `AgentCoreMemoryConverter.events_to_messages()` expects Bedrock Converse format

__Current hardcoded usage in `session_manager.py`:__

```python
messages = AgentCoreMemoryConverter.message_to_payload(session_message)  # line ~290
# ...
messages = AgentCoreMemoryConverter.events_to_messages(events)  # line ~350
```

**Describe the solution you'd like**
### Proposed Solution

Make the converter injectable into `AgentCoreMemorySessionManager` to support multiple message formats.

#### 1. Define a Converter Protocol

```python
from typing import Protocol

class MemoryConverter(Protocol):
    """Protocol for converting between Strands messages and STM event payloads."""
    
    @staticmethod
    def message_to_payload(session_message: SessionMessage) -> list[tuple[str, str]]:
        """Convert SessionMessage to STM event payload format."""
        ...
    
    @staticmethod
    def events_to_messages(events: list[dict[str, Any]]) -> list[SessionMessage]:
        """Convert STM events to SessionMessages."""
        ...
    
    @staticmethod
    def exceeds_conversational_limit(message: tuple[str, str]) -> bool:
        """Check if message exceeds conversational payload size limit."""
        ...
```

#### 2. Update `AgentCoreMemorySessionManager` Constructor

```python
def __init__(
    self,
    agentcore_memory_config: AgentCoreMemoryConfig,
    converter: Optional[Type[MemoryConverter]] = None,  # NEW PARAMETER
    region_name: Optional[str] = None,
    boto_session: Optional[boto3.Session] = None,
    boto_client_config: Optional[BotocoreConfig] = None,
    **kwargs: Any,
):
    self.converter = converter or AgentCoreMemoryConverter  # Default to current behavior
    # ...
```

[...]

PR #288, opened 2026-03-01
feat(strands-memory): add converter injection and optional restored-tool filtering

dharamendrak:
Issue #, if available:
  #254

  Description of changes:

  - Added injectable message converter support in AgentCoreMemorySessionManager to remove hardcoded converter dependency and enable custom STM canonical formats.
  - Introduced MemoryConverter protocol with:
    - message_to_payload(...)
    - events_to_messages(...)
    - exceeds_conversational_limit(...)
  - Updated session manager constructor to accept optional converter and default to AgentCoreMemoryConverter for backward compatibility.
  - Replaced hardcoded converter calls in write/read paths with self.converter.
  - Added converter modules under integrations/strands/converters/:
    - openai.py (OpenAI canonical format converter)
    - protocol.py (shared interface and size-limit utility)
  - Added optional config flag filter_restored_tool_context (default False) to strip historical toolUse/toolResult content blocks during restore when enabled.
  - Streamlined converter initialization — default AgentCoreMemoryConverter is now set once in __init__ instead of repeated self.converter or AgentCoreMemoryConverter fallbacks in each method.
  - Added tests for:
    - OpenAI converter behavior
    - Session manager converter injection and restore filtering
  - Updated comments/docstrings to clarify wording as "Strands-native message shape".

  ### Testing
Validated the AgentCoreMemorySessionManager with and without OpenAIConverseConverter   use below snippet
```
from bedrock_agentcore.memory.integrations.strands.converters import OpenAIConverseConverter

  session_manager = AgentCoreMemorySessionManager(
      agentcore_memory_config=config,
      converter=OpenAIConverseConverter,
      region_name="us-west-2",
  )
```

 Listed event from agent core memory: Raw AgentCore event confirming TOOL role is correctly persisted:

```
  {
    "memoryId": "strands_agentcore_memory_20260228_155315-R7WGNA4qCH",
    "actorId": "math_specialist",
    "sessionId": "math_20260304_181716",
    "eventId": "0000001772677043559#7ab31b1a",
    "eventTimestamp": "2026-03-04 18:17:23.559000-08:00",
    "payload": [
      {
        "conversational": {
          "content": {
            "text": "{\"role\": \"tool\", \"tool_call_id\": \"call_cRscVeeLyRbImBK4aQZcyCEG\", \"content\": \"Result: 425\", \"status\": \"success\"}"
          },
          "role": "TOOL"
        }
      }
    ],
    "branch": {
      "name": "main"
    }
  }
  ```

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -122,7 +122,7 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
                Defaults to None.
             **kwargs (Any): Additional keyword arguments.
         """
-        self.converter = converter
+        self.converter = converter or AgentCoreMemoryConverter
         self.config = agentcore_memory_config
         self.memory_client = MemoryClient(region_name=region_name)
         session = boto_session or boto3.Session(region_name=region_name)
@@ -427,12 +427,11 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
             raise SessionException(f"Session ID mismatch: expected {self.config.session_id}, got {session_id}")
 
         # Convert and check size ONCE (not again at flush)
-        converter = self.converter or AgentCoreMemoryConverter
-        messages = converter.message_to_payload(session_message)
+        messages = self.converter.message_to_payload(session_message)
         if not messages:
             return None
 
-        is_blob = converter.exceeds_conversational_limit(messages[0])
+        is_blob = self.converter.exceeds_conversational_limit(messages[0])
 
         # Parse the original timestamp and use it as desired timestamp
         original_timestamp = datetime.fromisoformat(session_message.created_at.replace("Z", "+00:00"))
@@ -556,8 +555,7 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
                 session_id=session_id,
                 max_results=max_results,
             )
-            converter = self.converter or AgentCoreMemoryConverter
-            messages = converter.events_to_messages(events)
+            messages = self.converter.events_to_messages(events)
             if self.config.filter_restored_tool_context:
                 messages = self._filter_restored_tool_context(messages)
             if limit is not None:

### CODE REVIEW

File: src/bedrock_agentcore/memory/integrations/strands/converters/auto.py [hunk 1]
```diff
@@ -0,0 +1,85 @@
+"""Automatic converter selection for mixed-model sessions."""
```
afarntrog: I'm not convinced this is necessary. Although it may be possible that someone will have mixed messages, I see it as a really rare use case so I'm not convinced it's worth the tech debt.

src/bedrock_agentcore/memory/integrations/strands/converters/bedrock.py
```diff
@@ -0,0 +1,7 @@
+"""Strands-native message-shape converter adapter."""
```
afarntrog: I think this should not be necessary as strands is natively bedrock. We can just leave the converter as None

src/bedrock_agentcore/memory/integrations/strands/converters/anthropic.py
```diff
@@ -0,0 +1,200 @@
+"""Anthropic-format converter for AgentCore Memory.
```
afarntrog: I would be in favor of progressively adding these converters only as we need them. I think we can leave this one out for now.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -107,12 +108,20 @@ def __init__(
...
             boto_client_config (Optional[BotocoreConfig], optional): Optional boto3 client configuration.
                Defaults to None.
             **kwargs (Any): Additional keyword arguments.
         """
+        self._auto_converter_enabled = converter == "auto"
+        if self._auto_converter_enabled:
+            self.converter = AutoConverseConverter
+            AutoConverseConverter.set_write_converter(BedrockConverseConverter)
+        else:
+            self.converter = converter or BedrockConverseConverter
```
afarntrog: nit: consider just leaving it as `None` for bedrock (native) format

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -107,12 +108,20 @@ def __init__(
...
             agentcore_memory_config (AgentCoreMemoryConfig): Configuration for AgentCore Memory integration.
+            converter (Optional[type[MemoryConverter]], optional): Converter used for message format transformation.
+                Defaults to BedrockConverseConverter.
             region_name (Optional[str], optional): AWS region for Bedrock AgentCore Memory. Defaults to None.
             boto_session (Optional[boto3.Session], optional): Optional boto3 session. Defaults to None.
             boto_client_config (Optional[BotocoreConfig], optional): Optional boto3 client configuration.
                Defaults to None.
             **kwargs (Any): Additional keyword arguments.
         """
+        self._auto_converter_enabled = converter == "auto"
```
afarntrog: I think this `_auto_converter_enabled` is over engineering for now. consider `yagni`

dharamendrak: I agree, thanks @afarntrog

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
dharamendrak: @afarntrog agree

src/bedrock_agentcore/memory/integrations/strands/converters/bedrock.py
dharamendrak: Removed

src/bedrock_agentcore/memory/integrations/strands/converters/auto.py
dharamendrak: I agree this is case is rare

src/bedrock_agentcore/memory/integrations/strands/converters/anthropic.py
dharamendrak: We can incrementally add it

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -417,11 +427,12 @@ def create_message(
             raise SessionException(f"Session ID mismatch: expected {self.config.session_id}, got {session_id}")
 
         # Convert and check size ONCE (not again at flush)
-        messages = AgentCoreMemoryConverter.message_to_payload(session_message)
+        converter = self.converter or AgentCoreMemoryConverter
```
afarntrog: instead of doing `converter = self.converter or AgentCoreMemoryConverter` here and in each of the below methods, consider just setting it in the `init` function initially:
```
def __init__(
...
self.converter = converter or AgentCoreMemoryConverter
```
and then here and in other methods we can use the `self.converter`

dharamendrak: Great catch, missed from earlier code pattern I had

src/bedrock_agentcore/memory/integrations/strands/bedrock_converter.py
```diff
@@ -8,7 +8,9 @@
 
 logger = logging.getLogger(__name__)
 
-CONVERSATIONAL_MAX_SIZE = 9000
+# Bedrock AgentCore Data Plane conversational payload text max is 100000 chars.
+# Ref: https://docs.aws.amazon.com/cli/latest/reference/bedrock-agentcore/create-event.html
+CONVERSATIONAL_MAX_SIZE = 100000
```
dharamendrak, line 13: @afarntrog Please double check as per doc it's 100000

PR #291, opened 2026-03-03
feat: Add automatic flush for batched messages, on AfterInvocationEvent hook and interval-based periodical flush

padmak30:
*Description of changes:*
Add automatic flush mechanisms for batched messages

Add two automatic flush mechanisms to AgentCoreMemorySessionManager
to ensure messages are persisted without manual intervention:

1. AfterInvocationEvent hook: Automatically flushes buffered messages after agent invocation completes. Only registered when batching is enabled (batch_size > 1) to avoid unnecessary overhead.

2. Interval-based flush: Periodically flushes messages at configurable intervals for long-running agents. Controlled by new optional flush_interval_seconds config parameter (default: None/disabled).

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Any, Optional
 
 import boto3
 from botocore.config import Config as BotocoreConfig
-from strands.hooks import MessageAddedEvent
+from strands.hooks import AfterInvocationEvent, MessageAddedEvent
 from strands.hooks.registry import HookRegistry
 from strands.session.repository_session_manager import RepositorySessionManager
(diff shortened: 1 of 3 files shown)


src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -811,3 +826,70 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
...
+        """Callback executed by the flush timer.
+
+        Flushes the buffer if it contains messages, then reschedules the timer.
+        """
+        try:
+            # Only flush if there are messages in the buffer
+            pending = self.pending_message_count()
+            if pending > 0:
+                logger.debug("Interval flush triggered: %d message(s) pending", pending)
+                self._flush_messages()
```
jariy17, line 868: nit: __flush_messages will return early if the buffer is empty. Therefore, we don't need to precheck the buffer.

Issue #292, opened 2026-03-03
append_message crashes with AttributeError: 'NoneType' object has no attribute 'get' when message has no text content


aws-pablito:
Describe the bug

When using AgentCoreMemorySessionManager with a Strands agent that has tools, the agent crashes on any invocation that triggers a tool call. The MessageAddedEvent fires for tool use/result messages, which contain no plain text content. create_message correctly returns None for these, but append_message doesn't guard against it and immediately calls .get() on the None return value.

Stack trace

File "bedrock_agentcore/memory/integrations/strands/session_manager.py", line 571, in append_message
    session_message = SessionMessage.from_message(message, created_message.get("eventId"))
AttributeError: 'NoneType' object has no attribute 'get'
Root cause

In session_manager.py, create_message returns None when message_to_payload produces an empty list (i.e. the message has no text content after filtering). The caller append_message doesn't check for this:

# line ~570 - missing None guard
created_message = self.create_message(self.session_id, agent.agent_id, SessionMessage.from_message(message, 0))
session_message = SessionMessage.from_message(message, created_message.get("eventId"))  # crashes if None
Suggested fix

created_message = self.create_message(self.session_id, agent.agent_id, SessionMessage.from_message(message, 0))
if created_message is None:
    return  # no text content to persist, skip silently
session_message = SessionMessage.from_message(message, created_message.get("eventId"))
To reproduce

Create an AgentCoreMemorySessionManager and attach it to a Strands Agent with at least one @tool
Send a prompt that triggers the tool
Observe crash on the tool use/result MessageAddedEvent
Environment

bedrock-agentcore: 1.4.1
strands-agents: 1.26.0
Python: 3.12

PR #293, opened 2026-03-03
fix: fix npe in memory session manager when messages have no text content

padmak30:
*Issue #, if available:* https://github.com/aws/bedrock-agentcore-sdk-python/issues/292

*Description of changes:* fix npe in memory session manager when messages have no text content

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -590,6 +590,8 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
             **kwargs: Additional keyword arguments for future extensibility.
         """
         created_message = self.create_message(self.session_id, agent.agent_id, SessionMessage.from_message(message, 0))
+        if created_message is None:
+            return
         session_message = SessionMessage.from_message(message, created_message.get("eventId"))
         self._latest_agent_message[agent.agent_id] = session_message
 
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -590,6 +590,8 @@ def append_message(self, message: Message, agent: "Agent", **kwargs: Any) -> Non
             **kwargs: Additional keyword arguments for future extensibility.
         """
         created_message = self.create_message(self.session_id, agent.agent_id, SessionMessage.from_message(message, 0))
+        if created_message is None:
+            return
         session_message = SessionMessage.from_message(message, created_message.get("eventId"))
         self._latest_agent_message[agent.agent_id] = session_message
```
jariy17, line 596: What are the repercussions of the not storing the latest_agent_message?

Kevanov88: @jariy17 I'm not sure this fix is ideal. While the error no longer appears, agent strands intermittently don't receive the contentBlockDelta events, which are required when streaming the response.

This will likely work fine if you're not streaming the contentBlockDelta chunks as they become available and instead yield or return the final result, but it may cause issues when streaming chunks.

For now I just added a check to see if we received any contentBlockDelta and if not, I yield the full result instead. Not ideal but we have to go in prod soon.

The issue is intermittent and even if give your agent the exact same input, sometimes it will work, sometimes it won't. Basically this fix just obfuscated the real issue.

PR #296, opened 2026-03-04
fix: allow custom HTTP status codes from entrypoint handlers (#284)

### LINKED ISSUE #284: Cannot return a 4xx status code from BedrockAgentCoreApp entrypoint
**Describe the bug**
Only status code 200 or 500 can be sent to a client when you use an `entrypoint` decorated function to handle requests for a `BedrockAgentCoreApp`. There's no way to send a status code in the 400s to indicate invalid user input.

**To Reproduce**
Run this server and send a request to the `/invocations` endpoint with body `{}`
```python
from bedrock_agentcore.runtime import BedrockAgentCoreApp
from starlette.exceptions import HTTPException

app = BedrockAgentCoreApp()

@app.entrypoint
def invoke(payload):
    prompt = payload.get("prompt")
    if prompt is not None:
        return {"message": "a message"}
    else:
        raise HTTPException(status_code=400, detail="Prompt missing")

if __name__ == "__main__":
    app.run()
```
**Expected behavior**
Server should respond with a 400 status code to the client but instead it responds with a 500 status code.

**Version (consider adding this to the template)**
0.1.0

**Additional context**
Starlette supports either throwing an `HTTPException` with a status code or returning an instance of a `Response` or `Response` subclass like `JSONResponse` with a status code. Neither method works here because the endpoint function's return value gets converted to JSON and wrapped in a `JSONResponse` with status code 200 and exceptions are caught and replaced by the framework.

aidandaly24:
### Issue

Closes #284

`@app.entrypoint` handlers could only produce HTTP 200 (success) or 500 (exception). There was no way to return 4xx status codes for invalid input because:
1. All return values were serialized and wrapped in `Response(200)`
2. All exceptions (including `HTTPException`) were caught and converted to `JSONResponse(500)`

### Changes

Two mechanisms for handlers to control HTTP status codes:

**1. Raise `HTTPException`** (the approach from the issue):
```
from starlette.exceptions import HTTPException

@app.entrypoint
def invoke(payload):
   if not payload.get("prompt"):
       raise HTTPException(status_code=400, detail="Prompt missing")
   return {"message": "ok"}
```
**2. Return a `Response` object directly:**
```
from starlette.responses import JSONResponse

@app.entrypoint
def invoke(payload):
   if not payload.get("prompt"):
       return JSONResponse({"error": "Prompt missing"}, status_code=400)
   return {"message": "ok"}
```
Both were mentioned in the issue as expected behavior.

### Logging improvements

- `HTTPException` with 4xx → logs at WARNING
- `HTTPException` with 5xx → logs at ERROR (matches generic exception severity)
- Returning a `Response` with 4xx/5xx → logs at WARNING instead of "Invocation completed successfully"

### What's unchanged

- Returning a dict/list/string → still wrapped in `Response(200)`
- Unhandled exceptions → still `JSONResponse(500)`
- Invalid JSON body → still `JSONResponse(400)`
- Streaming responses → unaffected

### Testing

Verified with 5 scenarios:
| Scenario | Expected | Result |
|----------|----------|--------|
| `raise HTTPException(400)` | 400 | ✅ |
| Return dict normally | 200 | ✅ |
| Return `JSONResponse(404)` | 404 | ✅ |
| `raise HTTPException(500)` | 500 | ✅ |
| `raise ValueError()` | 500 | ✅ |

### DIFF
--- a/tests/bedrock_agentcore/runtime/test_app.py
+++ b/tests/bedrock_agentcore/runtime/test_app.py
@@ -421,6 +421,151 @@ class TestBedrockAgentCoreApp:
         assert response.headers.get("x-custom-mw") == "mw"
 
 
+class TestCustomStatusCodes:
+    """Test that entrypoint handlers can return custom HTTP status codes (#284)."""
+
+    def test_http_exception_returns_custom_status(self):
+        """Test raising HTTPException returns its status code instead of 500."""
+        from starlette.exceptions import HTTPException
+
+        app = BedrockAgentCoreApp()
+
+        @app.entrypoint
+        def handler(payload):
+            raise HTTPException(status_code=400, detail="Prompt missing")
+
+        client = TestClient(app)
+        response = client.post("/invocations", json={})
+
+        assert response.status_code == 400
+        assert response.json() == {"error": "Prompt missing"}
+
+    def test_http_exception_422(self):
+        """Test HTTPException with 422 status code."""
+        from starlette.exceptions import HTTPException
+
+        app = BedrockAgentCoreApp()
+
+        @app.entrypoint
+        def handler(payload):
+            raise HTTPException(status_code=422, detail="Validation failed")
+
+        client = TestClient(app)
+        response = client.post("/invocations", json={})
+
+        assert response.status_code == 422
+        assert response.json() == {"error": "Validation failed"}
+
+    def test_http_exception_500_still_returns_500(self):
+        """Test HTTPException with 500 is not swallowed."""
+        from starlette.exceptions import HTTPException
+
+        app = BedrockAgentCoreApp()
+
+        @app.entrypoint
+        def handler(payload):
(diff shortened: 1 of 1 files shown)


src/bedrock_agentcore/runtime/app.py
```diff
@@ -418,6 +419,18 @@ async def _handle_invocation(self, request):
                 self.logger.info("Returning streaming response (async generator) (%.3fs)", duration)
                 return StreamingResponse(self._stream_with_error_handling(result), media_type="text/event-stream")
 
+            # If handler returned a Starlette Response directly, pass it through.
+            # This lets handlers control status codes (e.g. JSONResponse(data, status_code=404)).
+            if isinstance(result, Response):
```
jariy17, line 424: nit: this could be part of the if, elif statement above.

PR #300, opened 2026-03-06
feat: add EvaluationClient with run() for on-demand session evaluation

jariy17:
## Summary
- Add `EvaluationClient` with `run()` method that collects spans from CloudWatch and calls the evaluate API with level-aware batching (SESSION/TRACE/TOOL_CALL)
- Add internal `_agent_span_collector` package with `CloudWatchAgentSpanCollector` for span collection with retry/polling
- Add optional `query_string` and `end_time` parameters to `CloudWatchSpanHelper` to support collector delegation

## Details
- `run()` accepts evaluator_ids, session_id, and agent_id or log_group_name
- Auto-derives log group as `/aws/bedrock-agentcore/runtimes/{agent_id}-DEFAULT`
- CloudWatch query filters by `attributes.session.id` + `ispresent(scope.name)`
- Auto-batches evaluate requests (max 10 target IDs per request)
- Caches evaluator level lookups via control plane
- Operational logging at INFO/DEBUG levels for debugging

## Test plan
- [ ] Unit tests: `python -m pytest tests/bedrock_agentcore/evaluation/test_client.py -v` (35 tests)
- [ ] Full evaluation suite: `python -m pytest tests/bedrock_agentcore/evaluation/ -v` (111 tests)
- [ ] Manual integration test with real agent (see PR comment for test script)

### DIFF
--- a/src/bedrock_agentcore/evaluation/client.py
+++ b/src/bedrock_agentcore/evaluation/client.py
@@ -232,4 +232,3 @@ class EvaluationClient:
                 if span_id:
                     tool_span_ids.append(span_id)
         return tool_span_ids
-
--- a/tests/bedrock_agentcore/evaluation/test_client.py
+++ b/tests/bedrock_agentcore/evaluation/test_client.py
@@ -324,5 +324,3 @@ class TestExtractToolSpanIds:
 
     def test_empty_spans(self):
         assert EvaluationClient._extract_tool_span_ids([]) == []
-
-


src/bedrock_agentcore/evaluation/_agent_span_collector/agent_span_collector.py
```diff
@@ -0,0 +1,125 @@
...
+                    "Span collection timed out after %ds for session %s (0 spans found)",
+                    self.max_wait_seconds,
+                    session_id,
+                )
+                return spans
+
+            logger.info("No spans found yet, retrying in %ds...", self.poll_interval_seconds)
+            time.sleep(self.poll_interval_seconds)
+
+    def _fetch_spans(self, session_id: str, start_time: datetime, end_time: datetime) -> List[dict]:
```
aidandaly24, line 102: The existing `fetch_spans` in `CloudWatchSpanHelper` filters results through `_is_valid_adot_document()`, but this skips that validation. Could we add the same filtering here
so we don't send malformed span documents to the evaluate API?

src/bedrock_agentcore/evaluation/_agent_span_collector/agent_span_collector.py
```diff
@@ -0,0 +1,125 @@
+"""Span collector abstraction for the evaluation runner."""
```
aidandaly24, line 1: nit: No dedicated unit tests for `CloudWatchAgentSpanCollector`. The polling/timeout/retry logic is non-trivial but only tested indirectly through fully-mocked EvaluationClient tests. Would be good to add direct tests covering: immediate return when spans found, retry behavior, timeout with empty result.

src/bedrock_agentcore/evaluation/_agent_span_collector/agent_span_collector.py
jariy17: Just updated with a PR that will shifting adot filtering on to cloudwatch by adding this the query
```
        | filter ispresent(scope.name)
        | filter ispresent(traceId)
        | filter ispresent(spanId)
``

File: src/bedrock_agentcore/evaluation/_agent_span_collector/agent_span_collector.py (hunk 2 above)
Comment (jariy17) [reply src/bedrock_agentcore/evaluation/_agent_span_collector/agent_span_collector.py:1]: Added them now

File: src/bedrock_agentcore/evaluation/client.py [hunk 3]
```diff
@@ -0,0 +1,254 @@
...
+            return []
+
+        base_input = {"evaluationInput": {"sessionSpans": spans}}
+
+        # Steps 2-4: For each evaluator, look up level, build targets, call API
+        all_results = []
+        for evaluator_id in evaluator_ids:
+            level = self._get_evaluator_level(evaluator_id)
+            logger.info("Evaluating with %s (level=%s)", evaluator_id, level)
+            requests = self._build_requests_for_level(evaluator_id, level, base_input, spans)
```
aidandaly24, line 156: `_build_requests_for_level` raises `ValueError` when spans have no trace/tool IDs, but that exception isn't caught here — only the `evaluate()` call below is wrapped in try/except. So a TRACE evaluator with no trace IDs crashes the entire `run()`, while an API error just logs a warning and continues to the next evaluator. Could we wrap this call in the same try/except, or have `_build_requests_for_level` return `[]` + log a warning instead of raising?

jariy17: You're right, I'm just remove the try and catch from the for loop so if anything fails, the function errors out instead of swallowing the error.

PR #328, opened 2026-03-12
feat: add ResourcePolicyClient for resource-based policy management

Hweinstock:
*Issue #, if available:* N/A

*Description of changes:*

### Problem

The SDK has no support for [resource-based policies](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/resource-based-policies.html), which allow customers to control which principals can invoke and manage their Agent Runtime, Endpoint, and Gateway resources. This is useful for fine-grained cross-account access and OAuth authentication scenarios. Without SDK support, users must drop to raw boto3 calls.

### Solution

Added ResourcePolicyClient in src/bedrock_agentcore/services/resource_policy.py — a thin wrapper around the 3 control plane APIs, following the existing IdentityClient pattern:

- put_resource_policy(resource_arn, policy) — create/update (accepts dict or JSON str)
- get_resource_policy(resource_arn) — get
- delete_resource_policy(resource_arn) — delete

#### Direct passthrough to boto on client
- Pros: minimal code and logic to maintain. 
- Cons: provides little value over using boto directly. Ex. no serialization/deserialization, difficult to extend in future. 
#### Private Class Consumed by primitivates
- Pros: limits the number of classes we expose. 
- Cons: Requires manually wiring up resource based policies into each primitive we want to support it. 

### Testing

- unit tests. 
- integ tests: Verifies e2e behavior. Setup additional RESOURCE_POLICY_TEST_ARN and
RESOURCE_POLICY_TEST_PRINCIPAL secrets.(ran against my dev account). 

### DIFF
--- a/src/bedrock_agentcore/services/resource_policy.py
+++ b/src/bedrock_agentcore/services/resource_policy.py
@@ -1,0 +1,70 @@
+"""Client for managing resource-based policies on Bedrock AgentCore resources."""
+
+import json
+import logging
+from typing import Optional, Union
+
+import boto3
+
+from bedrock_agentcore._utils.endpoints import get_control_plane_endpoint
+
+
+class ResourcePolicyClient:
+    """Client for managing resource-based policies on Bedrock AgentCore resources.
+
+    Resource-based policies control which principals can invoke and manage
+    Agent Runtime, Endpoint, and Gateway resources.
+    """
+
+    def __init__(self, region: str):
+        """Initialize the client for the specified region."""
+        self.region = region
+        self.client = boto3.client(
+            "bedrock-agentcore-control",
+            region_name=region,
+            endpoint_url=get_control_plane_endpoint(region),
+        )
+        self.logger = logging.getLogger("bedrock_agentcore.resource_policy_client")
+
+    def put_resource_policy(self, resource_arn: str, policy: Union[str, dict]) -> dict:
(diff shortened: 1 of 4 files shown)


src/bedrock_agentcore/services/resource_policy.py
```diff
@@ -0,0 +1,70 @@
...
+import json
+import logging
+from typing import Optional, Union
+
+import boto3
+
+from bedrock_agentcore._utils.endpoints import get_control_plane_endpoint
+
+
+class ResourcePolicyClient:
```
jariy17, line 12: Managing an agent runtime requires two clients: the existing runtime client and a resource policy client. Customers who need to attach resource policies to a runtime will need this additional client.

Hweinstock: This is the tradeoff with this approach. If we hide the resource policy functionality behind other clients, the customers can manage an agent runtime with a single client, but we are responsible for wiring that up each time we want to add support for a new primitive or agentcore resource. 

Alternatively, if we create a separate class the customer will need separate clients for the mentioned flow, but it allows for a more extendable and flexible design that doesn't require explicitly wiring up each consumer.

jariy17: I understand why having a separate client. Thank you for weighing both options. I'm fine with merging this client.

Issue #44, opened 2025-08-06
Enhancement: support custom boto3.Session in MemoryClient


Ryo-Suzukii:
**Is your feature request related to a problem? Please describe.**  
I'm always frustrated when I want to specify an AWS profile or use a custom `boto3.Session` to create Bedrock AgentCore clients, but the code always uses the default `boto3.client(...)` calls and doesn’t allow passing in a session. This prevents usage of named profiles or custom session configurations.

**Describe the solution you'd like**  
I’d like to add an optional `boto3_session: boto3.Session | None = None` parameter to `MemoryClient.__init__`. Internally:

```python
class MemoryClient:
    """High-level Bedrock AgentCore Memory client with essential operations."""

    def __init__(self, region_name: Optional[str] = None, boto3_session: boto3.Session | None = None):
        session = boto3_session if boto3_session else boto3.Session()
        """Initialize the Memory client."""
        self.region_name = region_name or session.region_name or "us-west-2"

        self.gmcp_client = session.client("bedrock-agentcore-control", region_name=self.region_name)
        self.gmdp_client = session.client("bedrock-agentcore", region_name=self.region_name)
```

**Describe alternatives you've considered**

Manually patching the class in local code, but that’s tedious and not reusable.

Using environment variables (AWS_PROFILE, AWS_ACCESS_KEY_ID, etc.), but that isn't flexible for CI or code‑based configuration.

**Additional context**
This change would be consistent with other AWS SDK usage patterns where clients accept a session parameter. It improves flexibility and testability, especially for scenarios involving multiple AWS profiles or test harness setups.

PR #330, opened 2026-03-12
feat(memory): add boto3_session parameter to MemoryClient

jariy17:
*Issue #, if available:*
https://github.com/aws/bedrock-agentcore-sdk-python/issues/44

*Description of changes:*
Add optional `boto3_session` parameter to `MemoryClient.__init__()`, enabling named profiles, custom credentials, and other session-level configuration. This makes `MemoryClient` consistent with `MemorySessionManager`, which already accepts `boto3_session`.

Based on #90 by @manikawnth — rebased and updated to work with the current codebase.

## Changes
- Add `boto3_session` parameter to `MemoryClient.__init__()`
- Use `session.client()` instead of `boto3.client()` for creating boto3 clients
- Updated all tests to mock `boto3.Session` instead of `boto3.client`
- Added tests for `boto3_session` parameter (with and without explicit `region_name`)

## Test plan
- [x] All 112 unit tests pass
- [x] New tests for custom `boto3_session` usage

### DIFF
--- a/src/bedrock_agentcore/memory/client.py
+++ b/src/bedrock_agentcore/memory/client.py
@@ -65,17 +65,32 @@ class MemoryClient:
         "list_memory_strategies",
     }
 
-    def __init__(self, region_name: Optional[str] = None, integration_source: Optional[str] = None):
-        """Initialize the Memory client."""
-        self.region_name = region_name or boto3.Session().region_name or "us-west-2"
+    def __init__(
+        self,
+        region_name: Optional[str] = None,
+        integration_source: Optional[str] = None,
+        boto3_session: Optional[boto3.Session] = None,
+    ):
+        """Initialize the Memory client.
+
+        Args:
+            region_name: AWS region name. If not provided, uses the session's region or "us-west-2".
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/memory/client.py
```diff
@@ -65,17 +65,32 @@ class MemoryClient:
...
         # Build config with user-agent for telemetry
         user_agent_extra = build_user_agent_suffix(integration_source)
         client_config = Config(user_agent_extra=user_agent_extra)
 
-        self.gmcp_client = boto3.client("bedrock-agentcore-control", region_name=self.region_name, config=client_config)
-        self.gmdp_client = boto3.client("bedrock-agentcore", region_name=self.region_name, config=client_config)
+        self.gmcp_client = session.client(
+            "bedrock-agentcore-control", region_name=self.region_name, config=client_config
+        )
+        self.gmdp_client = session.client("bedrock-agentcore", region_name=self.region_name, config=client_config)
```
Hweinstock, line 93: does it make sense to inject client configuration as well?

PR #339, opened 2026-03-13
feat(strands-memory): add event metadata support to AgentCoreMemorySessionManager

tejaskash:
## Summary

Adds user-supplied event metadata support to `AgentCoreMemorySessionManager` (Phase 1 of #149).

**Static metadata:**
- New `default_metadata` field on `AgentCoreMemoryConfig` — attaches custom key-value metadata to every message event

**Dynamic metadata (for traceId / Langfuse integration):**
- New `metadata_provider` field — a callable invoked at each event creation, so it can return per-invocation values (e.g. current traceId). This is needed because Strands controls the `append_message` → `create_message` call path, so users can't pass per-call kwargs through `agent()`.
- Merge precedence: `default_metadata` < `metadata_provider()` < per-call `metadata` kwarg < internal keys

**Infrastructure:**
- `_build_metadata()` helper with validation: rejects reserved keys (`stateType`, `agentId`), enforces 15-key API limit
- Refactors internal `_message_buffer` from raw tuple to `BufferedMessage` NamedTuple for clarity and extensibility
- Metadata flows through both immediate-send and batched flush paths

### Usage example (Langfuse traceId)

```python
from langfuse.decorators import langfuse_context

config = AgentCoreMemoryConfig(
    memory_id=MEM_ID,
    session_id=SESSION_ID,
    actor_id=ACTOR_ID,
    metadata_provider=lambda: {
        "traceId": {"stringValue": langfuse_context.get_current_trace_id() or ""}
    },
)
sm = AgentCoreMemorySessionManager(agentcore_memory_config=config, region_name="us-east-1")
agent = Agent(session_manager=sm)
agent("Hello!")  # Event gets the current traceId automatically
```

## Test plan

- [x] 11 new unit tests in `TestMetadataSupport` (default metadata, per-call, merge precedence, reserved keys, max keys, no-metadata, batched, blob, provider called per event, provider merge with defaults, provider reserved keys rejected)
- [x] 3 new integration tests with positive/negative filter assertions (metadata round-trip, session resume, dynamic traceId with disjoint event sets)
- [x] 123 existing unit tests pass unchanged (buffer tuple → `BufferedMessage` migration)
- [x] Full suite: 1098 tests pass, 0 failures

## Related

- Phase 1 of #149
- Phase 2 (branches) is parked pending strands-agents/sdk-python#1228
- Phase 3 (multi-agent) will follow as a separate PR

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/config.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/config.py
@@ -2,7 +2,7 @@
 
 from typing import Any, Callable, Dict, Optional
 
-from pydantic import BaseModel, ConfigDict, Field, field_validator
+from pydantic import BaseModel, Field, field_validator
 
 
 def normalize_metadata(raw: Dict[str, Any]) -> Dict[str, Any]:
@@ -52,8 +52,6 @@ class AgentCoreMemoryConfig(BaseModel):
             dict is merged after default_metadata but before per-call metadata.
             Accepts plain strings (auto-wrapped) or explicit MetadataValue dicts.
     """
-
-    model_config = ConfigDict(arbitrary_types_allowed=True)
 
     memory_id: str = Field(min_length=1)
     session_id: str = Field(min_length=1)


src/bedrock_agentcore/memory/integrations/strands/config.py
```diff
@@ -38,6 +38,9 @@ class AgentCoreMemoryConfig(BaseModel):
             Default is "user_context".
         filter_restored_tool_context: When True, strip historical toolUse/toolResult blocks from
             restored messages before loading them into Strands runtime memory. Default is False.
+        default_metadata: Optional default metadata key-value pairs to attach to every message event.
```
jariy17, line 46: Does this actually solve the customer's ask? 
What if they need different metadata for each message event? Also, what exactly do they mean by "message_event" — are they
  referring to memory records, AgentCore Memory events, or individual conversation turns? Are they trying to attach a distinct metadata field to each conversation turn?

src/bedrock_agentcore/memory/integrations/strands/README.md
```diff
@@ -239,6 +240,46 @@ https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/memory-strategies.
...
+
+### Default Metadata (applied to all messages)
+
+```python
+config = AgentCoreMemoryConfig(
+    memory_id=MEM_ID,
+    session_id=SESSION_ID,
+    actor_id=ACTOR_ID,
+    default_metadata={
+        "project": {"stringValue": "atlas"},
```
Hweinstock: would we be able to build this map on behalf of the customer? It feels very verbose. 

Ex:
`{"project" : "atlas"}` --> `{"project" : { "stringValue": "atlas"}}`

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -46,6 +47,22 @@
...
 AGENT_ID_KEY = "agentId"
 
+# Maximum metadata key-value pairs per event (API limit)
+MAX_METADATA_KEYS = 15
+
+# Reserved internal metadata keys that users cannot override
+RESERVED_METADATA_KEYS = frozenset({STATE_TYPE_KEY, AGENT_ID_KEY})
+
+
+class BufferedMessage(NamedTuple):
```
Hweinstock, line 57: nice, I agree with the decision to add some structure here.

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
```diff
@@ -2984,3 +2986,147 @@ def test_config_flush_interval_validation(self):
...
+        session_manager_with_metadata.create_message("test-session-456", "agent-1", session_message, metadata=per_call)
+
+        call_kwargs = mock_memory_client.create_event.call_args[1]
+        assert call_kwargs["metadata"]["location"] == {"stringValue": "SF"}
+        assert call_kwargs["metadata"]["team"] == {"stringValue": "support"}
+        assert call_kwargs["metadata"]["project"] == {"stringValue": "beta"}
+
+    def test_metadata_reserved_keys_rejected(self, session_manager):
+        """ValueError raised when user metadata contains reserved keys."""
+        from bedrock_agentcore.memory.integrations.strands.session_manager import RESERVED_METADATA_KEYS
```
Hweinstock, line 3050: nit: personally still new to python, but in most languages I'm used to seeing imports at the top unless we have a strong reason not to. lmk if python convention is different.

src/bedrock_agentcore/memory/integrations/strands/config.py
Hweinstock: are we sure this is the interface the customer is looking for? Could we ask them to send an example code block of the support they want?

src/bedrock_agentcore/memory/integrations/strands/config.py
```diff
@@ -48,3 +56,5 @@ class AgentCoreMemoryConfig(BaseModel):
     flush_interval_seconds: Optional[float] = Field(default=None, gt=0)
     context_tag: str = Field(default="user_context", min_length=1)
     filter_restored_tool_context: bool = Field(default=False)
+    default_metadata: Optional[Dict[str, Any]] = None
```
Hweinstock, line 64: Is there a reason we need `Any` here instead of the `MetadataValue` used internally?

src/bedrock_agentcore/memory/integrations/strands/README.md
```diff
@@ -239,6 +241,68 @@ https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/memory-strategies.
...
+agent("Hello!")  # Event gets the current traceId automatically
+```
+
+### Per-call Metadata
+
+You can also pass metadata on individual `create_message` calls. Per-call metadata is merged
+with `default_metadata` and `metadata_provider` (per-call values override both for the same key):
+
+```python
+session_manager.create_message(
```
Hweinstock, line 300: what happens when we flush messages in a batch and the metadata is different on each message? Does all the metadata get merged?

src/bedrock_agentcore/memory/integrations/strands/config.py
tejaskash: Yes, it solves the ask. `metadata_provider` is a callable invoked at each `create_message()`, so each event gets whatever `traceId` is current at that moment. We have an integration test confirming two invocations with different `traceId`s produce disjoint, independently filterable event sets.

Per-turn metadata works out of the box with `batch_size=1` (the default) since each turn = its own event. With `batch_size > 1` multiple turns collapse into one event, so the last `traceId` wins — but that's an inherent tradeoff of batching, not a metadata limitation.

The customer is talking about STM events, not LTM records (those are extracted async and don't carry event metadata).

src/bedrock_agentcore/memory/integrations/strands/README.md
tejaskash: Good call. Added auto-normalization — plain strings are now auto-wrapped:

```python
{"project": "atlas"}  →  {"project": {"stringValue": "atlas"}}
```

Both forms accepted. Updated the README examples to use the simpler format.

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
tejaskash: Moved all inline `from datetime import ...` to the top of the test file. You're right — top-level is the Python convention too.

src/bedrock_agentcore/memory/integrations/strands/config.py
tejaskash: Tried switching to `MetadataValue` (a `TypedDict`) but Pydantic on Python < 3.12 rejects `TypedDict` from `typing` in model fields. Kept `Any` in the type annotation but added a `field_validator` that normalizes plain strings at config construction time, and `normalize_metadata()` at runtime for `metadata_provider` output. So the internal plumbing always gets the right shape regardless of what the user passes.

src/bedrock_agentcore/memory/integrations/strands/README.md
tejaskash: When messages in a batch have different metadata, the metadata dicts are merged (later message's keys override earlier ones for the same key). So with `batch_size > 1`, the last value for each key wins in the combined event. This is documented in the batching tradeoff — with `batch_size=1` (the default) each turn gets its own event with its own metadata, so no merging occurs.

src/bedrock_agentcore/memory/integrations/strands/config.py
tejaskash: Yes — the customer's use case is tagging events with a `traceId` from Langfuse that changes per invocation. `metadata_provider` (a callable) gives them exactly that. We have an integ test that confirms two invocations with different `traceId`s produce disjoint event sets filterable by `list_events`.

src/bedrock_agentcore/memory/integrations/strands/config.py
```diff
@@ -38,8 +43,18 @@ class AgentCoreMemoryConfig(BaseModel):
...
+            Merged with any per-call metadata. Maximum 15 total keys per event (including internal keys).
+            Accepts plain strings (auto-wrapped) or explicit MetadataValue dicts.
+            Example: {"location": "NYC"} or {"location": {"stringValue": "NYC"}}
+        metadata_provider: Optional callable that returns metadata key-value pairs. Called at each
+            event creation, so it can return dynamic values (e.g. current traceId). The returned
+            dict is merged after default_metadata but before per-call metadata.
+            Accepts plain strings (auto-wrapped) or explicit MetadataValue dicts.
     """
 
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
jariy17: By adding this value in the beginning, will this break existing customers.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -169,6 +186,55 @@ def __init__(
...
             self._start_flush_timer()
 
+    def _build_metadata(
+        self,
+        internal_metadata: Optional[Dict[str, MetadataValue]] = None,
+        per_call_metadata: Optional[Dict[str, MetadataValue]] = None,
+    ) -> Optional[Dict[str, MetadataValue]]:
+        """Build merged metadata from config defaults, provider, per-call overrides, and internal keys.
+
+        Merge precedence (highest wins):
```
jariy17, line 196: nit: I wish there was a way to throw a validation error if any of the user defined metadata's key are stateType or agentId.

Issue #346, opened 2026-03-16
bug: incorrect pagination in Memory Session Manager `list_messages`


Hweinstock:
### Before PR #244

State events and conversational events were naturally separated by actorId:

```python
# create_session used a prefixed actorId
create_event(actorId="session_my-session-id", payload=[{"blob": ...}])

# create_agent used a prefixed actorId
create_event(actorId="agent_my-agent-id", payload=[{"blob": ...}])

# create_message used the regular actorId
create_event(actorId="my-actor-id", payload=[{"conversational": ...}])
```

So when list_messages called list_events with actorId="my-actor-id", it only got conversational events back. State events lived under different actorIds and were invisible to this query. max_results counted only
what mattered.

### After PR #244 

Everything now uses the same actorId, with metadata distinguishing state events:

```python
# create_session — same actorId, metadata marks it as state
create_event(
    actorId="my-actor-id",  # ← same as messages now
    metadata={"stateType": {"stringValue": "SESSION"}},
    payload=[{"blob": ...}],
)


# create_agent — same actorId, metadata marks it as state
create_event(
    actorId="my-actor-id",  # ← same as messages now
    metadata={"stateType": {"stringValue": "AGENT"}, "agentId": {"stringValue": "..."}},
    payload=[{"blob": ...}],
)

# create_message — same actorId, no metadata
create_event(actorId="my-actor-id", payload=[{"conversational": ...}])
```

### Potential Issues

`list_messages` in `session_manager.py` fetches events without any metadata filter:

```python
events = self.memory_client.list_events(
    memory_id=self.config.memory_id,
    actor_id=self.config.actor_id,
    session_id=session_id,
    max_results=max_results,  # ← this now includes state events in the count
)
```

Since all events share the same actorId now, this query returns both conversational AND state events. The API returned `max_results` events total, some of which were state events. Those get discarded by the converter, so the final list is shorter than what the caller asked for. 

### Concrete example

A session with 1 session state event, 2 agent state events, and 20 conversational messages. Customer calls list_messages(limit=10):

1. `max_results` = limit + offset = 10 
2. `list_events` returns 10 events: 3 state + 7 conversational
3. `events_to_messages` discards the 3 state events → 7 messages
4. Customer gets 7 messages, expected 10

Customer does not get a pagination token, so they assume 7 is all that exists, and doesn't know there are more. 

### Acceptance Criteria
[...]

PR #347, opened 2026-03-16
fix(memory): paginate list_messages by converted message count, not raw event count

Hweinstock:
*Issue #, if available:* https://github.com/aws/bedrock-agentcore-sdk-python/issues/346

*Description of changes:*

## Problem

After [#244](https://github.com/aws/bedrock-agentcore-sdk-python/pull/244), state events (session/agent blobs) share the same actorId as conversational events, distinguished only by metadata. list_messages passes
max_results to list_events, which counts all raw events — including state events that the converter discards. When a caller passes limit, they can receive fewer messages than requested because state events consume
slots in the result set.

## Solution

Introduce a reusable paginate_for_n_results utility in _utils/pagination.py that paginates an arbitrary API, converting accumulated results after each page, and stops when enough converted items are collected or
pages are exhausted. list_messages now uses this utility to paginate gmdp_client.list_events directly, counting converted messages toward the limit instead of raw events.

## Testing

- 1 new unit test: `test_list_messages_with_limit_skips_state_events` — mocks 2 pages with mixed state + conversational events, asserts limit=5 returns exactly 5 messages
- 1 new unit test: `test_list_messages_with_limit_returns_fewer_when_not_enough` — verifies correct behavior when fewer messages exist than requested
- 1 new integration test: `test_list_messages_limit_excludes_state_events` — creates a real session with multiple agent turns, asserts limit=4 returns exactly 4 messages

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -19,6 +19,7 @@ from strands.types.exceptions import SessionException
 from strands.types.session import Session, SessionAgent, SessionMessage
 from typing_extensions import override
 
+from bedrock_agentcore._utils.pagination import DEFAULT_PAGE_SIZE, paginate_for_n_results
 from bedrock_agentcore.memory.client import MemoryClient
 from bedrock_agentcore.memory.models.filters import (
     EventMetadataFilter,
@@ -596,15 +597,24 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
             raise SessionException(f"Session ID mismatch: expected {self.config.session_id}, got {session_id}")
 
         try:
-            max_results = (limit + offset) if limit else MAX_FETCH_ALL_RESULTS
+            target = (limit + offset) if limit else MAX_FETCH_ALL_RESULTS
 
-            events = self.memory_client.list_events(
-                memory_id=self.config.memory_id,
-                actor_id=self.config.actor_id,
-                session_id=session_id,
-                max_results=max_results,
+            def fetch_page(params: dict) -> tuple[list, str | None]:
+                response = self.memory_client.gmdp_client.list_events(**params)
+                return response.get("events", []), response.get("nextToken")
+
+            messages = paginate_for_n_results(
+                fetch_page=fetch_page,
+                initial_params={
+                    "memoryId": self.config.memory_id,
+                    "actorId": self.config.actor_id,
+                    "sessionId": session_id,
+                    "maxResults": DEFAULT_PAGE_SIZE,
+                    "includePayloads": True,
+                },
+                converter=self.converter.events_to_messages,
+                target_count=target,
             )
-            messages = self.converter.events_to_messages(events)
             if self.config.filter_restored_tool_context:
                 messages = self._filter_restored_tool_context(messages)
             if limit is not None:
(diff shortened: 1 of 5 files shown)


tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
aidandaly24: This test now passes for the wrong reason. It mocks `mock_memory_client.list_events.side_effect = Exception("API Error")`, but the production code now calls `self.memory_client.gmdp_client.list_events(...)` — a different attribute. The intended exception is never raised.

What actually happens: `gmdp_client` is a `Mock()`, so `.list_events()` returns a Mock, `.get("events", [])` returns another Mock, and `all_items.extend(mock_obj)` raises `TypeError: 'Mock' object is not iterable`, which the broad `except Exception` catches. The assertion `len(messages) == 0` passes, but via the wrong error path.

**Fix**: `session_manager.memory_client.gmdp_client.list_events.side_effect = Exception("API Error")`

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -596,15 +597,24 @@ def list_messages(
...
-
-            events = self.memory_client.list_events(
-                memory_id=self.config.memory_id,
-                actor_id=self.config.actor_id,
-                session_id=session_id,
-                max_results=max_results,
+            target = (limit + offset) if limit else MAX_FETCH_ALL_RESULTS
+
+            def fetch_page(params: dict) -> tuple[list, str | None]:
+                response = self.memory_client.gmdp_client.list_events(**params)
```
aidandaly24, line 603: This bypasses `MemoryClient.list_events` and calls `gmdp_client.list_events` directly. `MemoryClient.list_events` wraps the raw API with error handling (catches `ClientError`), parameter normalization, and filter construction — none of that applies here. It also means the session manager now has two different data-plane access patterns: every other method goes through `MemoryClient`, but `list_messages` reaches through to the internal client. If someone later adds logging or error normalization to `MemoryClient.list_events`, this code path silently misses it.

**Recommendation**: Add a paginated variant to `MemoryClient` — e.g., `list_events_page(memory_id, actor_id, session_id, max_results, next_token) -> (events, next_token)` — that wraps the raw API call with the same error handling. Then `fetch_page` calls that instead of reaching through to `gmdp_client`.

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
Hweinstock: Good catch! Let me update that to the proper client.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
Hweinstock: The challenge here is that the `list_events` [wrapper in MemoryClient](https://github.com/Hweinstock/bedrock-agentcore-sdk-python/blob/3489f1bf0eafdaa4ccb5f22158695f99f3692c85/src/bedrock_agentcore/memory/client.py#L734) already paginates, but does so in a way that is incompatible with this fix (still counts agent state towards max results). 

I thought about trying to generalize the existing MemoryClient method, but doing so in a backwards-compatible way is going to be difficult. 

For those reasons, I chose to drop down to the raw client to control the pagination myself.

PR #388, opened 2026-04-02
fix: implement update_message() for guardrail redaction support

notgitika:
## Summary

- Implement `update_message()` in `AgentCoreMemorySessionManager` so that Strands' built-in guardrail redaction (`redact_latest_message()`) works out of the box
- Since AgentCore Memory events are immutable, updates are performed via create-then-delete (same pattern as `update_agent()`)
- Handle the `batch_size > 1` case by replacing messages in the send buffer before they are flushed

## Context

When using Bedrock Guardrails with AgentCore Memory as the Strands session store, `update_message()` was a no-op. This meant guardrail-blocked user messages were persisted unredacted, creating a permanent dead-end conversation — on subsequent turns or reconnect, the guardrail would block again on the persisted offending message.

## Test plan

- [x] `test_update_message` — verifies new event created + old event deleted for persisted messages
- [x] `test_update_message_wrong_session` — session ID mismatch raises `SessionException`
- [x] `test_update_message_no_message_id` — graceful skip when message has no event ID
- [x] `test_update_message_create_fails` — raises `SessionException` on create failure
- [x] `test_update_message_delete_fails` — raises `SessionException` on delete failure
- [x] `test_update_buffered_message` — in-buffer replacement when `batch_size > 1`
- [x] Full test suite passes (140/140)

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -682,9 +682,7 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
 
         new_event_id = new_event.get("eventId") if new_event else None
         if not new_event_id:
-            logger.warning(
-                "create_message did not return an eventId — skipping delete of old event %s", old_message_id
-            )
+            logger.warning("create_message did not return an eventId — skipping delete of old event %s", old_message_id)
             return
 
         # Delete the old event; if this fails, roll back the newly created event
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -628,8 +628,7 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
             Optional[SessionMessage]: The message if found, None otherwise.
 
         Note:
-            This is primarily used internally by the `update_message` method to read
-            the original event before replacing it.
+            This reads a single event by ID from AgentCore Memory.
         """
         result = self.memory_client.gmdp_client.get_event(
             memoryId=self.config.memory_id, actorId=self.config.actor_id, sessionId=session_id, eventId=message_id
@@ -681,6 +680,13 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
             logger.error("Failed to update message in AgentCore Memory: %s", e)
             raise SessionException(f"Failed to update message: {e}") from e
 
+        new_event_id = new_event.get("eventId") if new_event else None
+        if not new_event_id:
+            logger.warning(
+                "create_message did not return an eventId — skipping delete of old event %s", old_message_id
+            )
+            return
+
         # Delete the old event; if this fails, roll back the newly created event
         try:
             self.memory_client.gmdp_client.delete_event(
@@ -695,32 +701,29 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
                 old_message_id,
                 delete_error,
             )
-            new_event_id = new_event.get("eventId") if new_event else None
-            if new_event_id:
-                try:
-                    self.memory_client.gmdp_client.delete_event(
-                        memoryId=self.config.memory_id,
-                        actorId=self.config.actor_id,
-                        sessionId=session_id,
-                        eventId=new_event_id,
-                    )
-                    logger.info("Rolled back new event %s after failed delete of old event", new_event_id)
-                except Exception as rollback_error:
-                    logger.error(
-                        "Rollback failed: could not delete new event %s: %s. Both old (%s) and new events may exist.",
-                        new_event_id,
-                        rollback_error,
-                        old_message_id,
-                    )
+            try:
+                self.memory_client.gmdp_client.delete_event(
+                    memoryId=self.config.memory_id,
+                    actorId=self.config.actor_id,
+                    sessionId=session_id,
+                    eventId=new_event_id,
+                )
+                logger.info("Rolled back new event %s after failed delete of old event", new_event_id)
+            except Exception as rollback_error:
+                logger.error(
+                    "Rollback failed: could not delete new event %s: %s. Both old (%s) and new events may exist.",
+                    new_event_id,
+                    rollback_error,
+                    old_message_id,
+                )
             raise SessionException(
                 f"Failed to update message: could not delete old event: {delete_error}"
             ) from delete_error
 
         # Update _latest_agent_message so it doesn't hold a stale eventId
-        new_event_id = new_event.get("eventId") if new_event else None
         latest_messages = getattr(self, "_latest_agent_message", None)
-        if new_event_id and latest_messages and agent_id in latest_messages:
-            old_latest = latest_messages[agent_id]
+        if latest_messages and agent_id in latest_messages:
+            old_latest = self._latest_agent_message[agent_id]
             if old_latest.message_id == old_message_id:
                 self._latest_agent_message[agent_id] = SessionMessage(
                     message=session_message.message,
@@ -987,6 +990,7 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
 
         with self._message_lock:
             messages_to_send = list(self._message_buffer)
+            self._message_buffer.clear()
 
         if not messages_to_send:
             return []
@@ -1039,11 +1043,10 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
                     event.get("eventId"),
                 )
 
-            # Clear message buffer only after ALL messages succeed
-            with self._message_lock:
-                self._message_buffer.clear()
-
         except Exception as e:
+            # Restore messages to buffer so they aren't lost
+            with self._message_lock:
+                self._message_buffer.extend(messages_to_send)
             logger.error("Failed to flush messages to AgentCore Memory: %s", e)
             raise SessionException(f"Failed to flush messages: {e}") from e
 
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -628,8 +628,8 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
             Optional[SessionMessage]: The message if found, None otherwise.
 
         Note:
-            This should not be called as (as of now) only the `update_message` method calls this method and
-            updating messages is not supported in AgentCore Memory.
+            This is primarily used internally by the `update_message` method to read
+            the original event before replacing it.
         """
         result = self.memory_client.gmdp_client.get_event(
             memoryId=self.config.memory_id, actorId=self.config.actor_id, sessionId=session_id, eventId=message_id
@@ -637,27 +637,58 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
         return SessionMessage.from_dict(result) if result else None
 
     def update_message(self, session_id: str, agent_id: str, session_message: SessionMessage, **kwargs: Any) -> None:
-        """Update message data.
+        """Update message data in AgentCore Memory.
 
-        Note: AgentCore Memory doesn't support updating events,
-        so this is primarily for validation and logging.
+        Since AgentCore Memory events are immutable, this method performs an update by
+        creating a new event with the updated content and deleting the old event.
+        This enables features like guardrail redaction via Strands' redact_latest_message().
+
+        If the message has not yet been persisted (e.g., still in the message buffer when
+        batch_size > 1), the buffered message is replaced in-place instead.
 
         Args:
             session_id (str): The session ID containing the message.
             agent_id (str): The agent ID associated with the message.
-            session_message (SessionMessage): The message to update.
+            session_message (SessionMessage): The message to update (with updated content
+                and the original message_id/eventId).
             **kwargs (Any): Additional keyword arguments.
 
         Raises:
-            SessionException: If session ID doesn't match configuration.
+            SessionException: If session ID doesn't match configuration or update fails.
         """
         if session_id != self.config.session_id:
             raise SessionException(f"Session ID mismatch: expected {self.config.session_id}, got {session_id}")
 
-        logger.debug(
-            "Message update requested for message: %s (AgentCore Memory doesn't support updates)",
-            {session_message.message_id},
-        )
+        old_message_id = session_message.message_id
+
+        # If message hasn't been persisted yet (still in buffer), update it there
+        if old_message_id is None:
+            if self._update_buffered_message(session_message):
+                logger.debug("Updated buffered message (not yet persisted to AgentCore Memory)")
+                return
+            logger.debug("Message has no event ID and was not found in buffer - skipping update")
+            return
+
+        try:
+            # Create a new event with the updated message content
+            updated_message = SessionMessage(
+                message=session_message.message,
+                message_id=0,
+                created_at=session_message.created_at,
+            )
+            self.create_message(session_id, agent_id, updated_message)
+
+            # Delete the old event
+            self.memory_client.gmdp_client.delete_event(
+                memoryId=self.config.memory_id,
+                actorId=self.config.actor_id,
+                sessionId=session_id,
+                eventId=old_message_id,
+            )
+            logger.info("Updated message in AgentCore Memory: replaced event %s", old_message_id)
+        except Exception as e:
+            logger.error("Failed to update message in AgentCore Memory: %s", e)
+            raise SessionException(f"Failed to update message: {e}") from e
 
     def list_messages(
         self,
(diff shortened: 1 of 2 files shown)


src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -617,36 +617,67 @@ def read_message(self, session_id: str, agent_id: str, message_id: int, **kwargs
...
+            # Create a new event with the updated message content
+            updated_message = SessionMessage(
+                message=session_message.message,
+                message_id=0,
+                created_at=session_message.created_at,
+            )
+            self.create_message(session_id, agent_id, updated_message)
+
+            # Delete the old event
+            self.memory_client.gmdp_client.delete_event(
```
Hweinstock, line 686: what happens if create succeeds but delete fails?

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -617,36 +617,67 @@ def read_message(self, session_id: str, agent_id: str, message_id: int, **kwargs
...
+            return
+
+        try:
+            # Create a new event with the updated message content
+            updated_message = SessionMessage(
+                message=session_message.message,
+                message_id=0,
+                created_at=session_message.created_at,
+            )
+            self.create_message(session_id, agent_id, updated_message)
```
Hweinstock: for the common case (no guardrail), do we still need to make these API calls?

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -617,36 +617,67 @@ def read_message(self, session_id: str, agent_id: str, message_id: int, **kwargs
...
         return SessionMessage.from_dict(result) if result else None
 
     def update_message(self, session_id: str, agent_id: str, session_message: SessionMessage, **kwargs: Any) -> None:
-        """Update message data.
+        """Update message data in AgentCore Memory.
 
-        Note: AgentCore Memory doesn't support updating events,
-        so this is primarily for validation and logging.
+        Since AgentCore Memory events are immutable, this method performs an update by
+        creating a new event with the updated content and deleting the old event.
```
Hweinstock, line 643: is there a risk that anyone maintains a pointer/reference to the old eventId? My understanding is that recreating it will generate a completely different ID.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -617,36 +617,67 @@ def read_message(self, session_id: str, agent_id: str, message_id: int, **kwargs
...
+        if old_message_id is None:
+            if self._update_buffered_message(session_message):
+                logger.debug("Updated buffered message (not yet persisted to AgentCore Memory)")
+                return
+            logger.debug("Message has no event ID and was not found in buffer - skipping update")
+            return
+
+        try:
+            # Create a new event with the updated message content
+            updated_message = SessionMessage(
```
noorullr, line 674: message with unredacted content would still exist in memory right? Is that okay?

noorullr: I see we delete old message, why do we need to do that. Basically you are doing on the fly migration which ideally would wnat to avoid? If we check and redact before persisting, wont that be enough?

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: No The `update_message()` message is only called by the Strands framework when a guardrail triggers `redact_latest_message()`. it wont be invoked otherwise.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: That is cleaner yes but the Strands framework persists the message via `append_message` before the guardrail runs, so by the time we know redaction is needed, the event already exists. So I think the suggested method is not as pragmatic.

good news is that we already do redact-before-persist for the buffered case (batch_size > 1),`_update_buffered_message` swaps it in-place before it hits the API. The create+delete is the fallback for when the message was already flushed.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: good point, added rollback logic

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: added logic to update` _latest_agent_message` with the new eventId after a successful replacement, so it doesn't hold a reference to the deleted event

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -628,36 +628,107 @@ def read_message(self, session_id: str, agent_id: str, message_id: int, **kwargs
...
+
+        # Create a new event with the updated message content
+        try:
+            updated_message = SessionMessage(
+                message=session_message.message,
+                message_id=0,
+                created_at=session_message.created_at,
+            )
+            new_event = self.create_message(session_id, agent_id, updated_message)
+        except Exception as e:
```
jariy17, line 680: **P1 — Potential data loss when `create_message` returns `None` or `{}`**

Two scenarios:

1. **`batch_size > 1`**: `create_message()` returns `{}` (buffered, no eventId). The code then proceeds to `delete_event()` on the old message. The new content is sitting in an unflushed buffer. If the process crashes before flush, the message is permanently lost. Rollback is also impossible since `new_event.get("eventId")` returns `None`.

2. **Empty converter payload**: If `create_message()` returns `None` (converter produces empty payload), the code proceeds to delete the old event with no replacement.

**Suggestion**: Guard against both cases:
```python
new_event = self.create_message(session_id, agent_id, updated_message)
if not new_event or not new_event.get("eventId"):
    logger.warning("create_message did not return an eventId — skipping delete of old event %s", old_message_id)
    return
```

jariy17 (commented):
## Detailed Review: PR #388 — `update_message()` for guardrail redaction

Overall this is a well-motivated change that addresses a real gap (guardrail-redacted messages being persisted unredacted). The create-then-delete approach is reasonable given immutable events. However, there are several edge cases and test gaps that should be addressed before merge.

### Summary of Findings

**P1 — Potential Data Loss (2 issues)**
1. When `batch_size > 1` and a *persisted* message is updated, `create_message()` returns `{}` (buffered). The old event is then deleted immediately. If the process crashes before the buffer flushes, the message is lost with no recovery. Rollback is also impossible since `new_event_id` would be `None`.
2. If `create_message()` returns `None` (e.g. converter produces empty payload from redacted content), the code proceeds to delete the old event with no replacement created.

**P2 — Race Condition (1 issue)**
3. `_update_buffered_message` holds `_message_lock` while replacing the entry, but `_flush_messages_only` copies the buffer before clearing it. A flush concurrent with an update could send the old (unredacted) content, then the updated buffer entry is cleared — silently losing the redaction.

**P2 — Fragile Buffer Matching (1 issue)**
4. `_update_buffered_message` matches by role only. Multiple buffered messages with the same role, or two blob messages (`None == None`), could cause the wrong message to be replaced.

**P3 — Code Quality (3 issues)**
5. `read_message` docstring says it's "primarily used internally by `update_message`" but `update_message` never calls `read_message`.
6. `getattr(self, "_latest_agent_message", None)` is unnecessary — attribute is always initialized in parent `__init__`.
7. PR description says "same pattern as `update_agent()`" but `update_agent` does NOT delete the old event.

**Test Gaps (see inline comments on test file)**
- No test for `_latest_agent_message` being updated after successful update
- No test for rollback *success* path (only double-failure is accidentally tested)
- `test_update_buffered_message` checks buffer count but never verifies content actually changed
- `test_update_message_create_fails` doesn't assert `delete_event` was NOT called
- No test for `PersistenceMode.NONE` or `batch_size > 1` with persisted messages
- No test for buffer role mismatch or multi-message-same-role scenarios

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -857,6 +928,44 @@ def initialize(self, agent: "Agent", **kwargs: Any) -> None:
...
+    def _update_buffered_message(self, session_message: SessionMessage) -> bool:
+        """Attempt to update a message that is still in the send buffer.
+
+        When batch_size > 1, messages may not yet be persisted to AgentCore Memory.
+        This method finds the most recent buffered message matching the session_message's
+        content role and replaces it with the updated content.
+
+        Args:
+            session_message (SessionMessage): The message with updated content.
+
```
jariy17, line 940: **P2 — Race condition between buffer update and flush**

`_flush_messages_only` copies the buffer (`list(self._message_buffer)`) then later clears it. If `_update_buffered_message` runs between the copy and the clear:
- The flush sends the **old** (unredacted) content from the copy
- The buffer now contains the **new** (redacted) content
- After flush succeeds, the buffer is cleared — redaction silently lost

The fix would be to hold `_message_lock` around the entire flush-copy-and-clear operation, or to check if the flush path also needs to coordinate with updates.

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
```diff
@@ -496,20 +496,74 @@ def test_read_message_not_found(self, session_manager, mock_memory_client):
...
+            message_id="old_event_123",
+            created_at="2024-01-01T12:00:00Z",
+        )
 
-        # Should not raise any exceptions
         session_manager.update_message("test-session-456", "test-agent-123", message)
 
+        # Verify new event was created
+        mock_memory_client.create_event.assert_called_once()
+
```
jariy17, line 513: **Test gap — content of created event not verified**

`mock_memory_client.create_event.assert_called_once()` verifies the call happened but never inspects what content was passed. The replacement event could contain garbage and this test would still pass.

Suggest adding:
```python
create_kwargs = mock_memory_client.create_event.call_args.kwargs
# or inspect the payload to verify it contains "redacted"
```

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
```diff
@@ -496,20 +496,74 @@ def test_read_message_not_found(self, session_manager, mock_memory_client):
...
+        mock_memory_client.gmdp_client.delete_event.assert_called_once()
+        delete_kwargs = mock_memory_client.gmdp_client.delete_event.call_args.kwargs
+        assert delete_kwargs["eventId"] == "old_event_123"
+        assert delete_kwargs["memoryId"] == "test-memory-123"
+        assert delete_kwargs["actorId"] == "test-actor-789"
+        assert delete_kwargs["sessionId"] == "test-session-456"
+
     def test_update_message_wrong_session(self, session_manager):
         """Test updating a message with wrong session ID."""
         message = SessionMessage(message={"role": "user", "content": [{"text": "Hello"}]}, message_id=1)
```
jariy17, line 524: **Test gap — `_latest_agent_message` update not tested**

The source code explicitly updates `self._latest_agent_message[agent_id]` when the old message_id matches. This is critical for guardrail correctness — subsequent turns must reference the new eventId.

Suggest adding a test that:
1. Pre-populates `session_manager._latest_agent_message["test-agent-123"]` with a `SessionMessage` using `message_id="old_event_123"`
2. Calls `update_message`
3. Asserts `session_manager._latest_agent_message["test-agent-123"].message_id == "new_event_456"`

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
```diff
@@ -496,20 +496,74 @@ def test_read_message_not_found(self, session_manager, mock_memory_client):
...
+        )
+
+        with pytest.raises(SessionException, match="Failed to update message"):
+            session_manager.update_message("test-session-456", "test-agent-123", message)
+
+    def test_update_message_delete_fails(self, session_manager, mock_memory_client):
+        """Test update_message raises SessionException when delete fails."""
+        mock_memory_client.create_event.return_value = {"eventId": "new_event_456"}
+        mock_memory_client.gmdp_client.delete_event.side_effect = Exception("Delete failed")
+
```
jariy17, line 557: **Test gap — rollback not verified**

Setting `delete_event.side_effect = Exception(...)` makes ALL calls fail — including the rollback attempt. This accidentally tests only the double-failure path.

Missing tests:
1. **Rollback succeeds**: Use `side_effect=[Exception("Delete failed"), None]` so the first call (delete old) fails but the second (rollback new) succeeds. Assert `delete_event` called twice.
2. **Current test**: Should assert `delete_event` was called at least twice (once for old, once for rollback) and that the rollback targeted `new_event_456`.

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
```diff
@@ -496,20 +496,74 @@ def test_read_message_not_found(self, session_manager, mock_memory_client):
...
+    def test_update_message_create_fails(self, session_manager, mock_memory_client):
+        """Test update_message raises SessionException when create fails."""
+        mock_memory_client.create_event.side_effect = Exception("API Error")
+
+        message = SessionMessage(
+            message={"role": "user", "content": [{"text": "redacted"}]},
+            message_id="old_event_123",
+            created_at="2024-01-01T12:00:00Z",
+        )
```
jariy17, line 548: **Test gap — `create_fails` should verify no delete attempted**

When `create_event` raises, the code should bail out before calling `delete_event`. Add:
```python
mock_memory_client.gmdp_client.delete_event.assert_not_called()
```

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: Added an extra guard now — we check that create_message actually returned a valid eventId before attempting the delete. If it didn't (e.g. buffered or empty payload), we bail early. The rollback path is also simplified since new_event_id is guaranteed non-None by that point.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: Nope — update_message is only invoked by the Strands framework when a guardrail triggers redact_latest_message(). No guardrail = no call to this method.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: Good call — we update _latest_agent_message with the new eventId after a successful replacement so nothing holds a stale reference. Added a dedicated test for this too.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: No — the old event with unredacted content gets deleted right after the new (redacted) one is created. If the delete fails, we roll back the new event so we don't end up with duplicates.

notgitika: Yeah ideally we'd redact before persisting, and we do for the buffered case (batch_size > 1) — the buffer gets swapped in-place before flush. But for batch_size=1, Strands persists via append_message before the guardrail runs, so by the time we know redaction is needed the event already exists. The create+delete is the fallback for that case.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: fixed we now check for a valid eventId right after create_message returns. If it's None or empty (buffered/empty payload case), we bail early without touching the old event

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
notgitika: Fixed — buffer is now cleared atomically with the snapshot (same lock acquisition). If the flush fails, messages get restored to the buffer. No window for _update_buffered_message to race against a partial flush anymore.

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
notgitika: Added. test now asserts the created event payload contains the redacted content.

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
notgitika: Added test_update_message_updates_latest_agent_message — pre-populates with old eventId, calls update, asserts the new eventId is stored.

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
notgitika: Split into two tests now: test_update_message_delete_fails_rollback_succeeds uses side_effect=[Exception, None] and verifies the rollback call targets the new eventId. test_update_message_delete_fails_rollback_fails covers the double-failure path.

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
notgitika: Added — test_update_message_create_fails now asserts delete_event.assert_not_called() at the end.

jariy17 (approved):

PR #389, opened 2026-04-03
feat: add read_only flag to AgentCoreMemorySessionManager to disable ACM persistence

### LINKED ISSUE #2020: [FEATURE] Add read-only access mode to SessionManager for multi-tenant safety
### Problem Statement

Problem Statement:                                                                                                                                                                                                                        
When a SessionManager is hooked into an Agent, it automatically registers callbacks for both read operations (restoring conversation/state on init) and write operations (persisting every new message, syncing agent state, redacting content). There is no way to decouple these — it's all-or-nothing.                                                                                                                                                                        
                                                                                                                                                                                                                                            
In multi-tenant systems, this is a problem. A common pattern is: load a tenant's shared context or prior conversation into the agent (read), let the agent operate with short-term in-memory messages during the request (in-memory), but never write those ephemeral per-request interactions back to the persistent store (no write). Today this isn't possible without building a custom SessionManager subclass that no-ops every write method.                                 
                                                                                                                                                                                                                                            
The coupling exists in two places:                                                                                                                                                                                                        
1. SessionManager.register_hooks() unconditionally registers callbacks for MessageAddedEvent (append + sync), AfterInvocationEvent (sync), and their multi-agent/bidi equivalents — alongside the read-path AgentInitializedEvent         
2. Agent._run_loop and Agent._execute_event_loop_cycle directly call self._session_manager.redact_latest_message() and self._session_manager.sync_agent() outside of the hook system, bypassing any hook-level filtering.  

### Proposed Solution

Proposed Solution:                                                                                                                                                                                                                        
Add an access_mode parameter to the SessionManager base class with an AccessMode enum (READ_WRITE, READ_ONLY), defaulting to READ_WRITE to preserve current behavior.                                                                     
                                                                                                                                                                                                                                            
Changes:                                                                                                                                                                                                                                  
- Add AccessMode enum and access_mode property to SessionManager.                                                                                                                                                                         
- In register_hooks(), conditionally skip registration of write callbacks (append_message, sync_agent, sync_multi_agent, sync_bidi_agent, append_bidi_message) when access_mode is READ_ONLY. Read callbacks (initialize, initialize_multi_agent, initialize_bidi_agent) always register.                                                                                                                                                                           
- In Agent._run_loop and Agent._execute_event_loop_cycle, guard the direct redact_latest_message() and sync_agent() calls behind not self._session_manager.is_read_only.                                                                  
- Pass access_mode through RepositorySessionManager, FileSessionManager, and S3SessionManager constructors.                                                                                                                               
- Export AccessMode from strands.session.                                                                                                                                                                                                 

Usage:                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                            
  from strands import Agent                                                                                                                                                                                                                 
  from strands.session import S3SessionManager, AccessMode                                                                                                                                                                                  
                                                                                                                                                                                                                                            
  agent = Agent(                                                                                                                                                                                                                            
      session_manager=S3SessionManager(                                                                                                                                                                                                     
          session_id="tenant-123",                                                                                                                                                                                                          
          bucket="my-bucket",                                                                                                                                                                                                               
          access_mode=AccessMode.READ_ONLY,                                                                                                                                                                                                 
      )                                                                                                                                                                                                                                     
  )                                                                                                                                                                                                                                         
  # Agent loads conversation history from S3 but never writes back                                                                                                                                                                          
                                                                                                                                                                                                                                            
 Additional considerations for follow-up:                                                                                                                                                                                                  
- Granular control (e.g., read-only messages but writable state) could be added later as separate mode values.                                                                                                                            
- In read-only mode, guardrail-triggered redactions won't persist — the original unredacted message stays in storage. This should be documented as a known trade-off.                                                                     
- Conversation manager state (e.g., removed_message_count) won't sync in read-only mode, so subsequent session loads may re-read previously trimmed messages. This should also be documented.  

### Use Case

1. Multi-tenant customer support agent — A SaaS platform runs one agent per incoming customer request. The agent loads the tenant's shared knowledge base and prior conversation history from S3 on init, but each request is ephemeral.  You don't want thousands of concurrent request-level interactions polluting the shared tenant session. Read-only mode lets the agent consume context without writing back.                                                                
2. Dry-run / preview mode — During development or QA, you want to test an agent against a real production session to see how it would respond, without actually modifying the stored conversation. A developer points the agent at a production session ID with READ_ONLY and runs prompts against it safely.                                                                                                                                                                  
3. Shared context across agent replicas — In a horizontally scaled deployment, multiple agent instances serve the same tenant concurrently. One designated "writer" agent (or a separate ingestion pipeline) owns the session writes. All other replicas load the session read-only to avoid write conflicts, race conditions, and duplicate messages in storage.        

### Alternatives Solutions

Alternative: `ReadOnlySessionManager` wrapper class                                                                                                                                                                                       
Instead of adding a mode flag to the base class, introduce a standalone wrapper that takes any SessionManager and no-ops all write methods while delegating all read methods to the inner instance.                                       
                                                                                                                                                                                                                                            
  from strands.session import ReadOnlySessionManager, S3SessionManager                                                                                                                                                                      
                                                                                                                                                                                                                                            
  inner = S3SessionManager(session_id="tenant-123", bucket="my-bucket")                                                                                                                                                                     
  agent = Agent(session_manager=ReadOnlySessionManager(inner))                                                                                                                                                                              
                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                            
How it works:                                                                                                                                                                                                                             
- initialize(), initialize_multi_agent(), initialize_bidi_agent() → delegated to inner session manager (reads happen normally)                                                                                                            
- append_message(), sync_agent(), redact_latest_message(), sync_multi_agent(), sync_bidi_agent(), append_bidi_message() → no-op (writes silently skipped)                                                                                 
- register_hooks() is inherited unchanged from SessionManager — all hooks fire, but the write methods do nothing when called  

### Additional Context

_No response_

mohanasudhan:
## Description

Add a `read_only` option to `AgentCoreMemoryConfig` that disables persistence to AgentCore Memory (`create_event` calls) while keeping local session/agent state management and memory injection (LTM retrieval) working.

## Motivation

Resolves https://github.com/strands-agents/sdk-python/issues/2020

The `AgentCoreMemorySessionManager` currently persists agent state **and** injects memories. Customers who only need memory injection into agent context (without persisting agent state) had no way to opt out of persistence. This adds a `read_only=True` flag so existing customers can disable persistence with a single config change.

## Changes

- **`config.py`**: Added `read_only: bool = Field(default=False)` to `AgentCoreMemoryConfig`
- **`session_manager.py`**: Guarded ACM write calls (`create_event`, `delete_event`) behind `read_only` check while preserving local state management (validation, caching, `append_message` tracking)
- **Tests**: 20 new tests covering all read_only behavior

## What's guarded (ACM writes skipped when `read_only=True`)
- `create_session` — skips `create_event`, still validates and returns session
- `create_agent` — skips `create_event`/buffering, still caches `created_at`
- `update_agent` — delegates to `create_agent`, cache lookup still works
- `create_message` — skips `create_event`/buffering, returns `{}` so `append_message` tracks local state
- `_flush_messages_only` / `_flush_agent_states_only` — no-op
- Legacy migration paths — skips `create_event` + `delete_event`, still reads data

## What's unaffected
- All read operations (`read_session`, `read_agent`, `list_messages`)
- LTM retrieval (`retrieve_customer_context`)

## Usage

```python
config = AgentCoreMemoryConfig(
    memory_id="my-mem",
    session_id="sess-1",
    actor_id="user-1",
    read_only=True,  # no writes to AgentCore Memory
)
session_manager = AgentCoreMemorySessionManager(agentcore_memory_config=config)
```

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/config.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/config.py
@@ -1,5 +1,6 @@
 """Configuration for AgentCore Memory Session Manager."""
 
+from enum import Enum
 from typing import Any, Callable, Dict, Optional
 
 from pydantic import BaseModel, Field, field_validator
@@ -8,6 +9,19 @@ from pydantic import BaseModel, Field, field_validator
 def normalize_metadata(raw: Dict[str, Any]) -> Dict[str, Any]:
     """Normalize metadata values: plain strings become {"stringValue": value}."""
     return {k: {"stringValue": v} if isinstance(v, str) else v for k, v in raw.items()}
+
+
+class PersistenceMode(str, Enum):
+    """Controls what gets persisted to AgentCore Memory.
+
+    Attributes:
+        FULL: Persist everything (session, agent state, messages) to AgentCore Memory. Default behavior.
+        NONE: Disable all persistence. Local session/agent state management and memory injection
+            (LTM retrieval) still work, but no create_event calls are made to AgentCore Memory.
+    """
+
+    FULL = "FULL"
+    NONE = "NONE"
 
 
 class RetrievalConfig(BaseModel):
@@ -51,6 +65,9 @@ class AgentCoreMemoryConfig(BaseModel):
             event creation, so it can return dynamic values (e.g. current traceId). The returned
             dict is merged after default_metadata but before per-call metadata.
             Accepts plain strings (auto-wrapped) or explicit MetadataValue dicts.
+        persistence_mode: Controls what gets persisted to AgentCore Memory.
+            FULL (default): persist everything. NONE: disable all persistence while keeping
+            local state management and memory injection working.
     """
 
     memory_id: str = Field(min_length=1)
@@ -63,6 +80,7 @@ class AgentCoreMemoryConfig(BaseModel):
     filter_restored_tool_context: bool = Field(default=False)
     default_metadata: Optional[Dict[str, Any]] = None
     metadata_provider: Optional[Callable[[], Dict[str, Any]]] = None
+    persistence_mode: PersistenceMode = Field(default=PersistenceMode.FULL)
 
     @field_validator("default_metadata", mode="before")
     @classmethod
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -29,7 +29,7 @@ from bedrock_agentcore.memory.models.filters import (
 )
 
 from .bedrock_converter import AgentCoreMemoryConverter
-from .config import AgentCoreMemoryConfig, RetrievalConfig, normalize_metadata
+from .config import AgentCoreMemoryConfig, PersistenceMode, RetrievalConfig, normalize_metadata
 from .converters import MemoryConverter
 
 if TYPE_CHECKING:
@@ -142,6 +142,7 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
         """
         self.converter = converter or AgentCoreMemoryConverter
         self.config = agentcore_memory_config
+        self.read_only = agentcore_memory_config.persistence_mode is PersistenceMode.NONE
         self.memory_client = MemoryClient(region_name=region_name)
         session = boto_session or boto3.Session(region_name=region_name)
         self.has_existing_agent = False
@@ -255,17 +256,21 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
         if session.session_id != self.config.session_id:
             raise SessionException(f"Session ID mismatch: expected {self.config.session_id}, got {session.session_id}")
 
-        event = self.memory_client.gmdp_client.create_event(
-            memoryId=self.config.memory_id,
-            actorId=self.config.actor_id,
-            sessionId=self.session_id,
-            payload=[
-                {"blob": json.dumps(session.to_dict())},
-            ],
-            eventTimestamp=self._get_monotonic_timestamp(),
-            metadata={STATE_TYPE_KEY: {"stringValue": StateType.SESSION.value}},
-        )
-        logger.info("Created session: %s with event: %s", session.session_id, event.get("event", {}).get("eventId"))
+        if not self.read_only:
+            event = self.memory_client.gmdp_client.create_event(
+                memoryId=self.config.memory_id,
+                actorId=self.config.actor_id,
+                sessionId=self.session_id,
+                payload=[
+                    {"blob": json.dumps(session.to_dict())},
+                ],
+                eventTimestamp=self._get_monotonic_timestamp(),
+                metadata={STATE_TYPE_KEY: {"stringValue": StateType.SESSION.value}},
+            )
+            logger.info(
+                "Created session: %s with event: %s", session.session_id, event.get("event", {}).get("eventId")
+            )
+
         return session
 
     def read_session(self, session_id: str, **kwargs: Any) -> Optional[Session]:
@@ -318,14 +323,15 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
             session_data = json.loads(old_event.get("payload", {})[0].get("blob"))
             session = Session.from_dict(session_data)
             # Migrate: create new event with metadata, delete old
-            self.create_session(session)
-            self.memory_client.gmdp_client.delete_event(
-                memoryId=self.config.memory_id,
-                actorId=legacy_actor_id,
-                sessionId=session_id,
-                eventId=old_event.get("eventId"),
-            )
-            logger.info("Migrated legacy session event for session: %s", session_id)
+            if not self.read_only:
+                self.create_session(session)
+                self.memory_client.gmdp_client.delete_event(
+                    memoryId=self.config.memory_id,
+                    actorId=legacy_actor_id,
+                    sessionId=session_id,
+                    eventId=old_event.get("eventId"),
+                )
+                logger.info("Migrated legacy session event for session: %s", session_id)
             return session
 
         return None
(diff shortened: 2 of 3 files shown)


src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -255,17 +256,21 @@ def create_session(self, session: Session, **kwargs: Any) -> Session:
...
-            actorId=self.config.actor_id,
-            sessionId=self.session_id,
-            payload=[
-                {"blob": json.dumps(session.to_dict())},
-            ],
-            eventTimestamp=self._get_monotonic_timestamp(),
-            metadata={STATE_TYPE_KEY: {"stringValue": StateType.SESSION.value}},
-        )
-        logger.info("Created session: %s with event: %s", session.session_id, event.get("event", {}).get("eventId"))
+        if not self.read_only:
```
noorullr: By not persisting anything, it also means you are not persisting any conversation which in turn means that LTM memory is not getting updated, lagging or stale. How would you keep LTM refreshed if nothing is going into memory. My thinking was a flag that disables agent state persistence.

mohanasudhan: Conversation persistent is done in separate layer (not in the agent). In our case, we have (service A) calls -> (service B). Service A does quote, gateway, and writes to memory. Service B is the agent, which only processes the request. We don't want agent to write the memory.

noorullr: My suggestion here would be that rather than a boolean, I would say adding an enum, something like `persistence_mode` so that we can do `none` | `full` but also later add support for `conversation` which means agent state is not persisted.

mohanasudhan: Addressed

noorullr (approved):

src/bedrock_agentcore/memory/integrations/strands/config.py
```diff
@@ -10,6 +11,19 @@ def normalize_metadata(raw: Dict[str, Any]) -> Dict[str, Any]:
     return {k: {"stringValue": v} if isinstance(v, str) else v for k, v in raw.items()}
 
 
+class PersistenceMode(str, Enum):
```
Hweinstock, line 14: I would think the most common use would be in the middle, ex. persist messages (to get LTM generated from those messages), but not agent state.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -142,6 +142,7 @@ def __init__(
         """
         self.converter = converter or AgentCoreMemoryConverter
         self.config = agentcore_memory_config
+        self.read_only = agentcore_memory_config.persistence_mode is PersistenceMode.NONE
```
Hweinstock: nit: should we store the persistence mode directly for easier support with future values?

Hweinstock (dismissed):
Thanks for the change!

Hweinstock:
```
uv-lock..................................................................Passed
ruff (legacy alias)......................................................Failed
- hook id: ruff
- exit code: 1

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py:3262:121: E501 Line too long (122 > 120)
     |
3261 |     def test_config_accepts_none_mode(self):
3262 |         config = AgentCoreMemoryConfig(memory_id="m", session_id="s", actor_id="a", persistence_mode=PersistenceMode.NONE)
     |                                                                                                                         ^^ E501
3263 |         assert config.persistence_mode is PersistenceMode.NONE
     |

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py:3373:121: E501 Line too long (122 > 120)
     |
3371 |     def test_read_agent_works(self, no_persist_manager, mock_memory_client):
3372 |         mock_memory_client.list_events.return_value = [
3373 |             {"eventId": "e1", "payload": [{"blob": '{"agent_id": "a1", "state": {}, "conversation_manager_state": {}}'}]},
     |                                                                                                                         ^^ E501
3374 |         ]
3375 |         result = no_persist_manager.read_agent("test-session-456", "a1")
     |

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py:3384:121: E501 Line too long (169 > 120)
     |
3382 | …
3383 | …
3384 | …"role": "user", "content": [{"text": "Hello"}]}, "message_id": 1}'}, "role": "USER"}}],
     |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ E501
3385 | …
3386 | …
     |

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py:3395:121: E501 Line too long (124 > 120)
     |
3393 |         mock_memory_client.list_events.side_effect = [
3394 |             [],
3395 |             [{"eventId": "legacy-1", "payload": [{"blob": '{"session_id": "test-session-456", "session_type": "AGENT"}'}]}],
     |                                                                                                                         ^^^^ E501
3396 |         ]
3397 |         result = no_persist_manager.read_session("test-session-456")
     |

tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py:3405:121: E501 Line too long (130 > 120)
     |
3403 |         mock_memory_client.list_events.side_effect = [
3404 |             [],
3405 |             [{"eventId": "legacy-1", "payload": [{"blob": '{"agent_id": "a1", "state": {}, "conversation_manager_state": {}}'}]}],
     |                                                                                                                         ^^^^^^^^^^ E501
3406 |         ]
3407 |         result = no_persist_manager.read_agent("test-session-456", "a1")
     |

Found 5 errors.

ruff format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

2 files reformatted, 152 files left unchanged

trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check toml...............................................................Passed
check json...............................................................Passed
check yaml...............................................................Passed
check for merge conflicts................................................Passed
check for added large files..............................................Passed
debug statements (python)................................................Passed
```

small lint errors to address before merging.

Hweinstock (approved):

PR #400, opened 2026-04-09
fix(ci): add pytest-rerunfailures to integration test dependencies

jariy17:
## Summary
- Adds `pytest-rerunfailures` to the integration test CI pip install step
- Without it, `@pytest.mark.flaky(reruns=2, reruns_delay=5)` decorators in integration tests are silently ignored, so flaky tests never retry and pytest emits unknown mark warnings

## Test plan
- [x] Verify `pytest-rerunfailures` is already in dev dependencies in `pyproject.toml`
- [ ] CI integration tests should now properly retry flaky tests instead of failing on first attempt

### DIFF
--- a/.github/workflows/integration-testing.yml
+++ b/.github/workflows/integration-testing.yml
@@ -161,7 +161,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -e .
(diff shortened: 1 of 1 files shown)


Hweinstock (approved):
Nice find!

PR #415, opened 2026-04-14
fix: use correct 'score' field for relevance filtering in retrieve_customer_context

biascia:
## Summary

`AgentCoreMemorySessionManager.retrieve_customer_context` filters retrieved memories using
`m.get("relevanceScore", ...)`, but the `MemoryRecordSummary` shape returned by
`retrieve_memory_records` defines the field as `score` (confirmed via the AWS service model).

The field name mismatch means the fallback value is always used — set to
`retrieval_config.relevance_score` itself — so the condition `relevance_score >= relevance_score`
is always `True`. Every memory passes the filter regardless of the configured threshold.

## Root cause

```python
# Before — "relevanceScore" is never present; fallback makes condition always True
memories = [
    m for m in memories
    if m.get("relevanceScore", retrieval_config.relevance_score) >=
    retrieval_config.relevance_score
]
```

The `MemoryRecordSummary` boto3 service model shape:

```
score: Double   ← actual field name
```

## Fix

```python
# After — filter on the correct field; 0.0 default excludes records without a score
memories = [m for m in memories if m.get("score", 0.0) >= retrieval_config.relevance_score]
```

Using `0.0` as the default ensures records without a score are excluded rather than silently
passed through.

## Changes

- `session_manager.py`: fix field name in relevance filter
- `test_agentcore_memory_session_manager.py`: update mock memory records to use `"score"` to
  match the actual API response shape

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -781,11 +781,7 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
                 top_k=retrieval_config.top_k,
             )
             if retrieval_config.relevance_score:
-                memories = [
-                    m
-                    for m in memories
-                    if m.get("relevanceScore", retrieval_config.relevance_score) >= retrieval_config.relevance_score
-                ]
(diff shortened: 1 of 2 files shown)


notgitika (approved):
LGTM, thanks for adding this

PR #424, opened 2026-04-20
feat: add utility methods for primitive clients

nborges-aws:
*Description of changes:* 
Adding utility methods used in new primitive clients. Releasing utilities first so the PRs adding primitive clients can be done in parallel, as they all take dependency on these methods. The changes in this PR include:

* Config.py -> adds WaitConfig dataclass
* Polling.py -> adds wait_until and wait_until_deleted utility methods used in `*_and_wait` methods across primitives
* snake_case.py -> adds additional snake case conversion helper used across primitives
* test_utils.py -> Unit tests for added utilities

### DIFF
--- a/src/bedrock_agentcore/_utils/config.py
+++ b/src/bedrock_agentcore/_utils/config.py
@@ -1,0 +1,22 @@
+"""Shared configuration dataclasses for SDK clients."""
+
+from dataclasses import dataclass
+
+
+@dataclass
+class WaitConfig:
+    """Configuration for *_and_wait polling methods.
+
+    Args:
+        max_wait: Maximum seconds to wait. Default: 300. Must be >= 1.
+        poll_interval: Seconds between status checks. Default: 10. Must be >= 1.
+    """
+
+    max_wait: int = 300
+    poll_interval: int = 10
+
+    def __post_init__(self):
+        if self.max_wait < 1:
+            raise ValueError("max_wait must be at least 1")
+        if self.poll_interval < 1:
+            raise ValueError("poll_interval must be at least 1")
--- a/src/bedrock_agentcore/_utils/polling.py
+++ b/src/bedrock_agentcore/_utils/polling.py
@@ -1,0 +1,92 @@
+"""Shared polling helpers for SDK clients."""
+
+import logging
+import time
+from typing import Any, Callable, Dict, Optional, Set
+
+from .config import WaitConfig
+
+logger = logging.getLogger(__name__)
+
+
+def wait_until(
+    poll_fn: Callable[[], Dict[str, Any]],
+    target: str,
+    failed: Set[str],
+    wait_config: Optional[WaitConfig] = None,
+    error_field: str = "statusReasons",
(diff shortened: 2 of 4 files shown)


src/bedrock_agentcore/_utils/polling.py
```diff
@@ -0,0 +1,92 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def wait_until(
+    poll_fn: Callable[[], Dict[str, Any]],
+    target: str,
+    failed: Set[str],
+    wait_config: Optional[WaitConfig] = None,
+    error_field: str = "statusReasons",
```
jariy17, line 17: why was this default chosen?

src/bedrock_agentcore/_utils/config.py
```diff
@@ -0,0 +1,22 @@
...
+@dataclass
+class WaitConfig:
+    """Configuration for *_and_wait polling methods.
+
+    Args:
+        max_wait: Maximum seconds to wait. Default: 300. Must be >= 1.
+        poll_interval: Seconds between status checks. Default: 10. Must be >= 1.
+    """
+
+    max_wait: int = 300
```
jariy17, line 15: What happens if one function requires different max_waits than another? For example, if  create_memories takes longer than 5 minutes usually, then create_memories_and_wait would usually fail with these defaults.

src/bedrock_agentcore/_utils/config.py
```diff
@@ -0,0 +1,22 @@
...
+    poll_interval: int = 10
```
jariy17, line 16: same here.

jariy17: should we do exponential backoffs? (https://docs.aws.amazon.com/prescriptive-guidance/latest/cloud-design-patterns/retry-backoff.html)

src/bedrock_agentcore/_utils/config.py
nborges-aws: If a `*_and_wait` method requires a larger window than the standard, the default max_wait can be set at the method level. Right now all the `_and_wait` methods pass `None` if the caller does not provide a WaitConfig. But if necessary for a specific case we could instead pass in a WaitConfig we define.

src/bedrock_agentcore/_utils/polling.py
nborges-aws: The majority of CP resources use `statusReasons` as the error detail field, which is why I set it as the default. The few primitives whose CP resources use `failureReason` are set as such at the method level

src/bedrock_agentcore/_utils/config.py
nborges-aws: Same answer as above. As for the exponential backoff, I think it would be good to have. Should we include that as an optional parameter or a default behavior?

jariy17: lets keep that as a p1

jariy17 (approved):

Issue #390, opened 2026-04-06
feat: Gateway Control Plane Client


Hweinstock:
## Problem

The SDK has no gateway management surface area. Customers who need to programmatically create MCP gateways, manage targets (Lambda, OpenAPI, Smithy), configure credentials (API key, OAuth2), attach policy engines, or toggle observability on gateways have no SDK path — they must use the CLI or console. The starter toolkit's `GatewayClient` demonstrates the full scope of what the SDK should support: CRUD, target management, name-based resolution, and status polling.

## Acceptance Criteria

- [ ] Customers can create a gateway and the SDK polls until it reaches READY status
- [ ] Customers can get, list, update, and delete gateways
- [ ] Customers can create, get, list, and delete gateway targets (Lambda, OpenAPI, Smithy)
- [ ] Customers can configure credential providers (API key, OAuth2) on gateway targets
- [ ] Customers can attach/update policy engine configuration on a gateway
- [ ] Name-based lookup is supported (resolve gateway/target by name, not just ID)
- [ ] All functionality is verified via integration tests running in CI

## Relevant Links

- [`GatewayClient`](https://github.com/aws/bedrock-agentcore-starter-toolkit/blob/4b9387f0d48cb6639633437b669fc6cc09ef07be/src/bedrock_agentcore_starter_toolkit/operations/gateway/client.py#L25)
- [`create_mcp_gateway()`](https://github.com/aws/bedrock-agentcore-starter-toolkit/blob/4b9387f0d48cb6639633437b669fc6cc09ef07be/src/bedrock_agentcore_starter_toolkit/operations/gateway/client.py#L57)
- [`create_mcp_gateway_target()`](https://github.com/aws/bedrock-agentcore-starter-toolkit/blob/4b9387f0d48cb6639633437b669fc6cc09ef07be/src/bedrock_agentcore_starter_toolkit/operations/gateway/client.py#L134)
- [`delete_gateway()`](https://github.com/aws/bedrock-agentcore-starter-toolkit/blob/4b9387f0d48cb6639633437b669fc6cc09ef07be/src/bedrock_agentcore_starter_toolkit/operations/gateway/client.py#L268)
- [`update_gateway()`](https://github.com/aws/bedrock-agentcore-starter-toolkit/blob/4b9387f0d48cb6639633437b669fc6cc09ef07be/src/bedrock_agentcore_starter_toolkit/operations/gateway/client.py#L926)
[...]

PR #428, opened 2026-04-23
feat: add gateway client and tests

nborges-aws:
*Issue #, if available:* [Issue #390](https://github.com/aws/bedrock-agentcore-sdk-python/issues/390)

*Description of changes:*
* Add GatewayClient with `__getattr__` passthrough
* Control plane allowlisted methods (16):
  * Gateway CRUD: create_gateway, get_gateway, list_gateways, update_gateway, delete_gateway
  * Gateway target CRUD: create_gateway_target, get_gateway_target, list_gateway_targets, update_gateway_target, delete_gateway_target
  * Gateway rules: create_gateway_rule, get_gateway_rule, list_gateway_rules, update_gateway_rule, delete_gateway_rule
  * Other: synchronize_gateway_targets
* Add `*_and_wait` polling methods using shared wait_until/wait_until_deleted utilities:
  * create_gateway_and_wait, update_gateway_and_wait, delete_gateway_and_wait
  * create_gateway_target_and_wait, update_gateway_target_and_wait, delete_gateway_target_and_wait
* Add name-based lookup methods with per-page pagination and early termination:
  * `get_gateway_by_name` — paginates through gateways, returns full details on first match
  * `get_gateway_target_by_name` — paginates through targets for a given gateway, returns full details on first match
* All `*_and_wait` methods accept WaitConfig for configurable polling behavior and apply convert_kwargs for snake_case support

*Test Plan:*
* Unit tests (11 total, all passing):
  * Init: region, region fallback
  * Passthrough: CP forwarding, snake_case conversion, AttributeError on unknown method
  * Wait methods: create gateway, create gateway failed, create target, delete gateway
  * Name lookup: found, not found
* Integration tests (15 total):
  * Gateway CRUD lifecycle: create+wait → get (camelCase + snake_case) → name lookup (found + not found) → list → update+wait → delete+wait
  * Target CRUD lifecycle: create+wait → get → name lookup (found + not found) → list → update+wait → delete+wait

### DIFF
--- a/src/bedrock_agentcore/gateway/client.py
+++ b/src/bedrock_agentcore/gateway/client.py
@@ -156,11 +156,11 @@ class GatewayClient:
             TimeoutError: If the target doesn't become READY within max_wait.
         """
         response = self.cp_client.create_gateway_target(**convert_kwargs(kwargs))
-        gw_arn = response["gatewayArn"]
+        gw_id = response["gatewayArn"].rsplit("/", 1)[-1]
         target_id = response["targetId"]
         return wait_until(
             lambda: self.cp_client.get_gateway_target(
-                gatewayIdentifier=gw_arn,
+                gatewayIdentifier=gw_id,
                 targetId=target_id,
             ),
             "READY",
@@ -184,11 +184,11 @@ class GatewayClient:
             TimeoutError: If the target doesn't become READY within max_wait.
         """
         response = self.cp_client.update_gateway_target(**convert_kwargs(kwargs))
-        gw_arn = response["gatewayArn"]
+        gw_id = response["gatewayArn"].rsplit("/", 1)[-1]
         target_id = response["targetId"]
         return wait_until(
             lambda: self.cp_client.get_gateway_target(
-                gatewayIdentifier=gw_arn,
+                gatewayIdentifier=gw_id,
                 targetId=target_id,
             ),
             "READY",
@@ -232,11 +232,11 @@ class GatewayClient:
             TimeoutError: If the target isn't deleted within max_wait.
         """
         response = self.cp_client.delete_gateway_target(**convert_kwargs(kwargs))
-        gw_arn = response["gatewayArn"]
+        gw_id = response["gatewayArn"].rsplit("/", 1)[-1]
         target_id = response["targetId"]
         wait_until_deleted(
             lambda: self.cp_client.get_gateway_target(
-                gatewayIdentifier=gw_arn,
+                gatewayIdentifier=gw_id,
                 targetId=target_id,
             ),
             wait_config=wait_config,
--- a/src/bedrock_agentcore/gateway/client.py
+++ b/src/bedrock_agentcore/gateway/client.py
@@ -1,0 +1,302 @@
+"""AgentCore Gateway SDK - Client for MCP gateway and target operations."""
+
+import logging
+from typing import Any, Dict, Optional
+
+import boto3
+from botocore.config import Config
+
+from .._utils.config import WaitConfig
+from .._utils.polling import wait_until, wait_until_deleted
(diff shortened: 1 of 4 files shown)


tests_integ/gateway/test_gateway_client.py
```diff
@@ -0,0 +1,269 @@
...
+@pytest.mark.integration
+class TestGatewayClient:
+    """Integration tests for GatewayClient CRUD and wait methods."""
+
+    @classmethod
+    def setup_class(cls):
+        cls.region = os.environ.get("BEDROCK_TEST_REGION", "us-west-2")
+        cls.role_arn = os.environ.get("GATEWAY_ROLE_ARN")
+        if not cls.role_arn:
+            pytest.skip("GATEWAY_ROLE_ARN not set")
```
Hweinstock: could this lead to tests being silently skipped or would the CI action still fail?

tests_integ/gateway/test_gateway_client.py
```diff
@@ -0,0 +1,269 @@
...
+        if not self.gateway_ids:
+            pytest.skip("prerequisite test did not create gateway")
+        updated = self.client.update_gateway_and_wait(
+            gatewayIdentifier=self.gateway_ids[0],
+            name=f"{self.test_prefix}-gw",
+            roleArn=self.role_arn,
+            authorizerType="NONE",
+            description="updated by integ test",
+        )
+        assert updated["status"] == "READY"
```
Hweinstock, line 102: should we assert that the gateway was actually updated?

src/bedrock_agentcore/gateway/client.py
```diff
@@ -0,0 +1,302 @@
...
+        while True:
+            response = self.cp_client.list_gateways(**kwargs)
+            for gw in response.get("items", []):
+                if gw.get("name") == name:
+                    return self.cp_client.get_gateway(
+                        gatewayIdentifier=gw["gatewayId"],
+                    )
+            if not response.get("nextToken"):
+                return None
+            kwargs["nextToken"] = response["nextToken"]
```
Hweinstock: Two questions:
- why would a user pass in nextToken?
- should we use a local variable instead to avoid mutating the parameters?

tests_integ/gateway/test_gateway_client.py
```diff
@@ -0,0 +1,269 @@
...
+import time
+
+import pytest
+from botocore.exceptions import ClientError
+
+from bedrock_agentcore.gateway.client import GatewayClient
+
+
+@pytest.mark.integration
+class TestGatewayClient:
```
Hweinstock, line 18: nit: could we add one more get_gateway_target_by_name

Hweinstock (commented):
Looks good! Small qs about integ tests.

tests_integ/gateway/test_gateway_client.py
nborges-aws: yep I'll add same assertion that i did for policy to make sure its actually updated

tests_integ/gateway/test_gateway_client.py
nborges-aws: There's tests for `get_gateway_target_by_name` in the TestGatewayTargetClient class down below. Is there a specific case that you were thinking of missing?

src/bedrock_agentcore/gateway/client.py
nborges-aws: * Popping the nextToken was just a defensive guard to make sure _if_ the user passed in a nextToken, it does not conflict with our internal pagination. Maybe not necessary but cheap and harmless imo
* Good catch. I've changed this to a local var

Hweinstock (dismissed):

Hweinstock (approved):

PR #429, opened 2026-04-23
feat: add identity client passthrough and tests

nborges-aws:
*Description of changes:*
* Add `__getattr__` passthrough with `accept_snake_case_kwargs` to IdentityClient
* Control plane allowlisted methods (11):
  * OAuth2 credential provider CRUD: create_oauth2_credential_provider, get_oauth2_credential_provider, list_oauth2_credential_providers, update_oauth2_credential_provider, delete_oauth2_credential_provider
  * API key credential provider CRUD: create_api_key_credential_provider, get_api_key_credential_provider, list_api_key_credential_providers, delete_api_key_credential_provider
  * Workload identity: get_workload_identity, update_workload_identity
* Data plane allowlisted methods (5):
  * get_resource_oauth2_token, get_resource_api_key, get_workload_access_token, get_workload_access_token_for_jwt, get_workload_access_token_for_user_id
* Remove 4 explicit methods with no additional logic, now covered by passthrough: `create_oauth2_credential_provider,` `create_api_key_credential_provider`, `update_workload_identity`, `get_workload_identity`
* Retain explicit methods that have additional logic: `get_workload_access_token`, `create_workload_identity`, `complete_resource_token_auth`, `get_token`, `get_api_key`

**Note**: `create_oauth2_credential_provider_and_wait` and `delete_oauth2_credential_provider_and_wait` were originally part of this change, but were removed. Though the service model defines a status field on the `GetOauth2CredentialProvider` response, integ tests revealed that the API was not returning any status field. The provider became available synchronously after creation. Polling methods were timing out waiting for a status that never appears, even when provider was created successfully. Thus the `*_and_wait` methods were dead code.

*Test plan:*
* Unit tests (24 total):
  * Existing tests updated to use __getattr__ passthrough calling convention (kwargs instead of positional dict)
  * OAuth2/API key creation, workload identity CRUD, token flows, auth completion, token polling
* Integration tests (9 total):
  * Passthrough: list OAuth2 providers, list API key providers, list with snake_case kwargs
  * Error handling: get nonexistent OAuth2 provider, get nonexistent API key provider
  * OAuth2 CRUD lifecycle: create → get → delete

### DIFF
--- a/src/bedrock_agentcore/services/identity.py
+++ b/src/bedrock_agentcore/services/identity.py
@@ -93,19 +93,14 @@ class IdentityClient:
     # -------------------------------------------------------------------------
     _ALLOWED_CP_METHODS = {
         # OAuth2 credential provider CRUD
-        "create_oauth2_credential_provider",
         "get_oauth2_credential_provider",
         "list_oauth2_credential_providers",
         "update_oauth2_credential_provider",
         "delete_oauth2_credential_provider",
         # API key credential provider CRUD
-        "create_api_key_credential_provider",
         "get_api_key_credential_provider",
         "list_api_key_credential_providers",
         "delete_api_key_credential_provider",
-        # Workload identity
-        "get_workload_identity",
-        "update_workload_identity",
     }
 
     _ALLOWED_DP_METHODS = {
@@ -131,6 +126,16 @@ class IdentityClient:
             f"Available methods can be found in the boto3 documentation for "
             f"'bedrock-agentcore' and 'bedrock-agentcore-control' services."
         )
+
+    def create_oauth2_credential_provider(self, req):
+        """Create an OAuth2 credential provider."""
+        self.logger.info("Creating OAuth2 credential provider...")
+        return self.cp_client.create_oauth2_credential_provider(**req)
+
+    def create_api_key_credential_provider(self, req):
+        """Create an API key credential provider."""
+        self.logger.info("Creating API key credential provider...")
+        return self.cp_client.create_api_key_credential_provider(**req)
 
     def get_workload_access_token(
         self, workload_name: str, user_token: Optional[str] = None, user_id: Optional[str] = None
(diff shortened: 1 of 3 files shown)


tests_integ/identity/test_identity_client.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @pytest.mark.order(12)
+    def test_delete_oauth2_credential_provider(self):
+        self.client.delete_oauth2_credential_provider(
+            name=self.provider_name,
+        )
+        # Provider may take a moment to delete
+        import time
+
+        time.sleep(5)
```
Hweinstock: should we poll here?

src/bedrock_agentcore/services/identity.py
```diff
@@ -125,20 +163,6 @@ def create_workload_identity(
             name=name, allowedResourceOauth2ReturnUrls=allowed_resource_oauth_2_return_urls or []
         )
 
-    def update_workload_identity(self, name: str, allowed_resource_oauth_2_return_urls: list[str]) -> Dict:
```
Hweinstock, line 128: now that we are proxying it through boto core, does the signature change at all? Would this be a breaking change for customers that already rely on this signature?

Same question for the other ones where we switched to passthrough.

src/bedrock_agentcore/services/identity.py
```diff
@@ -85,15 +89,49 @@ def __init__(self, region: str):
...
+            return accept_snake_case_kwargs(method)
+
+        raise AttributeError(
+            f"'{self.__class__.__name__}' object has no attribute '{name}'. "
+            f"Method not found on data plane or control plane client. "
+            f"Available methods can be found in the boto3 documentation for "
+            f"'bedrock-agentcore' and 'bedrock-agentcore-control' services."
+        )
 
     def get_workload_access_token(
```
Hweinstock, line 140: this is defined as an explicit method and a passthrough. I think the passthrough takes precedence? If so, should we delete the old one?

nborges-aws: I think it's the other way around. Explicit methods are resolved first, and then it goes to passthrough.

But regardless, if I'm right the passthrough method is unreachable given we have an explicit method. Keeping it in the allowlist is pointless, and could confuse users. Will remove.

src/bedrock_agentcore/services/identity.py
nborges-aws: The signature and args do not change at all, but the methods go from accepting positional or keyword arguments to only keyword arguments. This is not a breaking change for any customer calling  `method(keyword="value")`, but will break for those using `method("value")`.  Boto3 does throw a clear error message when this happens:

`TypeError: get_workload_identity() only accepts keyword arguments.`

If we want to preserve full backwards compatibility, we would have to remove these methods from allowlist and add back explicit methods. 

@jariy17 thoughts?

tests_integ/identity/test_identity_client.py
nborges-aws: I'll change this to use `wait_until_deleted` utility

Hweinstock (approved):

src/bedrock_agentcore/services/identity.py
jariy17: we discussed and we are bring back these function with a note saying we should deprecate this functions.

jariy17 (approved):

PR #433, opened 2026-04-23
fix(gateway): use gateway ID instead of ARN for target polling

jesseturner21:
## Summary
- `get_gateway_target` returns `AccessDeniedException` when called with a gateway ARN as `gatewayIdentifier` — it only accepts gateway IDs
- `create_gateway_target_and_wait`, `update_gateway_target_and_wait`, and `delete_gateway_target_and_wait` were passing `gatewayArn` from the response directly to the polling call
- Fixed by extracting the gateway ID from the ARN: `response["gatewayArn"].rsplit("/", 1)[-1]`
- Added missing `credentialProviderConfigurations` to `test_update_gateway_target_and_wait` integ test (required by the API)

## Test plan
- [x] All 15 integ tests pass with `--order-scope=class`
- [x] Verified `get_gateway_target` works with gateway ID but fails with ARN

### DIFF
--- a/src/bedrock_agentcore/gateway/client.py
+++ b/src/bedrock_agentcore/gateway/client.py
@@ -156,11 +156,11 @@ class GatewayClient:
             TimeoutError: If the target doesn't become READY within max_wait.
         """
         response = self.cp_client.create_gateway_target(**convert_kwargs(kwargs))
-        gw_arn = response["gatewayArn"]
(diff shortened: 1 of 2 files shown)


nborges-aws:
niceeeeeeeee!

nborges-aws (approved):

Issue #391, opened 2026-04-06
feat: Runtime Resource Control Plane Client


Hweinstock:
## Problem

The SDK's runtime module (`BedrockAgentCoreApp`) only handles running the agent server process. There is no client for the runtime control plane — customers cannot programmatically create, update, get, or delete agent runtimes or endpoints. The starter toolkit wraps `bedrock-agentcore-control` APIs (`create_agent_runtime`, `get_agent_runtime_endpoint`, etc.) with status polling, endpoint readiness checks, and aggregated status reporting. Without this, customers have no programmatic path to manage agent runtime resources outside the CLI.

## Acceptance Criteria

- [ ] Customers can create, get, update, and delete agent runtimes
- [ ] Customers can get and delete agent runtime endpoints
- [ ] Create/update operations poll until the endpoint reaches READY status
- [ ] Customers can query aggregated status (runtime + endpoint + memory)
- [ ] Teardown deletes resources in correct order (endpoint → runtime)
- [ ] All functionality is verified via integration tests running in CI

## Relevant Links

- [`BedrockAgentCoreClient`](https://github.com/aws/bedrock-agentcore-starter-toolkit/blob/4b9387f0d48cb6639633437b669fc6cc09ef07be/src/bedrock_agentcore_starter_toolkit/services/runtime.py#L133)
- [`launch_bedrock_agentcore()`](https://github.com/aws/bedrock-agentcore-starter-toolkit/blob/4b9387f0d48cb6639633437b669fc6cc09ef07be/src/bedrock_agentcore_starter_toolkit/operations/runtime/launch.py#L784)
- [`destroy_bedrock_agentcore()`](https://github.com/aws/bedrock-agentcore-starter-toolkit/blob/4b9387f0d48cb6639633437b669fc6cc09ef07be/src/bedrock_agentcore_starter_toolkit/operations/runtime/destroy.py#L20)
- [`get_status()`](https://github.com/aws/bedrock-agentcore-starter-toolkit/blob/4b9387f0d48cb6639633437b669fc6cc09ef07be/src/bedrock_agentcore_starter_toolkit/operations/runtime/status.py#L12)

nborges-aws:
Updates to runtime client completed in #434. Changes merged to main

PR #417, opened 2026-04-15
fix: validate region parameter to prevent SSRF request redirection

jariy17:
## Summary

Fixes SSRF vulnerability where crafted `region` values (e.g., `x@attacker.com:443/#`) redirect SDK API calls — including SigV4-signed requests with credentials — to non-AWS hosts. Same vulnerability class as CVE-2026-22611.

### Two-part fix:

**1. Region validation for manual URL construction paths** (endpoints.py, a2a.py, AgentCoreRuntimeClient, BrowserClient)
- Layer 1: `validate_region()` with regex `\A[a-z]{2}(-[a-z]+)+-\d+\Z` using `\Z` anchor to prevent newline bypass
- Layer 2: `_validate_endpoint_url()` verifies constructed URL hostname ends with `.amazonaws.com`
- Custom `InvalidRegionError(ValueError)` for backward compatibility

**2. Remove redundant `endpoint_url` from boto3 clients** (CodeInterpreter, IdentityClient, ResourcePolicyClient, MemoryControlPlaneClient)
- Boto3's own endpoint resolution produces identical URLs and includes built-in region validation
- `endpoint_url` now only passed when env var override is explicitly set (`BEDROCK_AGENTCORE_DP_ENDPOINT`, `BEDROCK_AGENTCORE_CP_ENDPOINT`)

### Files changed

| File | Change |
|------|--------|
| `_utils/endpoints.py` | `validate_region()`, `_validate_endpoint_url()`, `InvalidRegionError` |
| `runtime/a2a.py` | Validate in `build_runtime_url()` (explicit + ARN-extracted region) |
| `runtime/agent_core_runtime_client.py` | Validate in `__init__()` |
| `tools/browser_client.py` | Validate in `__init__()`, remove `endpoint_url` from boto3 calls |
| `tools/code_interpreter_client.py` | Remove `endpoint_url` from boto3 calls |
| `services/identity.py` | Remove `endpoint_url` from boto3 calls |
| `services/resource_policy.py` | Remove `endpoint_url` from boto3 calls |
| `memory/controlplane.py` | Remove `endpoint_url` from boto3 calls |
| `test_region_validation.py` | 54 new tests |
| `test_code_interpreter_client.py` | Updated to remove endpoint mocks |
| `test_browser_client.py` | Updated to remove endpoint mocks |

## Test plan

- [x] 54 new SSRF validation tests (valid regions, 20 malicious payloads, ARN extraction, all client constructors)
- [x] 296 total tests passing, 0 regressions
[...]

PR #434, opened 2026-04-24
feat: update runtime client with passthrough

nborges-aws:
*Issue #, if available:* [Issue #391](https://github.com/aws/bedrock-agentcore-sdk-python/issues/391)

*Description of changes:*
* Add `__getattr__` passthrough with accept_snake_case_kwargs to AgentCoreRuntimeClient
* Update `__init__` to create cp_client and dp_client boto3 clients
* Control plane allowlisted methods (12):
  * Runtime CRUD: create_agent_runtime, get_agent_runtime, list_agent_runtimes, update_agent_runtime, delete_agent_runtime
  * Endpoint CRUD: create_agent_runtime_endpoint, get_agent_runtime_endpoint, update_agent_runtime_endpoint, delete_agent_runtime_endpoint, list_agent_runtime_endpoints
  * Other: list_agent_runtime_versions, delete_agent_runtime_version
* Data plane allowlisted methods (2):
  * invoke_agent_runtime, stop_runtime_session
* Add `*_and_wait` polling methods using shared wait_until/wait_until_deleted utilities:
  * create_agent_runtime_and_wait, update_agent_runtime_and_wait, delete_agent_runtime_and_wait
  * create_agent_runtime_endpoint_and_wait, update_agent_runtime_endpoint_and_wait
* Add higher-level orchestration methods:
  * get_aggregated_status — returns runtime + endpoint status in a single call
  * teardown_endpoint_and_runtime — ordered deletion: waits for endpoint deletion then deletes runtime
* All `*_and_wait` methods accept WaitConfig for configurable polling behavior and apply convert_kwargs for snake_case support
* Existing methods (generate_ws_connection, generate_presigned_url, generate_ws_connection_oauth) are untouched

## Test plan
* Unit tests (16 new):
  * Passthrough: CP forwarding, DP forwarding, snake_case conversion, AttributeError on unknown method, CP allowlist contents, DP allowlist contents
  * *_and_wait: create_agent_runtime_and_wait (success + failure), update_agent_runtime_and_wait, delete_agent_runtime_and_wait, update_agent_runtime_endpoint_and_wait
  * Orchestration: get_aggregated_status (success, not found, reraises non-not-found), teardown_endpoint_and_runtime (success, skips missing endpoint)
* Integration tests (10 total):
  * Passthrough (3): list_agent_runtimes, list_agent_runtimes with snake_case, non-allowlisted method raises
  * Runtime + endpoint CRUD (7): create_agent_runtime_and_wait → get_agent_runtime (camelCase + snake_case) → update_agent_runtime_and_wait → create_agent_runtime_endpoint_and_wait → get_aggregated_status (verifies both runtime + endpoint READY) → teardown_endpoint_and_runtime
  * Requires RUNTIME_ROLE_ARN and RUNTIME_S3_CODE_URI environment variables

### DIFF
--- a/src/bedrock_agentcore/runtime/agent_core_runtime_client.py
+++ b/src/bedrock_agentcore/runtime/agent_core_runtime_client.py
@@ -19,7 +19,7 @@ from botocore.config import Config
 from botocore.exceptions import ClientError
 
 from .._utils.config import WaitConfig
-from .._utils.endpoints import get_data_plane_endpoint
+from .._utils.endpoints import get_data_plane_endpoint, validate_region
 from .._utils.polling import wait_until, wait_until_deleted
 from .._utils.snake_case import accept_snake_case_kwargs, convert_kwargs
 from .._utils.user_agent import build_user_agent_suffix
@@ -81,7 +81,7 @@ class AgentCoreRuntimeClient:
                 telemetry.
         """
         session = session if session else boto3.Session()
-        self.region = region or session.region_name or "us-west-2"
+        self.region = validate_region(region or session.region_name or "us-west-2")
         self.session = session
         self.integration_source = integration_source
         self.logger = logging.getLogger(__name__)
@@ -572,13 +572,11 @@ class AgentCoreRuntimeClient:
             TimeoutError: If the endpoint doesn't become READY within
                 max_wait.
         """
+        converted = convert_kwargs(kwargs)
         response = self.cp_client.create_agent_runtime_endpoint(
-            **convert_kwargs(kwargs),
+            **converted,
         )
-        rid = kwargs.get(
-            "agentRuntimeId",
-            convert_kwargs(kwargs).get("agentRuntimeId"),
-        )
+        rid = converted.get("agentRuntimeId")
         ename = response.get("name", kwargs.get("name", "DEFAULT"))
         return wait_until(
             lambda: self.cp_client.get_agent_runtime_endpoint(
@@ -611,13 +609,11 @@ class AgentCoreRuntimeClient:
             TimeoutError: If the endpoint doesn't become READY within
                 max_wait.
         """
+        converted = convert_kwargs(kwargs)
         response = self.cp_client.update_agent_runtime_endpoint(
-            **convert_kwargs(kwargs),
+            **converted,
         )
-        rid = kwargs.get(
-            "agentRuntimeId",
-            convert_kwargs(kwargs).get("agentRuntimeId"),
-        )
+        rid = converted.get("agentRuntimeId")
         ename = response.get("name", kwargs.get("endpointName", "DEFAULT"))
         return wait_until(
             lambda: self.cp_client.get_agent_runtime_endpoint(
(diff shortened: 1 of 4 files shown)


src/bedrock_agentcore/runtime/agent_core_runtime_client.py
```diff
@@ -401,3 +472,243 @@ def generate_ws_connection_oauth(
...
+        Returns:
+            Endpoint details when READY.
+
+        Raises:
+            RuntimeError: If the endpoint reaches a failed state.
+            TimeoutError: If the endpoint doesn't become READY within
+                max_wait.
+        """
+        response = self.cp_client.create_agent_runtime_endpoint(
+            **convert_kwargs(kwargs),
```
Hweinstock: nit: should we call convert_kwargs once and re-use?

src/bedrock_agentcore/runtime/agent_core_runtime_client.py
```diff
@@ -401,3 +472,243 @@ def generate_ws_connection_oauth(
...
+        )
+        rid = kwargs.get(
```
Hweinstock: would the behavior here be the same if we checked the field on the converted kwargs once?

My current understanding of the flow is we check for agentRuntimeId, then fallback to the converted ones. However, I thought converting it handled it for us here?

src/bedrock_agentcore/runtime/agent_core_runtime_client.py
```diff
@@ -35,24 +44,86 @@ class AgentCoreRuntimeClient:
...
+                default session is created.
+            integration_source: Optional integration source for user-agent
+                telemetry.
         """
-        from .._utils.endpoints import validate_region
-
-        validate_region(region)
-        self.region = region
+        session = session if session else boto3.Session()
+        self.region = region or session.region_name or "us-west-2"
```
Hweinstock: do we still need to validate the region or does boto3 handle this for us? Wondering if this opens us up to poisoning the endpoint? Context: https://github.com/aws/bedrock-agentcore-sdk-python/pull/417

nborges-aws: boto3 validates the region for its client creation (raises `botocore.exceptions.InvalidRegionError` for malicious inputs). Self.region is used to construct the websocket URLs, but that method calls `validate_region()` internally, so technically we are still fine with this region construction

However, we can validate_region on the `__init__` construction for safety. I'll make this change

src/bedrock_agentcore/runtime/agent_core_runtime_client.py
nborges-aws: yep I'll make this change

src/bedrock_agentcore/runtime/agent_core_runtime_client.py
nborges-aws: You're correct, we can check the converted kwargs once with the same end behavior. I will update

Hweinstock (approved):

Issue #432, opened 2026-04-23
unreliable test: memory integration test is flaky


Hweinstock:
ex. https://github.com/aws/bedrock-agentcore-sdk-python/actions/runs/24842115061/job/72718841096
```
FAILED tests_integ/memory/integrations/test_session_manager.py::TestAgentCoreMemorySessionManager::test_session_manager_with_retrieval_config_adds_context - assert '<user_context>' in '[{\'role\': \'user\', \'content\': [{\'text\': \'I like sushi with tuna\'}]}, {\'role\': \'assistant\', \'content\': [{\'text\': "That\'s great! Tuna is one of the most popular sushi ingredients. There are several delicious varieties you might enjoy:\\n\\n- **Akami** - The deep red, lean cut from the sides and back\\n- **Chutoro** - Medium fatty tuna from the belly, with a buttery texture\\n- **Otoro** - The fattiest, most prized cut that melts in your mouth\\n- **Spicy tuna** - Minced tuna mixed with spicy mayo, often in rolls\\n\\nDo you have a favorite style of tuna sushi? Are you into traditional nigiri, or do you prefer it in rolls like spicy tuna rolls or tuna avocado rolls?"}], \'metadata\': {\'usage\': {\'inputTokens\': 20, \'outputTokens\': 159, \'totalTokens\': 179}, \'metrics\': {\'latencyMs\': 2876, \'timeToFirstByteMs\': 902}}}, {\'role\': \'user\', \'content\': [{\'text\': \'What do I like to eat?\'}]}, {\'role\': \'assistant\', \'content\': [{\'text\': "Based on what you\'ve told me, I know you like sushi with tuna! But that\'s the only food preference you\'ve shared so far. \\n\\nI don\'t have information about your other food likes or dislikes. Would you like to tell me more about what you enjoy eating? I\'d be happy to discuss other foods, suggest recipes, or talk about cuisines you\'re interested in!"}], \'metadata\': {\'usage\': {\'inputTokens\': 189, \'outputTokens\': 86, \'totalTokens\': 275}, \'metrics\': {\'latencyMs\': 2135, \'timeToFirstByteMs\': 973}}}]'
[...]

PR #437, opened 2026-04-24
fix: reduce flakiness in retrieval config integration test

Hweinstock:
## Problem

The integration test `test_session_manager_with_retrieval_config_adds_context` is flaky (#432). It fails because the relevance score filter (`relevance_score=0.7`) drops retrieved memories when the semantic similarity between the query `"What do I like to eat?"` and the extracted preference about sushi scores below 0.7. The test failed 3/3 attempts (original + 2 reruns) in [this CI run](https://github.com/aws/bedrock-agentcore-sdk-python/actions/runs/24842115061/job/72718841096).

CI logs confirm: `Retrieved 1 memories` appears but `Retrieved X customer context items` never does — the relevance filter drops the memory before context injection.

## Solution

Two changes to make the test more reliable:

1. **Lower `relevance_score` from `0.7` to `0.3`** 

2. **Change query from `"What do I like to eat?"` to `"What sushi do I like?"`** — semantically closer to the stored preference, producing higher and more consistent relevance scores.

## Testing

The test validates the same invariant: retrieval config with LTM causes `<user_context>` to appear in agent messages. The changes only reduce non-determinism in the relevance score path without weakening the assertion.

### DIFF
--- a/tests_integ/memory/integrations/test_session_manager.py
+++ b/tests_integ/memory/integrations/test_session_manager.py
@@ -144,7 +144,7 @@ class TestAgentCoreMemorySessionManager:
             memory_id=test_memory_ltm["id"],
             session_id=f"test-session-{int(time.time())}",
             actor_id=actor_id,
-            retrieval_config={"/preferences/{actorId}/": RetrievalConfig(top_k=5, relevance_score=0.7)},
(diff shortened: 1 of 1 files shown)


notgitika (approved):
LGTM thanks for looking into this!

PR #443, opened 2026-04-28
feat: improve observability clients

nborges-aws:
### Summary
Migrates observability functionality from the starter toolkit into the SDK across two areas:

1. New `ObservabilityClient` — manages CloudWatch delivery configuration and X-Ray Transaction Search for AgentCore resources.
2. Expanded `CloudWatchSpanHelper` — adds trace-based querying, runtime log querying, and session discovery.

#### ObservabilityClient

Ported from the starter toolkit's `ObservabilityDeliveryManager` and `xray.py`.

Methods added:
- enable_observability_for_resource
- disable_observability_for_resource
- enable_transaction_search
- get_observability_status

#### CloudWatchSpanHelper

Methods added:
- query_spans_by_trace
- query_runtime_logs_by_traces
- get_latest_session_id

Methods expanded (both non-breaking changes):
- query_log_group — added optional agent_id parameter to allow filtering by agent_id
- fetch_spans — added optional agent_id parameter to allow filtering by agent_id

### Testing

- **ObservabilityClient:** 22 unit tests + 9 integration tests
- **CloudWatchSpanHelper:** 18 new unit tests + 5 integration tests

### DIFF
--- a/src/bedrock_agentcore/evaluation/utils/cloudwatch_span_helper.py
+++ b/src/bedrock_agentcore/evaluation/utils/cloudwatch_span_helper.py
@@ -44,6 +44,7 @@ class CloudWatchSpanHelper:
         start_time: datetime,
         end_time: datetime,
         query_string: Optional[str] = None,
+        agent_id: Optional[str] = None,
     ) -> List[dict]:
         """Query a single CloudWatch log group for session data.
 
@@ -54,16 +55,23 @@ class CloudWatchSpanHelper:
             end_time: Query end time
             query_string: Optional custom query string. When provided, used instead
                 of the default substring match query.
+            agent_id: Optional agent ID to filter by (prevents cross-agent session collisions)
 
         Returns:
             List of parsed JSON log messages
         """
         if query_string is None:
+            agent_filter = ""
+            if agent_id is not None:
+                agent_filter = (
+                    f'\n        | parse resource.attributes.cloud.resource_id "runtime/*/" as parsedAgentId'
+                    f"\n        | filter parsedAgentId = '{agent_id}'"
+                )
             query_string = f"""fields @timestamp, @message
         | filter @message like "{session_id}"
         | filter ispresent(scope.name)
         | filter ispresent(traceId)
-        | filter ispresent(spanId)
+        | filter ispresent(spanId){agent_filter}
         | sort @timestamp asc"""
 
         max_attempts = 30
@@ -149,6 +157,7 @@ class CloudWatchSpanHelper:
         event_log_group: str,
         start_time: datetime,
         end_time: Optional[datetime] = None,
+        agent_id: Optional[str] = None,
     ) -> List[dict]:
         """Fetch ADOT spans from CloudWatch with configurable event log group.
 
@@ -162,6 +171,7 @@ class CloudWatchSpanHelper:
                 - For custom agents: Any log group you configured (e.g., "/my-app/agent-events")
             start_time: Start time for log query
             end_time: End time for log query
+            agent_id: Optional agent ID to filter by (prevents cross-agent session collisions)
 
         Returns:
             List of ADOT span and log record dictionaries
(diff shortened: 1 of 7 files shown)


src/bedrock_agentcore/observability/client.py
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        for suffix in ["logs", "traces"]:
+            source_name = f"{resource_id}-{suffix}-source"
+            try:
+                self._logs_client.get_delivery_source(name=source_name)
+                status[suffix]["configured"] = True
+                status[suffix]["source_name"] = source_name
+            except ClientError:
+                pass
```
jariy17: **[HIGH] `get_observability_status` only checks delivery-source existence — reports `configured: True` when logs aren't flowing**

A delivery source can exist without a corresponding destination or active delivery (e.g. if `enable_observability_for_resource` crashed between steps). This method will return `configured: True` for both `logs` and `traces` even when the full delivery chain is broken.

**Fix:** also verify `get_delivery_destination(name=dest_name)` and that at least one delivery links source to destination. Consider returning a richer status: `{source_exists, destination_exists, delivery_exists, log_group_exists}`.

src/bedrock_agentcore/observability/client.py
```diff
@@ -0,0 +1,345 @@
...
+                }
+            ],
+        }
+        try:
+            self._logs_client.put_resource_policy(
+                policyName="TransactionSearchXRayAccess", policyDocument=json.dumps(policy_document)
+            )
+        except ClientError as e:
+            if e.response["Error"]["Code"] != "InvalidParameterException":
+                raise
```
jariy17: **[MEDIUM] `InvalidParameterException` is the wrong idempotence signal for `put_resource_policy`**

`put_resource_policy` is already an upsert — it doesn't raise on overwrite. `InvalidParameterException` signals a **malformed policy document** (e.g., bad principal, invalid condition). Silently swallowing it means a bug in the policy JSON (e.g., wrong account ID interpolation) goes undetected and `enable_transaction_search()` returns `True` with a broken policy in place.

**Fix:** remove this `except` block entirely — `put_resource_policy` doesn't need idempotence handling.

src/bedrock_agentcore/observability/client.py
```diff
@@ -0,0 +1,345 @@
...
+                try:
+                    self._xray_client.update_trace_segment_destination(Destination="CloudWatchLogs")
+                except ClientError as e:
+                    if e.response["Error"]["Code"] != "InvalidRequestException":
+                        raise
+
+            if self._need_indexing_rule():
+                try:
+                    self._xray_client.update_indexing_rule(
+                        Name="Default", Rule={"Probabilistic": {"DesiredSamplingPercentage": 1}}
```
jariy17: **[MEDIUM] X-Ray indexing rule sampling hardcoded to 1% with no parameter to override**

```python
Rule={"Probabilistic": {"DesiredSamplingPercentage": 1}}
```

Note: `_need_indexing_rule` only triggers this if no `Default` rule exists yet — it will not overwrite an existing rule with a different percentage. However, on a fresh account `1%` is set silently with no documentation of the value or way to change it. A new user enabling Transaction Search would expect full (or at least high) coverage, not 1%.

**Fix:** expose as a parameter (e.g. `sampling_percentage: int = 100`) and document the cost/coverage tradeoff in the docstring.

src/bedrock_agentcore/observability/client.py
jariy17: **[HIGH] `get_observability_status` only checks delivery-source existence — reports `configured: True` when logs aren't flowing**

Per the AWS docs, a delivery source, destination, and delivery are three independent objects created in sequence. A source can exist without a delivery connecting it to a destination — the docs state: *"Use `CreateDelivery` to create a delivery by pairing exactly one delivery source and one delivery destination."* Without an active delivery, logs do not flow.

This method only calls `get_delivery_source`. It will return `configured: True` even when the delivery or destination is missing (e.g., if `enable_observability_for_resource` crashed midway).

**Fix:** also check that at least one active delivery exists linking source to destination via `describe_deliveries`.

src/bedrock_agentcore/evaluation/utils/cloudwatch_span_helper.py
```diff
@@ -54,16 +55,23 @@ def query_log_group(
...
         Returns:
             List of parsed JSON log messages
         """
         if query_string is None:
+            agent_filter = ""
+            if agent_id is not None:
+                agent_filter = (
+                    f'\n        | parse resource.attributes.cloud.resource_id "runtime/*/" as parsedAgentId'
+                    f"\n        | filter parsedAgentId = '{agent_id}'"
+                )
```
jariy17, line 69: **[CRITICAL] CWL Insights injection — `agent_id`, `session_id`, `trace_id`, `endpoint_name`**

Every new method interpolates caller-controlled values directly into CWL Insights query strings with only quote-wrapping and no validation:

```python
f"| filter parsedAgentId = '{agent_id}'"          # line 69
f'| filter @message like "{session_id}"'           # line 71
f"| filter traceId = '{trace_id}'"                 # lines 234, 356
f"/aws/bedrock-agentcore/runtimes/{agent_id}-{endpoint_name}"  # line 261 (log group name)
f"| filter parsedAgentId = '{agent_id}'"           # line 291
```

A value like `x' | stats count(*) as n #` rewrites the query semantics. A malicious `endpoint_name` can pivot the log group path to any group the caller's IAM role can read. In any multi-tenant SDK wrapper this is a cross-tenant log disclosure vector. No test exercises a hostile input.

**Fix:** validate all interpolated values against a strict pattern (e.g. `^[A-Za-z0-9_\-:.]{1,64}$`) and reject anything else before interpolation.

src/bedrock_agentcore/evaluation/utils/cloudwatch_span_helper.py
```diff
@@ -190,14 +200,167 @@ def fetch_spans(
...
+            queryString=query_string,
+        )
+        query_id = response["queryId"]
+
+        timeout = 60
+        poll_interval = 2
+        start = time.time()
+        while True:
+            if time.time() - start > timeout:
+                raise TimeoutError(f"Query {query_id} timed out after {timeout} seconds")
```
jariy17, line 336: **[CRITICAL] Running query is never cancelled on timeout, error, or exception**

When `TimeoutError` is raised (or any exception propagates out of the polling loop), the query continues running server-side. There is no `try/finally` and no call to `stop_query(queryId=query_id)`.

Per AWS docs: *"queries continue to run until completion"* unless `StopQuery` is explicitly called. Abandoned queries consume one of your **100 regional concurrent query slots** (shared with scheduled queries) for up to the **60-minute server-side timeout**. The `_query_runtime_logs_individually` fallback multiplies this: a single failed batch query spawns N individual queries that can each also leak a slot.

**Fix:**
```python
status = None
try:
    while True:
        ...
        if status == "Complete":
            return result.get("results", [])
finally:
    if status != "Complete":
        self.logs_client.stop_query(queryId=query_id)
```

src/bedrock_agentcore/evaluation/utils/cloudwatch_span_helper.py
```diff
@@ -190,14 +200,167 @@ def fetch_spans(
...
+            result = self.logs_client.get_query_results(queryId=query_id)
+            status = result["status"]
+            if status == "Complete":
+                return result.get("results", [])
```
jariy17, line 340: **[HIGH] No 10,000-row truncation warning — silent data loss**

`_execute_query` returns `result.get("results", [])` with no check on `statistics.recordsMatched`. Per AWS docs, `GetQueryResults` has a hard ceiling of 10,000 rows; `statistics.recordsMatched` reflects the total matching events regardless of this cap, so truncation is detectable.

Contrast with `query_log_group` (lines 106–117) which explicitly checks `if records_matched > 10000` and warns. All three new methods (`query_spans_by_trace`, `query_runtime_logs_by_traces`, `get_latest_session_id`) go through `_execute_query` and silently truncate. Evaluation pipelines computing metrics over truncated span data will produce incorrect results with no signal.

**Fix:** port the `recordsMatched` check from `query_log_group` into this method.

src/bedrock_agentcore/evaluation/utils/cloudwatch_span_helper.py
```diff
@@ -190,14 +200,167 @@ def fetch_spans(
...
+        results = []
+        for trace_id in trace_ids:
+            query = f"""fields @timestamp, @message, spanId, traceId, @logStream
+            | filter traceId = '{trace_id}'
+            | sort @timestamp asc"""
+            try:
+                results.extend(self._execute_query(query, log_group, start_time_ms, end_time_ms))
+            except Exception as e:
+                logger.warning("Failed to query runtime logs for trace %s: %s", trace_id, e)
+        return results
```
jariy17, line 362: **[HIGH] Fallback silently returns partial results — no indication of which traces failed**

Each per-trace `except Exception` is swallowed and the loop continues. The caller receives a plain `List[dict]` with no metadata about completeness — indistinguishable from a full successful result. Evaluation pipelines will compute error rates, latency, and span counts over an unknown partial dataset and report them as authoritative.

**Fix:** return a structured result such as `{"results": [...], "failed_trace_ids": [...], "complete": bool}`, or raise after the batch fallback already fired to ensure the caller is aware.

src/bedrock_agentcore/observability/client.py
```diff
@@ -0,0 +1,345 @@
...
+        try:
+            if resource_type not in AUTO_LOG_RESOURCE_TYPES:
+                self._create_log_group_if_not_exists(log_group_name)
+
+            if enable_logs and resource_type not in AUTO_LOG_RESOURCE_TYPES:
+                results["deliveries"]["logs"] = self._setup_logs_delivery(resource_arn, resource_id, log_group_arn)
+                results["logs_enabled"] = True
+            elif resource_type in AUTO_LOG_RESOURCE_TYPES:
+                results["logs_enabled"] = True
+                results["deliveries"]["logs"] = {"status": "auto-created by AWS"}
```
jariy17, line 93: **[MEDIUM] `logs_enabled: True` returned for runtimes with no delivery verification**

```python
elif resource_type in AUTO_LOG_RESOURCE_TYPES:
    results["logs_enabled"] = True
    results["deliveries"]["logs"] = {"status": "auto-created by AWS"}
```

AWS docs confirm that AgentCore does auto-create a CloudWatch log group when a runtime resource is created, so the assumption is generally correct. However, the code makes no AWS call to verify the log group actually exists before returning `logs_enabled: True`. If the runtime was never fully provisioned, or if auto-creation is delayed, the caller has no way to distinguish "confirmed" from "assumed."

**Fix:** document the assumption explicitly in the docstring, and consider a lightweight `describe_log_groups` check (with a short timeout) for callers who need confirmation rather than assumption.

src/bedrock_agentcore/observability/client.py
```diff
@@ -0,0 +1,345 @@
...
+        """Disable CloudWatch observability delivery for a resource."""
+        results: Dict[str, Any] = {"resource_id": resource_id, "deleted": [], "errors": []}
+
+        for suffix in ["logs", "traces"]:
+            source_name = f"{resource_id}-{suffix}-source"
+            dest_name = f"{resource_id}-{suffix}-destination"
+
+            # Delete deliveries referencing this source first
+            try:
+                deliveries = self._logs_client.describe_deliveries()
```
jariy17, line 121: **[HIGH] `describe_deliveries` is not paginated**

```python
deliveries = self._logs_client.describe_deliveries()
```

Per the AWS API docs, `DescribeDeliveries` returns at most 50 results per page and includes a `nextToken` for subsequent pages. Accounts with >50 deliveries will have the target delivery on a later page. The subsequent `delete_delivery_source` then fails with `ConflictException` (not handled here — only `ResourceNotFoundException` is caught), leaving orphaned delivery sources and returning `status: partial` with no clear explanation.

**Fix:**
```python
paginator = self._logs_client.get_paginator("describe_deliveries")
for page in paginator.paginate():
    for delivery in page.get("deliveries", []):
        ...
```

src/bedrock_agentcore/observability/client.py
```diff
@@ -0,0 +1,345 @@
...
+                if resource_type == "runtime":
+                    lg = f"/aws/bedrock-agentcore/runtimes/{resource_id}"
+                else:
+                    lg = f"/aws/vendedlogs/bedrock-agentcore/{resource_type}/APPLICATION_LOGS/{resource_id}"
+                try:
+                    self._logs_client.delete_log_group(logGroupName=lg)
+                    results["deleted"].append(f"log_group:{lg}")
+                except ClientError as e:
+                    if e.response["Error"]["Code"] != "ResourceNotFoundException":
+                        results["errors"].append(f"Failed to delete log group {lg}: {e}")
```
jariy17, line 158: **[CRITICAL] `disable_observability_for_resource(delete_log_group=True)` deletes log groups for ALL resource types, not just the one being disabled**

`disable_observability_for_resource` takes no `resource_type` parameter. When `delete_log_group=True`, it iterates all of `SUPPORTED_RESOURCE_TYPES` and attempts to delete:
- `/aws/bedrock-agentcore/runtimes/{resource_id}`
- `/aws/vendedlogs/bedrock-agentcore/gateway/APPLICATION_LOGS/{resource_id}`
- `/aws/vendedlogs/bedrock-agentcore/memory/APPLICATION_LOGS/{resource_id}`

Disabling a `memory` resource silently attempts to delete the auto-managed `runtime` log group and any `gateway` log group sharing the same ID. `ResourceNotFoundException` is silently swallowed so collateral deletion goes unnoticed. Note the asymmetry: `enable_observability_for_resource` requires `resource_type` and constructs exactly one path; `disable` does not.

**Fix:** add `resource_type` as a required parameter to `disable_observability_for_resource` and construct exactly one log group path, matching the enable-side logic.

src/bedrock_agentcore/observability/client.py
jariy17, line 179: **[MEDIUM] X-Ray indexing rule sampling hardcoded to 1% with no parameter to override**

```python
Rule={"Probabilistic": {"DesiredSamplingPercentage": 1}}
```

Note: `_need_indexing_rule` only triggers this if no `Default` rule exists yet — it will not overwrite an existing rule. However, on a fresh account `1%` is set silently with no documentation or way to change it. A new user enabling Transaction Search would expect full (or at least high) coverage, not 1%.

**Fix:** expose as a parameter (e.g. `sampling_percentage: int = 100`) and document the cost/coverage tradeoff in the docstring.

src/bedrock_agentcore/observability/client.py
jariy17, line 208: **[HIGH] `get_observability_status` only checks delivery-source existence — reports `configured: True` when logs aren't flowing**

Per the AWS docs, a delivery source, destination, and delivery are three independent objects created in sequence. A source can exist without a delivery connecting it to a destination — the docs state: *"Use `CreateDelivery` to create a delivery by pairing exactly one delivery source and one delivery destination."* Without an active delivery, logs do not flow.

This method only calls `get_delivery_source`. It will return `configured: True` even when the delivery or destination is missing (e.g., if `enable_observability_for_resource` crashed midway).

**Fix:** also check that at least one active delivery exists linking source to destination via `describe_deliveries`.

jariy17 (changes requested):

PR #446, opened 2026-04-30
Feat: Evaluation preview features - Batch evaluation and config bundles 

padmak30:
*Description of changes:*

**Batch Evaluation**
- Run batch evaluation on dataset
- Ground truth support
- Simulated dataset support

**ConfigBundles**
- Config bundle support - read config bundle values from baggage
- BaggageSpanProcessor and routing experiment context helpers 
              


notgitika:
Couple things I noticed:

1. `BatchEvaluationResult` doesn't have a `description` field, but `run_dataset_evaluation()` passes `description=response.get("description")` when building it. Pydantic v2 will throw a `ValidationError` for the unexpected field — this would crash on every successful batch eval. Needs `description: Optional[str] = None` added to the model.

2. The timeout test (`test_poll_for_results_timeout_raises_timeout_error`) patches `time.time` to control the clock, but `_poll_for_results` actually uses `time.monotonic()`. The test only passes because the real monotonic clock advances past the timeout between iterations — it's not actually testing the timeout path. The patch target should be `time.monotonic` instead.

notgitika (dismissed):

jariy17 (dismissed):

notgitika (approved):
LGTM

PR #449, opened 2026-05-01
Update DP APIs to support namespace re-design

nborges-aws:
### Summary:
Adds support for the new namespacePath field introduced on the memory data-plane APIs (ListMemoryRecords and RetrieveMemoryRecords). This lets callers opt into hierarchical path-prefix retrieval, separate from the existing namespace field (which is evolving toward exact-match semantics service-side).

 ### Description of changes:
New utility (`src/bedrock_agentcore/_utils/namespace.py`):
- build_namespace_params(namespace, namespace_path) — validates caller input (mutual exclusivity, presence, no wildcards) and produces the correct wire kwargs for boto.

MemoryClient (`memory/client.py`):
- retrieve_memories() — accepts both namespace and namespace_path. On invalid input (both provided, neither provided, wildcard present), logs an error and returns [] — preserves the existing soft-fail contract of this method.
- process_turn_with_llm() — adds retrieval_namespace_path parameter, forwards to retrieve_memories.

MemorySessionManager (`memory/session.py`):
- search_long_term_memories() and list_long_term_memory_records() now route the existing namespace_prefix argument to namespacePath. The parameter name and user-facing semantics remain unchanged; only the underlying field is updated.

Strands integration (`memory/integrations/strands/session_manager.py`):
- _load_ltm() now calls retrieve_memories(namespace_path=...) to preserve the historical hierarchical-scoping behavior under the new API semantics.

### Behavior changes for users

- Existing callers using retrieve_memories(namespace=...) continue to work. Namespace will eventually become the exact-match field
- Existing callers using search_long_term_memories(namespace_prefix=...) continue to work as before
- New callers can opt into hierarchical retrieval explicitly via retrieve_memories(namespace_path=...).

### DIFF
--- a/src/bedrock_agentcore/_utils/namespace.py
+++ b/src/bedrock_agentcore/_utils/namespace.py
@@ -1,0 +1,28 @@
+"""Namespace utilities for data plane API calls."""
+
+from typing import Dict, Optional
+
+
+def build_namespace_params(namespace: Optional[str] = None, namespace_path: Optional[str] = None) -> Dict[str, str]:
+    """Build the namespace kwargs for a data plane API call.
+
+    Exactly one of ``namespace`` (exact match) or ``namespace_path``
+    (hierarchical path prefix) must be provided. Wildcards (``*``) are not
+    supported in either field.
+
+    Raises:
+        ValueError: if both arguments are provided, neither is provided, or
+            the provided value contains a wildcard.
+    """
+    if namespace is not None and namespace_path is not None:
+        raise ValueError("'namespace' and 'namespace_path' are mutually exclusive.")
+    if namespace is None and namespace_path is None:
+        raise ValueError("At least one of 'namespace' or 'namespace_path' must be provided.")
+
+    value = namespace if namespace is not None else namespace_path
+    if "*" in value:
+        raise ValueError("Wildcards (*) are not supported in namespaces.")
+
+    if namespace is not None:
+        return {"namespace": namespace}
+    return {"namespacePath": namespace_path}
--- a/src/bedrock_agentcore/_utils/namespace.py
+++ b/src/bedrock_agentcore/_utils/namespace.py
@@ -73,3 +73,37 @@ def resolve_namespace_templates(
         return namespaces
 
     return namespace_templates
+
+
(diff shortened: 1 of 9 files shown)


src/bedrock_agentcore/_utils/namespace.py
```diff
@@ -0,0 +1,28 @@
+"""Namespace utilities for data plane API calls."""
+
+from typing import Dict, Optional
+
+
+def build_namespace_params(namespace: Optional[str] = None, namespace_path: Optional[str] = None) -> Dict[str, str]:
```
Hweinstock, line 7: q: is namespace a concept outside of memory? If not, does it make sense for this to live within the memory subfolder?

tests/unit/test_utils.py
```diff
@@ -112,3 +113,29 @@ def test_timeout(self, _mock_time, _mock_sleep):
         poll_fn = Mock(return_value={"status": "DELETING"})
         with pytest.raises(TimeoutError):
             wait_until_deleted(poll_fn)
+
+
+class TestBuildNamespaceParams:
```
Hweinstock, line 122: nit: I feel like we have some duplicate coverage of the functionality also tested in tests/bedrock_agentcore/memory/test_client.py. I actually prefer this one since it tests the utils directly, instead of mocking. Does it make sense to consolidate here?

Hweinstock (dismissed):
One question about namespace usage, otherwise looks good.

src/bedrock_agentcore/_utils/namespace.py
nborges-aws: Yeah this is a memory level concept; I can move utils into the memory folder. Let's do this as a fast-follow after merging open PRs. Given the namespace changes are split over two PRs updating will be easier post-merge.

tests/unit/test_utils.py
nborges-aws: Yeah, there are a few tests that are similar between these two. My rationale was that the tests in utils are verifying the utility method works as intended, and the tests in the client ensures the end user's experience (user returned`[]`, sees correct log message).

jesseturner21 (approved):

PR #450, opened 2026-05-01
chore: update CP APIs to support namespace re-design

nborges-aws:
### Summary: 
Migrates the memory control-plane to the new namespaceTemplates field, deprecating the legacy namespaces field while maintaining full backward compatibility.

### Description of changes:
* MemoryClient (memory/client.py) — 13 methods updated with parallel namespace_templates kwargs
* MemoryControlPlaneClient (memory/controlplane.py) — parallel namepsace_templates kwargs
* Episodic methods gained reflection_namespace_templates, deprecating reflection_namespaces

### Behavior changes for users
- **No breaking changes.** All existing code using namespaces= / reflection_namespaces= continues to work, with a DeprecationWarning surfaced to encourage migration.
- Calling a method with both namespaces= and namespace_templates= raises ValueError with a message pointing users to the new name.

### Testing
- New TestResolveNamespaceTemplates unit tests cover the happy paths, mutual exclusivity, and deprecation warning emission
- Existing method tests updated to assert on the new namespaceTemplates key rather than namespaces. Existing legacy-kwarg callers in tests are preserved as DeprecationWarning-triggering coverage for the backward-compatible path.
- Full suite passing : 1706 passed, 0 failures. 16 DeprecationWarnings are expected — messaging for legacy namespace field

### DIFF
--- a/src/bedrock_agentcore/_utils/namespace.py
+++ b/src/bedrock_agentcore/_utils/namespace.py
@@ -1,0 +1,75 @@
+"""Namespace utilities for data plane and control plane API calls."""
+
+import warnings
+from typing import Dict, List, Optional
+
+
+def build_namespace_params(namespace: Optional[str] = None, namespace_path: Optional[str] = None) -> Dict[str, str]:
+    """Build the namespace kwargs for a data plane API call.
+
+    Exactly one of ``namespace`` (exact match) or ``namespace_path``
+    (hierarchical path prefix) must be provided. Wildcards (``*``) are not
+    supported in either field.
+
+    Raises:
+        ValueError: if both arguments are provided, neither is provided, or
+            the provided value contains a wildcard.
+    """
+    if namespace is not None and namespace_path is not None:
+        raise ValueError("'namespace' and 'namespace_path' are mutually exclusive.")
+    if namespace is None and namespace_path is None:
+        raise ValueError("At least one of 'namespace' or 'namespace_path' must be provided.")
+
+    value = namespace if namespace is not None else namespace_path
+    if "*" in value:
+        raise ValueError("Wildcards (*) are not supported in namespaces.")
+
+    if namespace is not None:
+        return {"namespace": namespace}
+    return {"namespacePath": namespace_path}
+
+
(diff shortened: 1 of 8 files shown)


src/bedrock_agentcore/_utils/namespace.py
```diff
@@ -0,0 +1,75 @@
+"""Namespace utilities for data plane and control plane API calls."""
```
Hweinstock, line 1: same question: https://github.com/aws/bedrock-agentcore-sdk-python/pull/449#discussion_r3220643934

src/bedrock_agentcore/memory/client.py
```diff
@@ -1436,54 +1490,87 @@ def add_user_preference_strategy_and_wait(
...
+        namespace_templates: Optional[List[str]] = None,
+        reflection_namespace_templates: Optional[List[str]] = None,
     ) -> Dict[str, Any]:
         """Add an episodic memory strategy.
 
         Args:
             memory_id: Memory resource ID
             name: Strategy name
-            reflection_namespaces: Namespaces for reflections (can be less nested than episode namespaces)
+            reflection_namespaces: DEPRECATED. Use ``reflection_namespace_templates`` instead.
```
Hweinstock, line 1520: What does it mean that this is deprecated? Does it still work? Are we still maintaining it?

src/bedrock_agentcore/_utils/namespace.py
nborges-aws: Answered in PR #449 thread

src/bedrock_agentcore/memory/client.py
nborges-aws: When a method is called using the deprecated field, a deprecation warning will be surfaced to the user. The field still works, and on the service-side will function identically to its old behavior until the deprecation window closes.

During this time, we will continue maintaining the deprecated field. When the window closes, the service will switch the field to new behavior, and we will remove need to remove the deprecated field.

Hweinstock (approved):

PR #454, opened 2026-05-05
fix: preserve multi-turn history in Strands ConversationTurn

tejaskash:
## Summary

`StrandsEventParser.extract_conversation_turn` silently drops conversation history when a span carries multiple `gen_ai.*.message` events — common on every multi-turn agent invocation. This PR preserves the full history, separates prior assistant turns from the current model output, and keeps the existing API working via a deprecation shim.

## The bug

Two interacting defects in `span_to_adot_serializer/strands_converter.py`:

**1. User-message collapse** — `extract_conversation_turn` stored `gen_ai.user.message` content in a scalar:

```python
case cls.EVENT_USER_MESSAGE:
    user_message = event_attrs.get("content", "")  # overwrites on every event
```

A span carrying N `gen_ai.user.message` events (the conversation history a Strands agent replays to the model) would retain only the **last** one. For a 9-turn conversation, 8 user turns were silently dropped before the span was ever serialized.

**2. Assistant-message miscategorization** — `gen_ai.assistant.message` events and `gen_ai.choice` events were both `.append()`ed to the same `assistant_messages` list:

```python
case cls.EVENT_CHOICE:           # current-turn model output
    ...
    assistant_messages.append(...)
case cls.EVENT_ASSISTANT_MESSAGE:  # prior-turn history input
    ...
    assistant_messages.append(...)
```

Per [OTel GenAI semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/), `gen_ai.{system,user,assistant,tool}.message` events describe **input context** (including prior assistant turns the model reads as history), while `gen_ai.choice` describes the **current-turn output**. Mixing them means downstream consumers — `ADOTDocumentBuilder.build_conversation_log_record`, Online Evaluations, Observability dashboards — cannot tell history apart from new output and receive a ConversationTurn that misrepresents what the model actually did.

## Reproduction

A span from a 3-turn conversation (3 user msgs, 2 prior assistant msgs, 1 new choice) produced a `ConversationTurn` with:

| Field | Expected | Actual (before fix) |
|---|---|---|
| user content retained | 3 turns | **1** (newest only) |
| `assistant_messages` | 1 entry (choice) | **3** (2 history + choice) |
| `history_messages` | 2 entries | field did not exist |

## Fix

**`ConversationTurn`** (`adot_models.py`) adds two fields:
- `user_messages: List[str]` — every user turn observed on the span, in order.
- `history_messages: List[Dict]` — prior assistant turns from `gen_ai.assistant.message` events, kept separate from the current-turn output.

`user_message` is retained as a `@property` that returns the most recent entry and emits a `DeprecationWarning` when it truncates more than one turn, so existing callers keep working while being nudged toward `user_messages`. Constructing `ConversationTurn(user_message=...)` still works — the initializer accepts either the legacy scalar or the new list.

**`StrandsEventParser.extract_conversation_turn`** (`strands_converter.py`) appends every `gen_ai.user.message` to `user_messages`, routes `gen_ai.assistant.message` to `history_messages`, and keeps `assistant_messages` scoped to `gen_ai.choice` output only.

**`ADOTDocumentBuilder.build_conversation_log_record`** (`adot_models.py`) emits all user turns followed by all history assistant turns as `input.messages`, so the serialized ADOT record carries the full context the model actually received.

## Tests

- Updated `test_extract_conversation_turn_assistant_message` — `gen_ai.assistant.message` now lands in `history_messages`, not `assistant_messages`.
- New `test_extract_conversation_turn_preserves_all_user_messages` — multiple `gen_ai.user.message` events all retained.
- New `test_extract_conversation_turn_separates_history_from_choice` — the end-to-end multi-turn case from the bug report.
- New `ConversationTurn` dataclass tests: legacy scalar init, list init, deprecation-warning behavior, rejection of both-at-once, history preservation.
- New `test_build_conversation_log_record_multi_turn_history` — builder emits `[user, user, user, assistant-history, assistant-history]` on input and single-entry output.

Full suite: **1726 passed, 0 failed**. Serializer module: **41/41**. ruff, ruff format, and pre-commit clean.

## Backwards compatibility

- `ConversationTurn(user_message=...)` constructor form still works.
- `turn.user_message` still returns a string (the most recent user turn).
- ADOT wire format `body.input.messages` / `body.output.messages` shape is unchanged; `input.messages` just contains more entries on multi-turn spans.
- Consumers that read `assistant_messages` now get only the current-turn output. If a caller previously relied on the buggy behavior where history messages showed up in `assistant_messages`, they should migrate to `history_messages`. This is the intended correctness change.

## Related

There is a companion bug with the same failure mode in `aws-opentelemetry-distro`'s `LLOHandler.process_spans` (`llo_handler.py:329` dedupes events by name; `llo_handler.py:148-152` tags `gen_ai.assistant.message` with `source: "output"`). That needs a separate fix in the `aws-observability/aws-otel-python-instrumentation` repo; this PR only addresses the AgentCore SDK side.

## Test plan

- [x] Existing serializer tests pass unchanged (where correct) or updated (where they asserted the buggy behavior)
- [x] New regression tests cover multi-turn user, multi-turn history, and mixed-role preservation
- [x] Full `tests/` suite passes (1726 passed)
- [x] `ruff check` / `ruff format` / pre-commit hooks clean
- [x] End-to-end repro shows `user_messages` holds all turns and `assistant_messages` holds only the current `gen_ai.choice`

### DIFF
--- a/src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
+++ b/src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
@@ -10,7 +10,8 @@ telemetry frameworks (Strands, LangGraph, etc.).
 """
 
 import logging
-from dataclasses import dataclass
+import warnings
+from dataclasses import dataclass, field
 from typing import Any, Dict, List, Optional
 
 logger = logging.getLogger(__name__)
@@ -48,11 +49,67 @@ class ResourceInfo:
 
 @dataclass
 class ConversationTurn:
-    """A single user-assistant conversation turn."""
+    """A single user-assistant conversation turn, with prior history preserved.
 
-    user_message: str
-    assistant_messages: List[Dict[str, Any]]
-    tool_results: List[str]
+    ``user_messages`` holds every ``gen_ai.user.message`` event observed on the
+    span (conversation history + the current user turn). ``history_messages``
+    holds every ``gen_ai.assistant.message`` event — these are prior assistant
+    turns replayed as input context, not the current model output.
+    ``assistant_messages`` holds only the current turn's output, derived from
+    ``gen_ai.choice`` events.
+
+    The ``user_message`` attribute is retained as a backwards-compatible alias
+    returning the most recent user turn; new code should read ``user_messages``.
+    """
+
+    user_messages: List[str] = field(default_factory=list)
+    assistant_messages: List[Dict[str, Any]] = field(default_factory=list)
+    tool_results: List[str] = field(default_factory=list)
+    history_messages: List[Dict[str, Any]] = field(default_factory=list)
+
+    def __init__(
+        self,
+        user_message: Optional[str] = None,
+        assistant_messages: Optional[List[Dict[str, Any]]] = None,
+        tool_results: Optional[List[str]] = None,
+        user_messages: Optional[List[str]] = None,
+        history_messages: Optional[List[Dict[str, Any]]] = None,
+    ):
+        """Initialize a conversation turn.
+
+        Accepts either the new ``user_messages`` list or the legacy
+        ``user_message`` scalar for backwards compatibility.
+        """
+        if user_messages is not None and user_message is not None:
+            raise ValueError("Provide either user_messages or user_message, not both")
+        if user_messages is not None:
+            self.user_messages = list(user_messages)
+        elif user_message is not None:
+            self.user_messages = [user_message]
+        else:
+            self.user_messages = []
+        self.assistant_messages = list(assistant_messages) if assistant_messages is not None else []
+        self.tool_results = list(tool_results) if tool_results is not None else []
+        self.history_messages = list(history_messages) if history_messages is not None else []
+
+    @property
+    def user_message(self) -> str:
+        """Return the most recent user turn (backwards-compatible alias).
+
+        Deprecated: read ``user_messages`` to access the full list of user
+        turns on the span. This alias drops all but the last entry.
+        """
+        if not self.user_messages:
+            return ""
+        if len(self.user_messages) > 1:
+            warnings.warn(
+                "ConversationTurn.user_message drops prior turns when the span "
+                "carries multiple gen_ai.user.message events. Read "
+                "ConversationTurn.user_messages for the full list.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+        return self.user_messages[-1]
 
 
 @dataclass
@@ -191,7 +248,13 @@ class ADOTDocumentBuilder:
         metadata: SpanMetadata,
         resource_info: ResourceInfo,
     ) -> Dict[str, Any]:
-        """Build ADOT log record for conversation turn."""
+        """Build ADOT log record for conversation turn.
+
+        Input messages preserve conversation history: user turns and prior
+        assistant turns (``history_messages``) are merged in the order they
+        were observed on the span so downstream consumers see the full context
+        the model received. Output messages carry only the current turn.
+        """
         output_messages = []
         for i, msg in enumerate(conversation.assistant_messages):
             output_msg = msg.copy()
@@ -204,9 +267,14 @@ class ADOTDocumentBuilder:
         for tool_result in conversation.tool_results:
             output_messages.append({"content": tool_result, "role": "assistant"})
 
+        input_messages: List[Dict[str, Any]] = [
+            {"content": {"content": user_msg}, "role": "user"} for user_msg in conversation.user_messages
+        ]
+        input_messages.extend(conversation.history_messages)
+
         body = {
             "output": {"messages": output_messages},
-            "input": {"messages": [{"content": {"content": conversation.user_message}, "role": "user"}]},
+            "input": {"messages": input_messages},
         }
 
         return cls._build_log_record_base(metadata, resource_info, body)
--- a/src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
+++ b/src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
@@ -11,7 +11,7 @@ telemetry frameworks (Strands, LangGraph, etc.).
 
 import logging
 import warnings
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import Any, Dict, List, Optional
 
 logger = logging.getLogger(__name__)
@@ -47,69 +47,89 @@ class ResourceInfo:
     scope_version: str
 
 
-@dataclass
 class ConversationTurn:
     """A single user-assistant conversation turn, with prior history preserved.
 
-    ``user_messages`` holds every ``gen_ai.user.message`` event observed on the
-    span (conversation history + the current user turn). ``history_messages``
-    holds every ``gen_ai.assistant.message`` event — these are prior assistant
-    turns replayed as input context, not the current model output.
+    ``input_messages`` holds every input event (``gen_ai.user.message`` and
+    ``gen_ai.assistant.message``) in the order observed on the span, so
+    downstream consumers can reconstruct the actual conversation flow.
     ``assistant_messages`` holds only the current turn's output, derived from
-    ``gen_ai.choice`` events.
+    ``gen_ai.choice`` events. ``tool_results`` holds any tool outputs observed
+    on the span.
 
     The ``user_message`` attribute is retained as a backwards-compatible alias
-    returning the most recent user turn; new code should read ``user_messages``.
+    returning the most recent user turn's content; new code should iterate
+    ``input_messages`` filtered by role.
     """
-
-    user_messages: List[str] = field(default_factory=list)
-    assistant_messages: List[Dict[str, Any]] = field(default_factory=list)
-    tool_results: List[str] = field(default_factory=list)
-    history_messages: List[Dict[str, Any]] = field(default_factory=list)
 
     def __init__(
         self,
         user_message: Optional[str] = None,
         assistant_messages: Optional[List[Dict[str, Any]]] = None,
         tool_results: Optional[List[str]] = None,
-        user_messages: Optional[List[str]] = None,
-        history_messages: Optional[List[Dict[str, Any]]] = None,
+        input_messages: Optional[List[Dict[str, Any]]] = None,
     ):
         """Initialize a conversation turn.
 
-        Accepts either the new ``user_messages`` list or the legacy
-        ``user_message`` scalar for backwards compatibility.
+        ``input_messages`` is the chronological list of input events
+        (``{"role": "user"|"assistant", "content": {...}}``). For backwards
+        compatibility, callers may instead pass a legacy ``user_message``
+        scalar, which is converted into a single-entry ``input_messages``
+        list. Supplying both is an error.
         """
-        if user_messages is not None and user_message is not None:
-            raise ValueError("Provide either user_messages or user_message, not both")
-        if user_messages is not None:
-            self.user_messages = list(user_messages)
+        if input_messages is not None and user_message is not None:
+            raise ValueError("Provide either input_messages or user_message, not both")
+        if input_messages is not None:
+            self.input_messages = list(input_messages)
         elif user_message is not None:
-            self.user_messages = [user_message]
+            self.input_messages = [{"content": {"content": user_message}, "role": "user"}]
         else:
-            self.user_messages = []
-        self.assistant_messages = list(assistant_messages) if assistant_messages is not None else []
-        self.tool_results = list(tool_results) if tool_results is not None else []
-        self.history_messages = list(history_messages) if history_messages is not None else []
+            self.input_messages = []
+        self.assistant_messages = list(assistant_messages or [])
+        self.tool_results = list(tool_results or [])
+
+    def __repr__(self) -> str:
+        """Return a debug representation listing every stored field."""
+        return (
+            f"ConversationTurn(input_messages={self.input_messages!r}, "
+            f"assistant_messages={self.assistant_messages!r}, "
+            f"tool_results={self.tool_results!r})"
+        )
+
+    def __eq__(self, other: object) -> bool:
+        """Compare turns by the full set of instance fields."""
+        if not isinstance(other, ConversationTurn):
+            return NotImplemented
+        return (
+            self.input_messages == other.input_messages
+            and self.assistant_messages == other.assistant_messages
+            and self.tool_results == other.tool_results
+        )
 
     @property
     def user_message(self) -> str:
-        """Return the most recent user turn (backwards-compatible alias).
+        """Return the most recent user turn's content (backwards-compatible alias).
 
-        Deprecated: read ``user_messages`` to access the full list of user
-        turns on the span. This alias drops all but the last entry.
+        Deprecated: iterate ``input_messages`` to access the full conversation
+        history. This alias returns only the last user entry.
         """
-        if not self.user_messages:
+        user_entries = [m for m in self.input_messages if m.get("role") == "user"]
+        if not user_entries:
             return ""
-        if len(self.user_messages) > 1:
+        if len(user_entries) > 1 or any(m.get("role") == "assistant" for m in self.input_messages):
             warnings.warn(
-                "ConversationTurn.user_message drops prior turns when the span "
-                "carries multiple gen_ai.user.message events. Read "
-                "ConversationTurn.user_messages for the full list.",
+                "ConversationTurn.user_message drops prior user turns and history "
+                "when the span carries multiple input events. Iterate "
+                "ConversationTurn.input_messages for the full conversation.",
                 DeprecationWarning,
                 stacklevel=2,
             )
-        return self.user_messages[-1]
+        last = user_entries[-1]
+        content = last.get("content")
+        if isinstance(content, dict):
+            inner = content.get("content", "")
+            return inner if isinstance(inner, str) else ""
+        return content if isinstance(content, str) else ""
 
 
 @dataclass
@@ -250,10 +270,11 @@ class ADOTDocumentBuilder:
     ) -> Dict[str, Any]:
         """Build ADOT log record for conversation turn.
 
-        Input messages preserve conversation history: user turns and prior
-        assistant turns (``history_messages``) are merged in the order they
-        were observed on the span so downstream consumers see the full context
-        the model received. Output messages carry only the current turn.
+        ``body.input.messages`` carries ``ConversationTurn.input_messages`` in
+        event arrival order, so downstream consumers reconstruct the exact
+        conversation the model received (user and prior assistant turns
+        interleaved). ``body.output.messages`` carries only the current turn's
+        output.
         """
         output_messages = []
         for i, msg in enumerate(conversation.assistant_messages):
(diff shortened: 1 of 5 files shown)


tejaskash:
Thanks, bot. Walking through the three flags:

1. **`ConversationTurn.assistant_messages: None -> list(assistant_messages) if assistant_messages is not None else []`** — false positive. This is a signature diff, not a value change. The class moved from an auto-generated `@dataclass __init__` (where the field annotation was `List[Dict[str, Any]]` with no default, so callers had to pass a value) to a custom `__init__` that defaults missing args to `[]`. Existing callers that passed `assistant_messages=...` get identical behavior; callers that omitted it previously raised `TypeError` and now get an empty list. Not a breaking change.

2. **`ConversationTurn.tool_results: None -> list(tool_results) if tool_results is not None else []`** — same as above. Signature change from required-positional to defaulted, non-breaking.

3. **`ConversationTurn.__init__: Return types are incompatible: None -> None`** — false positive, the bot is diffing an auto-synthesized dataclass `__init__` against the explicit one. Both return `None`.

The checker did miss the one **intentional** behavior change worth calling out: `assistant_messages` no longer contains prior `gen_ai.assistant.message` turns — those move to the new `history_messages` field. This is the bug fix (per OTel GenAI semconv, those events are input context, not current-turn output). Callers that iterated `assistant_messages` expecting both should either read both fields, or read `body.input.messages` from `build_conversation_log_record`, which now carries the full history. The legacy `user_message` scalar accessor and constructor form continue to work via a deprecation shim.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/strands_converter.py
```diff
@@ -66,18 +78,19 @@ def extract_conversation_turn(cls, events: List[Any]) -> Optional[ConversationTu
...
-                        assistant_messages.append({"content": {"content": content}, "role": "assistant"})
+                        history_messages.append({"content": {"content": content}, "role": "assistant"})
 
                 case cls.EVENT_TOOL_MESSAGE:
                     content = event_attrs.get("content", "")
                     if content:
                         tool_results.append(content)
 
-        if user_message and assistant_messages:
+        if user_messages and (assistant_messages or history_messages):
```
aidandaly24: nit: The condition changed from `user_message and assistant_messages` to `user_messages and (assistant_messages or history_messages)`. This means a span with `user_messages` + `tool_results` but no `assistant_messages` or `history_messages` now returns `None` — which could drop valid tool-use-only turns that previously returned a `ConversationTurn` (when history messages were incorrectly bucketed into `assistant_messages`).

Might be worth adding a test that confirms the intended behavior for a user-message + tool-result-only span (no choice, no history), and deciding whether `tool_results` should also satisfy this guard.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
```diff
@@ -204,9 +267,14 @@ def build_conversation_log_record(
         for tool_result in conversation.tool_results:
             output_messages.append({"content": tool_result, "role": "assistant"})
 
+        input_messages: List[Dict[str, Any]] = [
+            {"content": {"content": user_msg}, "role": "user"} for user_msg in conversation.user_messages
+        ]
+        input_messages.extend(conversation.history_messages)
```
aidandaly24: nit: This serializes input as `[user, user, user, assistant-history, assistant-history]` rather than preserving the original interleaving (`[user, assistant, user, assistant, user]`). The test asserts this flat order so it seems intentional — but worth confirming with the Observability/Evaluations team whether downstream consumers expect role-grouped input or chronological interleaving. If they ever need to reconstruct the actual conversation flow, this ordering makes it impossible.

aidandaly24 (dismissed):
Clean fix for a real data-loss bug. Backwards compatibility is well handled, tests are thorough. Two minor nits inline — neither is blocking.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
```diff
@@ -48,11 +49,67 @@ class ResourceInfo:
 
 @dataclass
```
jariy17: `@dataclass` + custom `__init__` is fighting itself. The `field(default_factory=list)` declarations below are dead — `__init__` overwrites them immediately. Also, `__eq__` and `__repr__` operate on the unpopulated class-level fields, not the actual instance values, so equality checks may pass vacuously. Switch to `@dataclass(init=False)` or drop the decorator and make this a plain class.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
```diff
@@ -48,11 +49,67 @@ class ResourceInfo:
...
+    holds every ``gen_ai.assistant.message`` event — these are prior assistant
+    turns replayed as input context, not the current model output.
+    ``assistant_messages`` holds only the current turn's output, derived from
+    ``gen_ai.choice`` events.
+
+    The ``user_message`` attribute is retained as a backwards-compatible alias
+    returning the most recent user turn; new code should read ``user_messages``.
+    """
+
+    user_messages: List[str] = field(default_factory=list)
```
jariy17: These four `field(default_factory=list)` declarations are dead weight — the custom `__init__` below sets the instance vars directly, so these are never used. Delete them.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
```diff
@@ -48,11 +49,67 @@ class ResourceInfo:
...
+        """
+        if user_messages is not None and user_message is not None:
+            raise ValueError("Provide either user_messages or user_message, not both")
+        if user_messages is not None:
+            self.user_messages = list(user_messages)
+        elif user_message is not None:
+            self.user_messages = [user_message]
+        else:
+            self.user_messages = []
+        self.assistant_messages = list(assistant_messages) if assistant_messages is not None else []
```
jariy17: nit: Same pattern copy-pasted three times: `list(x) if x is not None else []`. Collapse to `list(x or [])` for all three.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
```diff
@@ -204,9 +267,14 @@ def build_conversation_log_record(
         for tool_result in conversation.tool_results:
             output_messages.append({"content": tool_result, "role": "assistant"})
 
+        input_messages: List[Dict[str, Any]] = [
```
jariy17: Ordering is wrong and contradicts the docstring. The docstring says events come out "in the order they were observed on the span", but this emits all user messages first, then all history assistants. For a real stream `[u1, a1, u2, a2, u3, choice]`, downstream gets `[user, user, user, assistant, assistant]` — that's not a real conversation.

The `user_messages: List[str]` + `history_messages: List[Dict]` split can't represent interleaved order. Fix: collect `(role, content)` tuples in event arrival order during parsing and pass a single ordered list through `ConversationTurn`.

Also `test_adot_models.py:363` hard-codes the wrong order as expected — that's a regression anchor for the bug. Fix that test too.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/strands_converter.py
```diff
@@ -37,17 +37,29 @@ class StrandsEventParser:
...
+        tool_results: List[str] = []
 
         for event in events:
             event_attrs = dict(event.attributes) if hasattr(event, "attributes") and event.attributes else {}
 
             match event.name:
                 case cls.EVENT_USER_MESSAGE:
-                    user_message = event_attrs.get("content", "")
+                    content = event_attrs.get("content", "")
+                    if content:
```
jariy17, line 59: `if content:` silently drops empty events. The old scalar path had different behavior — `user_message = ""` would fail the outer `if user_message` gate and reject the whole turn. Now empty events are skipped individually. If the intent is to drop empty content, add a log line so it's visible.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/strands_converter.py
jariy17: Return condition was broadened — history alone now satisfies the gate. A span with `[user.message, assistant.message]` and no choice event now returns a `ConversationTurn` with empty `assistant_messages`, which means `build_conversation_log_record` emits `output.messages: []`. Downstream eval consumers weren't designed for that shape.

Let's add a test for this path before merge and confirm the eval API accepts empty output records — or tighten the gate back to require `assistant_messages`.

tests/bedrock_agentcore/evaluation/span_to_adot_serializer/test_adot_models.py
```diff
@@ -163,15 +163,53 @@ class TestConversationTurn:
...
+        """ConversationTurn accepts a list of user messages preserving order."""
+        turn = ConversationTurn(
+            user_messages=["first", "second", "third"],
+            assistant_messages=[{"content": {"message": "ok"}, "role": "assistant"}],
+        )
+        assert turn.user_messages == ["first", "second", "third"]
+
+    def test_user_message_alias_returns_latest(self):
+        """user_message returns the latest entry; accessing it with history emits a DeprecationWarning."""
+        import warnings
```
jariy17: `import warnings` inside the test body — move it to the module-level imports. Same issue at line ~199 with `import pytest`, which is already imported at line 5.

tests/bedrock_agentcore/evaluation/span_to_adot_serializer/test_adot_models.py
```diff
@@ -308,6 +346,33 @@ def test_build_conversation_log_record(self, span_metadata, resource_info):
...
+            history_messages=[
+                {"content": {"content": "prior-1"}, "role": "assistant"},
+                {"content": {"content": "prior-2"}, "role": "assistant"},
+            ],
+        )
+
+        doc = ADOTDocumentBuilder.build_conversation_log_record(conversation, span_metadata, resource_info)
+
+        input_msgs = doc["body"]["input"]["messages"]
+        assert [m.get("role") for m in input_msgs] == ["user", "user", "user", "assistant", "assistant"]
```
jariy17: This test hard-codes the broken ordering as the expected value. `[user, user, user, assistant, assistant]` is wrong for a `[u1, a1, u2, a2, u3, choice]` input — the correct order is `[user, assistant, user, assistant, user]`. Fix the ordering bug and update this assertion, otherwise the test is locking in the bug.

tests/bedrock_agentcore/evaluation/span_to_adot_serializer/test_strands_converter.py
```diff
@@ -129,7 +129,43 @@ def test_extract_conversation_turn_assistant_message(self, mock_event):
         turn = StrandsEventParser.extract_conversation_turn(events)
```
jariy17, line 55: Let's add a test for the no-choice path here — `[user.message, assistant.message]` with no `gen_ai.choice` event. The broadened gate enables it now but there's no coverage for what `output.messages` looks like in that case.

tests/bedrock_agentcore/evaluation/span_to_adot_serializer/test_strands_converter.py
```diff
@@ -129,7 +129,43 @@ def test_extract_conversation_turn_assistant_message(self, mock_event):
...
+        """Multiple gen_ai.user.message events are all preserved, not deduped."""
+        events = [
+            mock_event("gen_ai.user.message", {"content": "first"}),
+            mock_event("gen_ai.user.message", {"content": "second"}),
+            mock_event("gen_ai.user.message", {"content": "third"}),
+            mock_event("gen_ai.choice", {"message": "ok"}),
+        ]
+
+        turn = StrandsEventParser.extract_conversation_turn(events)
+
```
jariy17, line 79: Fixtures at the top of this file (lines 18–70) are a copy of the ones in `test_adot_models.py`. Move them to a shared `conftest.py` so they don't drift.

Also, `test_extract_conversation_turn` at line 94 is still calling `turn.user_message` (the deprecated property) — update it to assert `turn.user_messages[0]` instead.

jariy17 (changes requested):

src/bedrock_agentcore/evaluation/span_to_adot_serializer/strands_converter.py
notgitika: ^ agree

jariy17:
> ## ⚠️ Breaking Change Warning
> 
> Found **3** potential breaking change(s) in this PR:
> 
> �[1msrc/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py�[0m:91: ConversationTurn.assistant_messages: �[33mAttribute value was changed�[39m: None -> list(assistant_messages) if assistant_messages is not None else [] �[1msrc/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py�[0m:92: ConversationTurn.tool_results: �[33mAttribute value was changed�[39m: None -> list(tool_results) if tool_results is not None else [] �[1msrc/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py�[0m:70: ConversationTurn.**init**: �[33mReturn types are incompatible�[39m: None -> None
> 
> > **Note:** This is an automated static analysis check. Some flagged changes may be intentional.
> > Please confirm each item is expected and, if so, add a migration note to `CHANGELOG.md`.

look into these too

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
tejaskash: Fixed. Dropped `@dataclass` entirely on `ConversationTurn`, added explicit `__init__`, `__repr__`, `__eq__` so equality/repr operate on real instance state. Other dataclasses in the file (`SpanMetadata`, `ResourceInfo`, `ToolExecution`) were always fine — they use the auto-generated `__init__`.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
tejaskash: Gone. The new class is a plain class with explicit attribute assignment in `__init__`.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
tejaskash: Done — collapsed to `list(x or [])`.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
tejaskash: Fixed. `ConversationTurn` now has a single `input_messages: List[Dict]` field; the parser collects events in arrival order so `[u1, a1, u2, a2, u3]` serializes as `[user, assistant, user, assistant, user]`. Dropped `user_messages` and `history_messages` fields; consumers that want just user or just history entries filter by role. Test that anchored the wrong order (`test_adot_models.py:363`) is updated to assert chronological interleaving.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/strands_converter.py
tejaskash: Added `logger.debug(...)` on all three empty-content skip paths (user, assistant, tool). The outer-gate semantics are also restored — a parse that yields only an empty user event now fails the gate (`has_user_input` requires a non-empty `input_messages` entry with role=user), matching the old `if user_message` behavior.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/strands_converter.py
tejaskash: Tightened the gate back to require `assistant_messages`. A span with `[user.message, assistant.message]` and no choice event now returns `None` (no empty-output `ConversationTurn`). Added `test_extract_conversation_turn_assistant_message_only_returns_none` to lock this in, and `test_extract_conversation_turn_tool_results_only_returns_none` for the user+tool_result-only case Aidan raised separately.

tests/bedrock_agentcore/evaluation/span_to_adot_serializer/test_adot_models.py
tejaskash: Hoisted `import warnings` to the module-level imports. `import pytest` was already imported at line 5 — the duplicate inside the test body is gone.

tests/bedrock_agentcore/evaluation/span_to_adot_serializer/test_adot_models.py
tejaskash: Fixed. The test now asserts chronological order `[user, assistant, user, assistant, user]` and matches the new `input_messages` field.

tests/bedrock_agentcore/evaluation/span_to_adot_serializer/test_strands_converter.py
tejaskash: Added `test_extract_conversation_turn_assistant_message_only_returns_none`. Since the gate now requires `assistant_messages`, the no-choice path returns `None` and never reaches `build_conversation_log_record`.

tests/bedrock_agentcore/evaluation/span_to_adot_serializer/test_strands_converter.py
tejaskash: Moved the duplicated fixtures (`mock_span_context`, `mock_resource`, `mock_instrumentation_scope`, `mock_status`, `mock_span`, `mock_event`, `span_metadata`, `resource_info`) into a new `conftest.py`. `test_extract_conversation_turn` now asserts `turn.input_messages == [...]` instead of the deprecated `turn.user_message`.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/strands_converter.py
tejaskash: Fixed — gate is back to requiring `assistant_messages`. See the reply in the parent thread for details.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/adot_models.py
tejaskash: Went with chronological. `ConversationTurn.input_messages` now interleaves user and assistant events in arrival order — for `[u1, a1, u2, a2, u3]` downstream sees `[user, assistant, user, assistant, user]`, matching the real conversation the model received. Test asserts that order.

src/bedrock_agentcore/evaluation/span_to_adot_serializer/strands_converter.py
tejaskash: Added `test_extract_conversation_turn_tool_results_only_returns_none` for your suggested case. Quick verification of the "previously returned `ConversationTurn`" claim: the old guard was `user_message and assistant_messages`, and tool_results were never in the guard — so a user+tool_result-only span with no assistant events returned `None` in the old code too. No regression. The gate was broadened to `user_messages and (assistant_messages or history_messages)` in the first pass of this PR, but per jariys comment it is now tightened back to require `assistant_messages`. Whether tool-only turns should produce a `ConversationTurn` is a separate feature decision for the Evaluations team; out of scope for this bugfix.

jariy17 (approved):

PR #455, opened 2026-05-05
test: add multi-turn conversation serialization integration test

tejaskash:
## Summary

Follow-up to #454. That PR fixed two bugs in `StrandsEventParser` / `ConversationTurn` / `ADOTDocumentBuilder`, and the unit tests covered them with MagicMock-based spans. This adds an **end-to-end integration test** that exercises the full pipeline — real Strands agent → real Bedrock call → real OTel spans → `convert_strands_to_adot` → ADOT document shape assertions.

## What the test does

Runs a 4-turn Strands conversation against Claude (default model, configurable region via `BEDROCK_TEST_REGION`), captures the in-memory spans, and locates the final `chat` span — the Strands span kind that carries the full conversation history as `gen_ai.user.message` / `gen_ai.assistant.message` / `gen_ai.choice` events. Asserts the ADOT log record derived from that span:

| Assertion | Rationale |
|---|---|
| Every prior user turn appears in `body.input.messages` | Bug 1 regression — the scalar-overwrite dedup would have left only the newest |
| Every prior assistant turn appears in `body.input.messages` (not `output.messages`) | Bug 2 regression — prior turns were being bucketed as current-turn output |
| Exactly one entry in `body.output.messages`, containing the current-turn `gen_ai.choice` | Output contamination check |
| Output does not contain any stale history text | Explicit contamination check |
| `input_messages` role order interleaves `[user, assistant, user, assistant, ..., user]` | Chronological ordering — the ordering issue raised during review of #454 |

## Why this test matters over the unit tests

Unit tests assert against a MagicMock event stream shaped the way I thought Strands emitted spans. This test **pins the actual shape** Strands produces in production. Specifically, it discovered during development that Strands emits three span kinds per turn (`chat`, `execute_event_loop_cycle`, `invoke_agent`) — only `chat` carries the full history events where the bug manifested. Generic "grab the last log record" checks would miss this.

If Strands upstream changes its instrumentation (renames events, splits span kinds, changes attribute shapes), this test catches it. If a future refactor of `span_to_adot_serializer` breaks serialization for real span objects in ways mocked spans don't reflect, this test catches it.

## Running it

```bash
AWS_PROFILE=<profile> BEDROCK_TEST_REGION=us-west-2 \
  uv run pytest tests_integ/evaluation/integrations/strands_agents_evals/test_multi_turn_conversation_serialization.py -xvs
```

Requires Bedrock model access (default model is whatever Strands picks, currently `us.anthropic.claude-sonnet-4-*`).

## Verification

Ran locally against Bedrock in `us-east-1`. Final-turn chat span produced:
- `input.messages` = 4 user entries + 3 assistant-history entries (all 4 prior user turns preserved, all 3 prior assistant turns land in input)
- `output.messages` = 1 entry (the current `gen_ai.choice`, content `"Your name is Taro and you live in Seattle."`)
- Role order = `[user, assistant, user, assistant, user, assistant, user]` (chronological)

Pre-fix, this test would fail at the very first assertion (4 expected user entries, 1 observed).

## Test plan

- [x] Runs green against real Bedrock (us-east-1, Claude Sonnet 4, 4-turn conversation)
- [x] `ruff check` / `ruff format` clean
- [x] Would fail pre-#454 fix — assertions target the exact shape the bug produced

### DIFF
--- a/tests_integ/evaluation/integrations/strands_agents_evals/test_multi_turn_conversation_serialization.py
+++ b/tests_integ/evaluation/integrations/strands_agents_evals/test_multi_turn_conversation_serialization.py
@@ -1,0 +1,178 @@
+"""Integration test: multi-turn Strands conversation history preservation.
+
+Ticket V2197498381 reported that the span serializer dropped all but the most
+recent user turn and mixed prior assistant turns into the current-turn output.
+This test runs a real multi-turn Strands agent, captures the emitted OTel
+spans, and asserts that `convert_strands_to_adot` serializes the full
+conversation in chronological order.
+
+Run with:
+    AWS_PROFILE=<profile> BEDROCK_TEST_REGION=us-west-2 \
+        uv run pytest tests_integ/evaluation/integrations/strands_agents_evals/\
+test_multi_turn_conversation_serialization.py -xvs --log-cli-level=INFO
+"""
+
+import logging
+import os
+
+import pytest
+from strands import Agent
+from strands_evals.telemetry import StrandsEvalsTelemetry
+
+from bedrock_agentcore.evaluation.span_to_adot_serializer import convert_strands_to_adot
+
+logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
+logger = logging.getLogger(__name__)
+
+REGION = os.environ.get("BEDROCK_TEST_REGION", "us-west-2")
+
+pytestmark = pytest.mark.filterwarnings("ignore::UserWarning:pydantic.main")
+
+
+@pytest.mark.integration
+class TestMultiTurnConversationSerialization:
+    """End-to-end validation that multi-turn history survives serialization."""
+
+    def _run_multi_turn_agent(self, turns: list[str]) -> list:
+        """Run an agent through several user turns and return the captured spans."""
+        telemetry = StrandsEvalsTelemetry().setup_in_memory_exporter()
+        agent = Agent(
+            system_prompt=(
+                "You are a helpful assistant. Keep responses short — one sentence. "
+                "Remember details the user shares across turns."
+            ),
+        )
+        for user_msg in turns:
+            response = agent(user_msg)
+            logger.info("user=%r -> assistant=%r", user_msg, str(response)[:80])
+
+        return list(telemetry.in_memory_exporter.get_finished_spans())
+
+    @staticmethod
+    def _flatten(msg: dict) -> str:
+        """Extract the text payload from a serialized message entry."""
+        content = msg.get("content", "")
+        if isinstance(content, dict):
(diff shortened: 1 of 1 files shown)


tests_integ/evaluation/integrations/strands_agents_evals/test_multi_turn_conversation_serialization.py
```diff
@@ -0,0 +1,178 @@
...
+        expected_pattern = []
+        for i in range(len(turns)):
+            expected_pattern.append("user")
+            if i < expected_history_count:
+                expected_pattern.append("assistant")
+        assert roles_in_order == expected_pattern, (
+            f"Expected chronological role interleaving {expected_pattern}; got {roles_in_order}. "
+            f"Pre-fix role-grouped ordering would have emitted "
+            f"[{'user, ' * len(turns)}{'assistant, ' * expected_history_count}] instead."
+        )
```
aidandaly24, line 178: **Potential contradiction with PR #454's builder ordering.**

This asserts chronological interleaving: `[user, assistant, user, assistant, ..., user]`.

However, `build_conversation_log_record` in PR #454 serializes as role-grouped:
```python
input_messages = [{...} for user_msg in conversation.user_messages]  # all users
input_messages.extend(conversation.history_messages)  # then all assistants
```

That would produce `[user, user, user, user, assistant, assistant, assistant]` — failing this assertion.

Can you confirm: does `convert_strands_to_adot` have separate ordering logic that produces interleaved output? Or was the builder updated to interleave after my review on #454? If neither, this test will fail in CI against the current `build_conversation_log_record` implementation.

aidandaly24 (commented):
Good integration test — high value for pinning real Strands behavior. One high-risk concern and one nit inline.

tests_integ/evaluation/integrations/strands_agents_evals/test_multi_turn_conversation_serialization.py
```diff
@@ -0,0 +1,178 @@
...
+        joined_output = " ".join(output_texts).lower()
+        assert joined_output, "output.messages was empty"
+        for stale in (
+            "nice to meet you, taro",
+            "got it, taro from seattle",
+            "that's interesting work",
+        ):
+            assert stale not in joined_output, (
+                f"output.messages contains stale history entry containing {stale!r}: {output_texts}"
+            )
```
aidandaly24, line 163: nit: These stale-content assertions are fragile — they assume specific model responses (`"nice to meet you, taro"`, etc.) that Claude may not produce. When the model responds differently, these pass vacuously without actually testing anything. The `len(output_messages) == 1` assertion above already structurally validates that history isn't leaking into output. Consider removing these or replacing with a structural check (e.g., verify no output text substring appears in any input entry).

PR #474, opened 2026-05-12
fix: allow bound methods in entrypoint() registration

### LINKED ISSUE #473: BedrockAgentCoreApp.entrypoint() raises AttributeError when registering a bound method
**Describe the bug**
  
  `BedrockAgentCoreApp.entrypoint(func)` raises `AttributeError` when `func` is a bound method. Python bound methods have no writable `__dict__`, so the `func.run = lambda ...` assignment on
  `src/bedrock_agentcore/runtime/app.py:223` fails. This makes the natural object-oriented pattern — registering `self.handle` from a class `__init__` — unusable.

  **To Reproduce**

  1. `pip install bedrock-agentcore==1.9.0`
  2. Save the following as `repro.py`:
     ```python
     from bedrock_agentcore import BedrockAgentCoreApp


     class MyAgent:
         def __init__(self):
             self.app = BedrockAgentCoreApp()
             self.app.entrypoint(self.handle)
  
         async def handle(self, payload):
             yield payload


     MyAgent()
     ```
  3. Run `python repro.py`
  4. See the traceback:
     ```    
     Traceback (most recent call last):
       File ".../repro.py", line 11, in <module>
         MyAgent()
       File ".../repro.py", line 6, in __init__
         self.app.entrypoint(self.handle)
       File ".../bedrock_agentcore/runtime/app.py", line 223, in entrypoint
         func.run = lambda port=8080, host=None: self.run(port, host)
     AttributeError: 'method' object has no attribute 'run' and no __dict__ for setting new attributes
     ```
  
  **Expected behavior**

  Registering a bound method should succeed. The request-dispatch path (`_handle_invocation` reading `self.handlers["main"]`) handles bound methods correctly — only the `.run` sugar assignment fails. Class-based agents are a common
  pattern (e.g. extending an abstract base and registering `self._entrypoint` from `__init__`), and there is no documented restriction against this.
  
  **Screenshots**

  N/A — text-only.
  
  - OS: macOS 15 (Darwin 25.4.0) — pure-Python issue, platform-independent
  - Python: 3.13.13   
  - `bedrock-agentcore` version: 1.9.0 (latest on PyPI; same code on `main` at `src/bedrock_agentcore/runtime/app.py:223`)

  **Smartphone (please complete the following information):**

  N/A.
  
  **Additional context**

  **Root cause** — `src/bedrock_agentcore/runtime/app.py:213-224`:

  ```python
  def entrypoint(self, func: Callable) -> Callable:
      self.handlers["main"] = func
      func.run = lambda port=8080, host=None: self.run(port, host)   # ← fails on bound methods
      return func
  ```

  The `func.run = ...` line attaches a convenience method so callers can write `handler.run()` instead of `app.run()`. It is **not** used by the request dispatcher (`_handle_invocation` at line 549 only reads
  `self.handlers.get("main")`). The line breaks bound methods to provide an alternative invocation form many users don't need.

  Side note: even when `func` is a plain function, `func.run = ...` mutates *shared state* on the function object — two agents in the same process would clobber each other's `.run` lambda.
  
  **Suggested fix** — guard the attribute assignment:
  
  ```python
  def entrypoint(self, func: Callable) -> Callable:
      self.handlers["main"] = func
      try:
          func.run = lambda port=8080, host=None: self.run(port, host)
      except AttributeError:
          pass  # bound methods / built-ins / objects without __dict__
      return func
  ```

  Or, more explicitly:
  
  ```python
  import inspect
  ...
  if inspect.isfunction(func):
      func.run = lambda port=8080, host=None: self.run(port, host)                                                                                                                                                       
  ```

  Either preserves the existing ergonomic behaviour for module-level functions without breaking class-based registration.
  
  **Workarounds (until fixed):**

  1. Bypass `entrypoint()` and write to `handlers` directly:
     ```python 
     self.app = BedrockAgentCoreApp()
     self.app.handlers["main"] = self.handle
     ```
  2. Wrap in a closure:
     ```python
     async def handler(payload):
         async for event in self.handle(payload):
             yield event
     self.app.entrypoint(handler)
     ```

jesseturner21:
## Summary

- Guard the `.run` attribute assignment in `entrypoint()` with `try/except AttributeError` so that bound methods (and other callables without a writable `__dict__`) can be registered without raising
- Adds a regression test for the class-based agent pattern

Fixes #473

## Test plan

- [x] `test_entrypoint_bound_method` passes — registers a bound method without error
- [x] `test_entrypoint_decorator` still passes — plain functions still get `.run` sugar

### DIFF
--- a/src/bedrock_agentcore/runtime/app.py
+++ b/src/bedrock_agentcore/runtime/app.py
@@ -220,7 +220,10 @@ class BedrockAgentCoreApp(Starlette):
             The decorated function with added serve method
         """
         self.handlers["main"] = func
-        func.run = lambda port=8080, host=None: self.run(port, host)
+        try:
+            func.run = lambda port=8080, host=None: self.run(port, host)
+        except AttributeError:
+            pass
         return func
 
     def ping(self, func: Callable) -> Callable:
--- a/tests/bedrock_agentcore/runtime/test_app.py
+++ b/tests/bedrock_agentcore/runtime/test_app.py
@@ -61,6 +61,21 @@ class TestBedrockAgentCoreApp:
         assert hasattr(test_handler, "run")
         assert callable(test_handler.run)
 
+    def test_entrypoint_bound_method(self):
+        """Test entrypoint accepts a bound method without raising AttributeError."""
+
+        class MyAgent:
+            def __init__(self):
+                self.app = BedrockAgentCoreApp()
+                self.app.entrypoint(self.handle)
+
+            def handle(self, payload):
+                return {"result": "success"}
+
+        agent = MyAgent()
+        assert "main" in agent.app.handlers
+        assert agent.app.handlers["main"] == agent.handle
+
     def test_invocation_without_context(self):
         """Test handler without context parameter works correctly."""
         bedrock_agentcore = BedrockAgentCoreApp()


tejaskash (approved):
LGTM

Issue #452, opened 2026-05-04
[Enhancement] AgentCoreMemorySessionManager: add async_mode setting to prevent event loop blocking


jariy17:
## Problem

`AgentCoreMemorySessionManager` makes synchronous boto3 calls on the asyncio event loop hot path when used with `Agent.stream_async()` inside an async WebSocket server (e.g. AgentCore Runtime). This adds **200–800ms of blocked event loop time per agent turn**, directly degrading time-to-first-token (TTFT) for streaming responses.

The four blocking call sites are:
1. `initialize()` → `read_session()`, `read_agent()`, `list_messages()` (sync gmdp client calls)
2. `append_message()` → `create_event()` (sync, once per message when `batch_size=1`)
3. `retrieve_customer_context()` → uses `ThreadPoolExecutor` + `as_completed()`, but `as_completed()` still blocks the **calling** coroutine
4. `sync_agent()` → `update_agent()` → `create_event()` (sync, per turn)

Because all four run inside the Strands hook lifecycle (`BeforeInvocationEvent`, `MessageAddedEvent`, `AfterInvocationEvent`), callers cannot wrap them without subclassing or forking the SDK.

## Proposed Solution

Add an `async_mode` configuration setting to `AgentCoreMemorySessionManager` (or its config class):

```python
class MemorySessionManagerConfig:
    async_mode: bool = False  # default: sync (backwards-compatible)
```

When `async_mode=True`, the session manager wraps all 4 blocking call sites with `asyncio.to_thread()` to offload boto3 calls to a thread pool, keeping the event loop unblocked:

```python
# Pseudocode
if self.config.async_mode:
    await asyncio.to_thread(self._blocking_call, ...)
else:
    self._blocking_call(...)
```

This is a **non-breaking change** — existing sync users default to `sync` behavior unchanged. Async users opt in by setting `async_mode=True`.

## Impact

- **Without fix:** 200–800ms per turn of event loop blocking → degraded TTFT for streaming agents
- **With fix:** Boto3 calls offloaded to thread pool → event loop stays free for I/O

## Affected File

`src/bedrock_agentcore/memory/integrations/strands/session_manager.py`

## Acceptance Criteria

- [ ] `async_mode: bool = False` added to config (backwards-compatible default)
- [ ] All 4 blocking call sites wrapped with `asyncio.to_thread` when `async_mode=True`
[...]

PR #478, opened 2026-05-12
feat: add async support to MemorySessionManager

nborges-aws:
*Issue #, if available:* 
#452 

*Description of changes:*
  - New `async_mode: bool = False` on AgentCoreMemoryConfig. Opt-in; default preserves existing behavior.
  - When True, `register_hooks` installs async callbacks that wrap the existing sync methods in `asyncio.to_thread()`
  - logger.warning at `register_hooks` time when `async_mode=True` pointing users to `stream_async` / `invoke_async` — sync invocation will raise RuntimeError from Strands.
  - 7 new tests covering newly added logic

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -1,5 +1,6 @@
 """AgentCore Memory-based session manager for Bedrock AgentCore Memory integration."""
 
+import asyncio
 import json
 import logging
 import threading
@@ -10,7 +11,13 @@ from typing import TYPE_CHECKING, Any, Dict, NamedTuple, Optional
 
 import boto3
 from botocore.config import Config as BotocoreConfig
+from strands.experimental.hooks.multiagent.events import (
+    AfterMultiAgentInvocationEvent,
+    AfterNodeCallEvent,
+    MultiAgentInitializedEvent,
+)
 from strands.hooks import AfterInvocationEvent, MessageAddedEvent
+from strands.hooks.events import AgentInitializedEvent
 from strands.hooks.registry import HookRegistry
 from strands.session.repository_session_manager import RepositorySessionManager
 from strands.session.session_repository import SessionRepository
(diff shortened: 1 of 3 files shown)


src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -906,16 +913,79 @@ def retrieve_for_namespace(namespace: str, retrieval_config: RetrievalConfig):
...
+        # methods to a worker thread via asyncio.to_thread. AgentInitializedEvent
+        # must stay sync (Strands disallows async callbacks on this event; see
+        # strands/hooks/registry.py:174).
+        logger.warning(
+            "AgentCoreMemorySessionManager async_mode=True: the agent must be invoked "
+            "via the async path (e.g. agent.stream_async(...) or agent.invoke_async(...)). "
+            "Sync invocation will raise RuntimeError from Strands' hook registry."
+        )
+
+        registry.add_callback(AgentInitializedEvent, lambda event: self.initialize(event.agent))
```
Hweinstock, line 970: note (for my own understanding): because this path doesn't call `RepositorySessionManager.register_hooks`, we must manually register the initialize hook: https://github.com/strands-agents/sdk-python/blob/main/src/strands/session/session_manager.py#L43. 

In other words, we pick this synchronous hook from that implementation and leave out the rest to overwrite them with our own async hooks.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -906,16 +913,79 @@ def retrieve_for_namespace(namespace: str, retrieval_config: RetrievalConfig):
...
+
+        registry.add_callback(MessageAddedEvent, _on_message_added_persist)
+        registry.add_callback(AfterInvocationEvent, _on_after_invocation_sync)
+        registry.add_callback(MessageAddedEvent, _on_message_added_retrieve)
 
-        # Only register AfterInvocationEvent hook when batching is enabled
         if self.config.batch_size > 1:
-            registry.add_callback(AfterInvocationEvent, lambda event: self._flush_messages())
+
+            async def _on_after_invocation_flush(event: AfterInvocationEvent) -> None:
```
Hweinstock: nit: could we add a small helper to reduce boilerplate here?

Hweinstock (dismissed):
changes LGTM! I think an integ test for this flag would be awesome (perhaps as a follow-up), and perhaps worth getting @jariy17 to take a look.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -906,16 +913,79 @@ def retrieve_for_namespace(namespace: str, retrieval_config: RetrievalConfig):
...
+        docstring for mitigations.
+
         Args:
             registry (HookRegistry): The hook registry to register callbacks with.
             **kwargs: Additional keyword arguments.
         """
-        RepositorySessionManager.register_hooks(self, registry, **kwargs)
-        registry.add_callback(MessageAddedEvent, lambda event: self.retrieve_customer_context(event))
+        if not self.config.async_mode:
+            RepositorySessionManager.register_hooks(self, registry, **kwargs)
```
jariy17, line 940: note: We're manually registering hooks in the async path instead of going through RepositorySessionManager.register_hooks(), so if strands adds new hooks upstream, we won't pick them up here.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -906,16 +913,79 @@ def retrieve_for_namespace(namespace: str, retrieval_config: RetrievalConfig):
...
+        blocks the calling thread in async mode — see AgentCoreMemoryConfig
+        docstring for mitigations.
+
         Args:
             registry (HookRegistry): The hook registry to register callbacks with.
             **kwargs: Additional keyword arguments.
         """
-        RepositorySessionManager.register_hooks(self, registry, **kwargs)
-        registry.add_callback(MessageAddedEvent, lambda event: self.retrieve_customer_context(event))
+        if not self.config.async_mode:
```
jariy17, line 939: Nit: I prefer if self.config.async_mode: — easier to read when the primary condition isn't negated.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -906,16 +913,79 @@ def retrieve_for_namespace(namespace: str, retrieval_config: RetrievalConfig):
...
+
+        async def _on_after_node_call(event: AfterNodeCallEvent) -> None:
+            await asyncio.to_thread(self.sync_multi_agent, event.source)
+
+        async def _on_after_multi_agent_invocation(event: AfterMultiAgentInvocationEvent) -> None:
+            await asyncio.to_thread(self.sync_multi_agent, event.source)
+
+        registry.add_callback(MultiAgentInitializedEvent, _on_multi_agent_initialized)
+        registry.add_callback(AfterNodeCallEvent, _on_after_node_call)
+        registry.add_callback(AfterMultiAgentInvocationEvent, _on_after_multi_agent_invocation)
```
jariy17: nit: Do we support BidiAgent? The sync path picks up BidiAgent hooks through the parent, but async mode doesn't register them. Should we add them here or explicitly document that BidiAgent + async_mode is unsupported?

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
jariy17, line 1137: This function copies the buffer under one lock acquisition (L1142-1143), does network I/O, then clears it under a separate lock acquisition (L1175-1176). Anything appended
between the copy and the clear gets destroyed without ever being sent. 

With async mode, multiple to_thread workers can be calling create_agent concurrently, which makes the window between copy and clear much easier to hit. Let's copy the same locking mechanism in  `_flush_messages_only()` where it sends and clears under the same lock.

jariy17 (changes requested):
Everything is good but look into the _flush_agent_state bug and missing bidi hooks.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
Hweinstock: great find, agree we should copy and clear within the same lock acquire.

jariy17 (approved):

PR #479, opened 2026-05-15
docs: document time_of_last_update in /ping response

### LINKED ISSUE #471: docs: /ping response requires undocumented `time_of_last_update` field — without it AgentCore silently reaps microVMs even when status is HealthyBusy
## Problem

  The `/ping` contract documented at
  [runtime-long-run.html](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-long-run.html)
  and [runtime-troubleshooting.html](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-troubleshooting.html)
  shows only:

      {"status": "HealthyBusy"}

  In practice, the platform's idle reaper requires a second field —
  `time_of_last_update` — that the AWS-facing docs never mention.
  Without it, `idleRuntimeSessionTimeout` fires at the configured
  boundary even while `/ping` returns `HealthyBusy`, silently terminating
  microVMs mid-execution.

  The official SDK has emitted this field since the first public commit
  (`4f5c80d`, 2025-07-15) — see
  [`runtime/app.py:612`](https://github.com/aws/bedrock-agentcore-sdk-python/blob/main/src/bedrock_agentcore/runtime/app.py#L612).
  A third-party post documents the full schema correctly:
  https://eashank16.medium.com/demystifying-the-http-protocol-contract-for-amazon-bedrock-agentcore-runtime-30ae130485b4

  ## Reproduction (controlled, ~10 min)

  1. Build a minimal ARM64 container with two endpoints:
     - `/invocations`: spawn a background asyncio task, return `{"status":"accepted"}` immediately
  2. `CreateAgentRuntime` with `lifecycleConfiguration: {idleRuntimeSessionTimeout: 60, maxLifetime: 28800}`
  3. Invoke once. Observe CloudWatch:
     - Pings continue every ~2s returning HealthyBusy
     - Container is reaped at exactly +60s

  Repeat with the only change being the ping body:

      {"status": "HealthyBusy", "time_of_last_update": <int unix>}

  Container now survives indefinitely (verified to 256 s in our test).

  ## Other affected customers

  - re:Post (May 2025): https://repost.aws/questions/QUXrgkw_c_QmC2m9fh-6LIzg/aws-bedrock-agentcore-issues — describes identical silent-restart symptom; AWS-bot answer recommends `HEALTHY_BUSY` pattern but doesn't
  mention the field
  - `aws-samples/sample-host-openclaw-on-amazon-bedrock-agentcore` PR #36 — merged a fix that returns `{"status": "HealthyBusy"}` without the field, almost certainly still broken under load
  - Internal: we hit this on N=4 production agent runs across two AWS accounts before isolating it via the test above

  ## Asks

  1. Update [runtime-long-run.html](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-long-run.html) and
     [runtime-troubleshooting.html](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-troubleshooting.html)
     to document the full ping schema (both fields, with semantics).
  2. Either make the field optional on the platform side (so the
     currently-documented contract works) or surface a clear error/warning
     when a runtime returns a `/ping` body missing the field.
  3. Patch the broken reference implementation in
     `aws-samples/sample-host-openclaw-on-amazon-bedrock-agentcore` PR #36.

jesseturner21:
## Summary

- Documents the previously undocumented `time_of_last_update` field in the `/ping` endpoint response
- Adds the full response schema (`status` + `time_of_last_update`) to the API reference section
- Adds a note in the Custom Ping Handler section explaining the timestamp semantics

## Context

The `/ping` endpoint has always returned `{"status": "...", "time_of_last_update": <unix_timestamp>}` across all runtime protocols (main app, A2A, AG-UI), but the `time_of_last_update` field was never documented. This caused confusion for developers integrating with the health check endpoint (#471).

The code is correct — this is purely a documentation fix.

## Test plan

- [ ] Verify markdown renders correctly
- [ ] No code changes, so no runtime testing needed

Fixes #471

### DIFF
--- a/docs/examples/a2a_protocol_examples.md
+++ b/docs/examples/a2a_protocol_examples.md
@@ -142,7 +142,12 @@ Builds a Starlette ASGI application without starting a server. Useful for testin
 Returns a `Starlette` application with routes:
 - `POST /` — A2A JSON-RPC endpoint (`message/send`, `message/stream`, `tasks/get`, `tasks/cancel`)
 - `GET /.well-known/agent-card.json` — Agent card discovery
-- `GET /ping` — Bedrock health check
+- `GET /ping` — Bedrock health check. Returns JSON:
+  ```json
+  {"status": "Healthy" | "HealthyBusy", "time_of_last_update": 1715000000}
+  ```
+  - `status` — current health status from the `ping_handler` (or `"Healthy"` by default)
+  - `time_of_last_update` — Unix timestamp (seconds) indicating when the status was last updated
 
 ### `build_runtime_url(agent_arn, region=None)`
 
@@ -192,3 +197,5 @@ serve_a2a(executor, ping_handler=my_ping)
 ```
 
 If the ping handler raises an exception, the server falls back to `PingStatus.HEALTHY`.
+
+The `/ping` response always includes a `time_of_last_update` field (Unix timestamp in seconds) reflecting when the status last changed. Clients can use this to detect stale health state.


jesseturner21:
Closing — the real fix belongs in the public AWS docs (AWSGenesisDocs), not the SDK docs. The SDK already emits the field correctly.

PR #481, opened 2026-05-16
add metadata support for LTM

nborges-aws:
*Description of changes:*
Adds support for AgentCore Memory's long-term metadata feature on the data plane and control plane.

Indexed keys (control plane):
  - `MemoryClient.create_memory()`, `create_memory_and_wait()`, and `create_or_get_memory()` accept an optional `indexed_keys: List[IndexedKey]` parameter
  - New `IndexedKey` TypedDict and `MetadataValueType` enum in `bedrock_agentcore.memory.models. IndexedKey.build(key, value_type)`
  - `update_memory(addIndexedKeys=...)` is supported via the existing passthrough on MemoryClient
  
Metadata filters (data plane):
  - `MemoryClient.retrieve_memories()` and `MemorySessionManager.search_long_term_memories()` accept optional `metadata_filters: List[MemoryMetadataFilter]`
  - Builder method `MemoryMetadataFilter.build_expression(left, operator, right=None)` constructs a filter dict for boto3
  - `MemoryRecordRightExpression` exposes `build_string`, `build_number`, `build_datetime`, and `build_string_list` factories that match the four service-side value types

### DIFF
--- a/src/bedrock_agentcore/memory/models/filters.py
+++ b/src/bedrock_agentcore/memory/models/filters.py
@@ -1,7 +1,14 @@
-"""Event metadata filter models for querying events based on metadata."""
+"""Metadata filter models for querying events and memory records."""
 
+from datetime import datetime
 from enum import Enum
-from typing import Optional, TypedDict, Union
+from typing import List, Optional, TypedDict, Union
+
+from typing_extensions import NotRequired
+
+# ============================================================================
+# Event Metadata Filters (existing)
+# ============================================================================
 
 
 class StringValue(TypedDict):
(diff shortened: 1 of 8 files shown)


tests/bedrock_agentcore/memory/test_client.py
```diff
@@ -3498,3 +3504,242 @@ def test_get_last_k_turns_explicit_max_results():
         # Total events fetched should not exceed max_results
         total_fetched = sum(1 for _ in mock_gmdp.list_events.call_args_list)
         assert total_fetched <= 50  # Should stop after fetching 50 events worth of calls
+
+
+# ============================================================================
+# LTM Metadata: indexed_keys and metadata_filters tests
```
Hweinstock, line 3510: these tests are pretty heavily mocked, I think it might be worth adding some integ tests to verify the happy path for creating with memory keys at least. ( we don't need to wait for extraction ).

src/bedrock_agentcore/memory/models/filters.py
```diff
@@ -124,3 +131,213 @@ def build_expression(
...
+    | `BEFORE`                 | datetime                    | `build_datetime`         |
+    | `AFTER`                  | datetime                    | `build_datetime`         |
+    | `CONTAINS`               | string list                 | `build_string_list`      |
+    | `GREATER_THAN`           | number                      | `build_number`           |
+    | `GREATER_THAN_OR_EQUALS` | number                      | `build_number`           |
+    | `LESS_THAN`              | number                      | `build_number`           |
+    | `LESS_THAN_OR_EQUALS`    | number                      | `build_number`           |
+    """
+
+    EQUALS_TO = "EQUALS_TO"
```
Hweinstock, line 162: q: these feel like generic comparisons? Does it make sense to decouple from memory or do we not see a use for them outside of it?

Hweinstock (approved):

Issue #468, opened 2026-05-11
BedrockAgentCoreApp SDK needs update to support expanded Request Header Allowlist


naviret:
Hi team,

We recently expanded the AgentCore Runtime **Request Header Allowlist** feature ([docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-header-allowlist.html)). Previously, the allowlist only supported `Authorization` and headers prefixed with `X-Amzn-Bedrock-AgentCore-Runtime-Custom-`. The expansion loosens the validation performed in CP to allow customers to configure any valid HTTP header in their allowlist (e.g., `X-Custom-Header`, `X-Api-Key`) while blocking certain restricted headers. 
The Runtime service (data plane) correctly forwards all allowlisted headers to the container — this has been validated end-to-end in prod PDX.

**Where this breaks in `BedrockAgentCoreApp`:**

The `_build_request_context` method in [app.py#L395-L430](https://github.com/aws/bedrock-agentcore-sdk-python/blob/main/src/bedrock_agentcore/runtime/app.py#L395-L430) still has the old hardcoded filter that only populates `context.request_headers` with:
1. The `Authorization` header
2. Headers matching the `X-Amzn-Bedrock-AgentCore-Runtime-Custom-` prefix

Any newly-supported allowlisted header (e.g., `X-Custom-Header`, `X-Api-Key`) that arrives at the container is silently dropped by the SDK before reaching the customer's `@app.entrypoint` handler via `context.request_headers`.

**Ask:**

Since `BedrockAgentCoreApp` is the recommended SDK for building agents on AgentCore, and the [docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-header-allowlist.html) show `context.request_headers` as the access pattern, we'd like the SDK to reflect the expanded feature's behavior. The Runtime service has already enforced the allowlist at the CP and forwarded only valid headers to the container — so the SDK should surface all headers that arrive rather than applying its own secondary filter based on the old prefix rules.

We'd really appreciate your help if you guys could take a look. Thanks so much!

Ivan :)

notgitika:
This issue was fixed in latest SDK release 1.10.0. Thank you!

PR #483, opened 2026-05-18
feat: expand custom request header forwarding to match runtime allowlist

padmak30:
_build_request_context in BedrockAgentCoreApp, BedrockCallContextBuilder (A2A), and AGUIApp previously only forwarded Authorization and headers prefixed with X-Amzn-Bedrock-AgentCore-Runtime-Custom-, silently dropping any other custom headers (e.g. X-Api-Key, X-Custom-Signature) even after the runtime service was updated to forward them to the container.

Adds RESTRICTED_HEADERS (frozenset) and is_forwardable_header() to models.py implementing the full allowlist rules from the docs: block the restricted header set, x-amz-* (SigV4 reserved), and x-amzn-* (except the legacy AgentCore custom prefix); pass everything else through.

*Issue #, if available:* https://github.com/aws/bedrock-agentcore-sdk-python/issues/468

### DIFF
--- a/src/bedrock_agentcore/runtime/models.py
+++ b/src/bedrock_agentcore/runtime/models.py
@@ -22,6 +22,140 @@ AUTHORIZATION_HEADER = "Authorization"
 CUSTOM_HEADER_PREFIX = "X-Amzn-Bedrock-AgentCore-Runtime-Custom-"
 AGENTCORE_RUNTIME_URL_ENV = "AGENTCORE_RUNTIME_URL"
 
+# Headers that cannot be forwarded to agent code.
+# Source: https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-header-allowlist.html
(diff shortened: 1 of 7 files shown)


src/bedrock_agentcore/runtime/models.py
```diff
@@ -22,6 +22,140 @@ class PingStatus(str, Enum):
 CUSTOM_HEADER_PREFIX = "X-Amzn-Bedrock-AgentCore-Runtime-Custom-"
 AGENTCORE_RUNTIME_URL_ENV = "AGENTCORE_RUNTIME_URL"
 
+# Headers that cannot be forwarded to agent code.
+# Source: https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-header-allowlist.html
+RESTRICTED_HEADERS: frozenset[str] = frozenset(
```
notgitika, line 27: what if this gets out of sync?

notgitika (approved):

PR #491, opened 2026-05-21
feat(evaluation): add DatasetClient and dataset management service provider

jariy17:
## Summary
- Adds `DatasetClient`, a high-level wrapper for Bedrock AgentCore dataset management operations (create/get/list/delete datasets and versions, upload/download via presigned URLs, polling helpers).
- Adds `DatasetManagementServiceProvider` (formerly `ServiceDatasetProvider`) that loads evaluation scenarios from a managed dataset, alongside the existing `FileDatasetProvider` which now also supports JSONL files.
- Refactors `Scenario` classes so each owns its `schema_type`, and moves `_parse_scenario` to module level for reuse between file- and service-backed providers.
- Streams JSONL downloads, extracts region from agent ARN (no separate `BEDROCK_TEST_REGION`), and consolidates test fixtures.
- Unit and integration tests for the new client and provider, plus a runner integ test that exercises a real agent.

## Test plan
- [ ] `uv run pytest tests/bedrock_agentcore/evaluation` passes
- [ ] `uv run pytest tests_integ/evaluation` passes against a configured AWS account
- [ ] Lint/format checks pass (`ruff`, line-length)
- [ ] Manual: create dataset → upload JSONL → run evaluation via `DatasetManagementServiceProvider` end-to-end


src/bedrock_agentcore/evaluation/runner/dataset_providers.py
```diff
@@ -16,49 +59,86 @@ def get_dataset(self) -> Dataset:
...
-        )
+        download_url = response.get("downloadUrl")
+        if not download_url:
+            raise ValueError(f"Dataset {self._dataset_id} has no downloadUrl. Status: {response.get('status')}")
+
+        try:
+            r = requests.get(download_url, timeout=60, stream=True)
+            r.raise_for_status()
+        except requests.RequestException as e:
+            raise RuntimeError(f"Couldn't download dataset from S3 bucket: {e}") from e
```
Hweinstock, line 136: is there anything here that requires the dataset is in an s3 bucket? AFAICT from the code it does a generic fetch. Or is that the only possible value of download_url from the service `get_dataset` call?

Hweinstock (approved):
LGTM!

src/bedrock_agentcore/evaluation/runner/dataset_providers.py
jariy17: No, because the service will store the dataset in a s3 bucket for them and they share a s3 bucket url.

PR #507, opened 2026-06-06
fix: lazy import requests in dataset_providers to fix ModuleNotFoundE…

saeekailas:
Problem:
importing anything from bedrock_agentcore.evaluation on a fresh install
failed with ModuleNotFoundError: No module named 'requests' because
requests was imported at the top of dataset_providers.py but was not
declared as a runtime dependency in pyproject.toml.

Fix:
Moved `import requests` from the module-level into the body of
DatasetManagementServiceProvider.get_dataset(), so it is only imported
when actually needed. Users who never download datasets from a URL are
not affected and no longer require requests to be installed.

Files changed:
src/bedrock_agentcore/evaluation/runner/dataset_providers.py

### DIFF
--- a/src/bedrock_agentcore/evaluation/runner/dataset_providers.py
+++ b/src/bedrock_agentcore/evaluation/runner/dataset_providers.py
@@ -4,7 +4,7 @@ import json
 from abc import ABC, abstractmethod
 from typing import Any, Dict, List, Optional
 
-import requests
+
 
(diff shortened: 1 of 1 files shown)


jariy17:
Hey @saeekailas We have solved this issue with this [PR](https://github.com/aws/bedrock-agentcore-sdk-python/pull/508). Closing this issue now

saeekailas:
Hey @jariy17, thanks for the update!

PR #508, opened 2026-06-08
fix: declare requests as optional 'datasets' extra

jariy17:
## Problem

On a fresh install, importing from `bedrock_agentcore.evaluation` and downloading a dataset from the Dataset Management service fails with `ModuleNotFoundError: No module named 'requests'`. `requests` was imported at module top level in `dataset_providers.py` but was never declared as a dependency.

## Fix

Declare `requests` as a new optional `datasets` extra, matching the repo's existing pattern for optional features (`strands-agents-evals`, `simulation`, `a2a`, `ag-ui`):

- **`pyproject.toml`** — add `datasets = ["requests>=2.31.0"]` under `[project.optional-dependencies]`, keeping `requests` out of the base install for users who never download datasets.
- **`dataset_providers.py`** — import `requests` lazily inside `get_dataset()`, raising a clear `ImportError` that points users at `pip install 'bedrock-agentcore[datasets]'` when the extra is not installed.
- **tests** — patch `requests.get` (the lazy import target) and add a test covering the missing-dependency path.

## Why this over a bare lazy import

A lazy import alone never declares the dependency, so a fresh install still fails — just later, at call time. This approach declares the dependency, fails with actionable install guidance when it's absent, and follows the convention already used elsewhere in the SDK.

## Testing

`test_service_dataset_provider.py` passes (9 tests, including the new missing-dependency case); `ruff check` clean.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -173,3 +173,6 @@ simulation = [
     "jinja2>=3.1.0",
     "strands-agents-evals>=0.1.0",
 ]
+datasets = [
+    "requests>=2.31.0",
+]
--- a/src/bedrock_agentcore/evaluation/runner/dataset_providers.py
+++ b/src/bedrock_agentcore/evaluation/runner/dataset_providers.py
@@ -4,8 +4,6 @@ import json
 from abc import ABC, abstractmethod
 from typing import Any, Dict, List, Optional
 
-import requests
-
 from bedrock_agentcore.evaluation.dataset_client import DatasetClient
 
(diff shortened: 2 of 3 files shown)


Hweinstock:
Same error as CLI tests `botocore.exceptions.ClientError: An error occurred (ForbiddenException) when calling the CreateDataset operation: Forbidden`, I wonder if this is a service issue.

Hweinstock (approved):

Issue #509, opened 2026-06-08
Migrate A2A runtime integration to a2a-sdk v1.x


jariy17:
## Summary

Migrate the A2A runtime integration (`src/bedrock_agentcore/runtime/a2a.py`) to support **a2a-sdk v1.x**. The SDK currently only works with a2a-sdk `0.3.x`; #438 reported that v1.0+ breaks at import time, and PR (cap fix) pins `a2a-sdk[http-server]>=0.3,<1.0` as a stopgap. This issue tracks lifting that cap by porting to the new v1.x server API.

## Background

a2a-sdk v1.0.0 (March 2026) redesigned the server API. Verified against a2a-sdk `1.1.0`:

| What `a2a.py` uses today (0.3.x) | Status in v1.x |
|---|---|
| `a2a.server.apps` (module) | ❌ removed — replaced by `a2a.server.routes` |
| `a2a.server.apps.A2AStarletteApplication` | ❌ removed — app model replaced (`create_jsonrpc_routes`, `add_a2a_routes_to_fastapi`) |
| `a2a.server.apps.CallContextBuilder` | ❌ removed — replaced by `a2a.server.routes.ServerCallContextBuilder` / `DefaultServerCallContextBuilder` |
| `a2a.types.DataPart` | ❌ moved — now under `a2a.compat.v0_3.types` |
| `a2a.server.request_handlers.DefaultRequestHandler` | ✅ still present |
| `a2a.server.tasks.InMemoryTaskStore` | ✅ still present |
| `a2a.server.context.ServerCallContext` | ✅ still present |
| `a2a.types.AgentCard` / `AgentCapabilities` / `AgentSkill` | ✅ still present |

## Scope of work

- Rewrite `build_a2a_app()` to construct the app via the v1.x `a2a.server.routes` API instead of `A2AStarletteApplication.add_routes_to_app()`.
- Replace `BedrockCallContextBuilder`'s base (`CallContextBuilder`) with the v1.x `ServerCallContextBuilder` interface, preserving the Bedrock header/baggage extraction behavior.
- Audit all `a2a.*` imports in `a2a.py` against v1.x and update moved symbols.
- Decide on the support window: either move the floor to `>=1.0` (drop 0.3.x), or support both 0.3.x and 1.x behind a version check. Recommendation to be settled in the PR.
- Update the `a2a` extra pin in `pyproject.toml` accordingly and refresh `uv.lock`.
[...]

PR #510, opened 2026-06-08
fix(a2a): cap a2a-sdk below 1.0 to restore A2A server startup

### LINKED ISSUE #438: Bedrock AgentCore >=1.6.3 + A2A v1.0 is throwing error  No module named 'a2a.server.apps'
**Describe the bug**
The A2A protocol hit v1.0.0 recently (March 2026) and introduced breaking changes into the API  but it seems bedrock-agentcore still leverages  a2a-sdk 0.3.0 API calls.

When starting the a2a server  leveraging  the bedrock_agentcore.runtime.a2a  serve_a2a function got this error  - "An error occurred during server startup: No module named 'a2a.server.apps' "  This is due to API incompatibility  with new a2a-sdk v 1.0.0

**To Reproduce**
Create a sample A2A app leveraging bedrock-agentcore python sdk > 1.6.3  and a2a-sdk > 1.0.0 you would get the error descrobed above

**Expected behavior**
A2A server should start gradefully without any issue on port 9000

**Screenshots**

 - OS: macOS
 - Browser : Safari
 - Version : latest macOS

jariy17:
## Problem

`pip install "bedrock-agentcore[a2a]"` fails to start the A2A server with `No module named 'a2a.server.apps'` (and a related `cannot import name 'DataPart' from 'a2a.types'`). The `a2a` extra and dev dependency pinned `a2a-sdk[http-server]>=0.3` with no upper bound, so fresh installs now resolve to a2a-sdk **1.x**, whose server API was redesigned in v1.0.0 and is incompatible with `runtime/a2a.py` (which targets 0.3.x).

Verified against a2a-sdk 1.1.0:
- `a2a.server.apps` module removed
- `A2AStarletteApplication` and `CallContextBuilder` removed
- `a2a.types.DataPart` moved to `a2a.compat.v0_3.types`

## Fix

Cap the dependency to the supported API in both the `a2a` extra and the dev dependency:

```
a2a-sdk[http-server]>=0.3,<1.0
```

This restores a working install immediately, with no code change. Migrating `runtime/a2a.py` to the a2a-sdk v1.x API (which would lift this cap) is a larger effort tracked in #509.

## Testing

Confirmed in a clean venv that a2a-sdk 1.1.0 raises `No module named 'a2a.server.apps'` on the import path used by `serve_a2a`, and that `<1.0` resolves to the 0.3.x line the integration supports.

Fixes #438.
Follow-up migration: #509.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -154,13 +154,13 @@ dev = [
     "wheel>=0.45.1",
     "strands-agents>=1.20.0",
     "strands-agents-evals>=0.1.0",
-    "a2a-sdk[http-server]>=0.3",
+    "a2a-sdk[http-server]>=0.3,<1.0",
     "ag-ui-protocol>=0.1.10",
     "mcp-proxy-for-aws>=0.1.0",
 ]
 
 [project.optional-dependencies]
-a2a = ["a2a-sdk[http-server]>=0.3"]
+a2a = ["a2a-sdk[http-server]>=0.3,<1.0"]
 ag-ui = ["ag-ui-protocol>=0.1.10"]
 strands-agents = [
     "strands-agents>=1.20.0",
--- a/uv.lock
+++ b/uv.lock
@@ -429,7 +429,7 @@ dev = [
 
 [package.metadata]
 requires-dist = [
-    { name = "a2a-sdk", extras = ["http-server"], marker = "extra == 'a2a'", specifier = ">=0.3" },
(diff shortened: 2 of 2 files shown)


Hweinstock:
Are these tests known flaky: ```tests_integ/memory/integrations/test_session_manager.py:205: AssertionError```?

Hweinstock (approved):

PR #513, opened 2026-06-11
fix(payments): correct "extension" typo to "extensions" in x402 v2 he…

rajuans:
…ader

The payment signature field name should be "extensions" (plural) to match the x402 protocol specification.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/bedrock_agentcore/payments/manager.py
+++ b/src/bedrock_agentcore/payments/manager.py
@@ -1452,7 +1452,7 @@ class PaymentManager:
                     "x402Version": 2,
                     "resource": x402_payload.get("resource"),
                     "accepted": selected_accept,
-                    "extension": x402_payload.get("extension", {}),
+                    "extensions": x402_payload.get("extensions", {}),
                     "payload": crypto_x402_proof.get("payload"),
                 }
                 header_json = json.dumps(payment_signature)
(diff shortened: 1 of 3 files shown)


aidandaly24:
### Code review

Found 1 issue:

1. The PR renames the x402 v2 header field `extension` → `extensions` in `manager.py` and updates the unit test, but the parallel integration test at `tests_integ/payments/test_payment_manager.py:542` still asserts the old singular name. It calls the same `generate_payment_header()` and parses the same `PAYMENT-SIGNATURE` JSON, so this assertion is now stale and will fail when the integration suite runs against the renamed field. Suggest updating it to `assert "extensions" in header_json` in the same PR for consistency.

https://github.com/aws/bedrock-agentcore-sdk-python/blob/b9979fa88d25538b00abafe484752cfdacc182cc/tests_integ/payments/test_payment_manager.py#L540-L542

aidandaly24 (approved):
Thanks for making the update

peterjiang-dev (approved):

Hweinstock:
Breaking change failure unreleated: `RequestError [HttpError]: Resource not accessible by integration`. 

Tools and policy hit known 403 forbidden issue https://github.com/aws/agentcore-cli/issues/1494. 

Memory hit a known flaky test: https://github.com/aws/bedrock-agentcore-sdk-python/issues/514

Will verify tests re-run on main before releasing this fix.

PR #522, opened 2026-06-15
fix: use ConfigDict instead of class-based Config in RequestContext

### LINKED ISSUE #320: Pydantic deprecation warning: Support for class-based `config` is deprecated, use ConfigDict instead
On import of `BedrockAgentCoreApp` a Pydantic deprecation warning occurs, e.g. during `pytest` runs. Besides noise, it is confusing since it is not clear if the issue is in application or Bedrock SDK code.

**To Reproduce**

1. open a python REPL
2. import `BedrockAgentCoreApp`

  ```
  ✗ python                                   
  Python 3.13.11 (main, Dec  5 2025, 16:06:33) [Clang 17.0.0 (clang-1700.6.3.2)] on darwin
  Type "help", "copyright", "credits" or "license" for more information.
  Cmd click to launch VS Code Native REPL
  >>> import warnings
  >>> warnings.simplefilter("always")
  >>> from bedrock_agentcore import BedrockAgentCoreApp
  [REDACTED]/.venv/lib/python3.13/site-packages/bedrock_agentcore/runtime/context.py:12: PydanticDeprecatedSince20: Support for class-based `config` is deprecated, use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.12/migration/
    class RequestContext(BaseModel):
  ```
  
**Expected behavior**
No deprecation warning.

 - OS: macOS
 - Browser n/a
 - Version 26.3

**Additional context**
`bedrock-agentcore` Version: 1.4.4
I'm using `pydantic` Version: 2.12.5, as required by `bedrock-agentcore`, `mcp`, `pydantic-settings`, `strands-agents`

citizen204:
## Summary

`RequestContext` in `src/bedrock_agentcore/runtime/context.py` uses the Pydantic V1-style inner `class Config` to set `arbitrary_types_allowed = True`. Pydantic V2 deprecated this pattern and emits a `PydanticDeprecatedSince20` warning on every import of `BedrockAgentCoreApp`, adding noise to user applications and test suites.

Replace the inner `Config` class with `model_config = ConfigDict(arbitrary_types_allowed=True)` — the idiomatic Pydantic V2 pattern — so the warning is no longer emitted.

Fixes #320

## Changes

- `src/bedrock_agentcore/runtime/context.py`: import `ConfigDict` from `pydantic`; replace `class Config` with `model_config = ConfigDict(arbitrary_types_allowed=True)`

### DIFF
--- a/src/bedrock_agentcore/runtime/context.py
+++ b/src/bedrock_agentcore/runtime/context.py
@@ -7,7 +7,7 @@ import logging
 from contextvars import ContextVar
 from typing import Any, Callable, Dict, Optional
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field
 
 from ..config_bundle.bundle import ConfigBundleRef
 
@@ -21,10 +21,7 @@ class RequestContext(BaseModel):
     request_headers: Optional[Dict[str, str]] = Field(None)
     request: Optional[Any] = Field(None, description="The underlying Starlette request object")
 
-    class Config:
-        """Allow non-serializable types like Starlette Request."""
-
-        arbitrary_types_allowed = True
+    model_config = ConfigDict(arbitrary_types_allowed=True)
 
 
 class BedrockAgentCoreContext:
--- a/tests/bedrock_agentcore/runtime/test_context.py
+++ b/tests/bedrock_agentcore/runtime/test_context.py
@@ -253,6 +253,22 @@ class TestRequestContext:
 
         assert context.request is mock_request
 
+    def test_request_context_no_pydantic_v1_deprecation_warning(self):
+        """RequestContext must not raise PydanticDeprecatedSince20 on instantiation.
+
+        Regression guard: class-based Config triggers this warning in Pydantic v2;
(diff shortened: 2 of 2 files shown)


valter-silva-au:
Nice — this is the idiomatic V2 pattern and matches the existing `model_config = ConfigDict(...)` usage elsewhere in the repo (e.g. `evaluation/runner/dataset_types.py`).

One suggestion to lock it in: add a regression test in `tests/.../runtime/test_context.py` that instantiates `RequestContext` under `warnings.catch_warnings()` + `warnings.simplefilter("error")` and asserts no `PydanticDeprecatedSince20` warning is raised. That guards against the class-based `Config` silently creeping back in, and gives the fix a failing-before/passing-after anchor. Otherwise LGTM.

citizen204:
Thanks for the suggestion @valter-silva-au — added the regression test in a3175ec. It instantiates `RequestContext` under `warnings.simplefilter("error", PydanticDeprecatedSince20)` so any reintroduction of the class-based `Config` fails the suite.

PR #528, opened 2026-06-16
feat: Add generic BaseAdapter framework for third-party evaluator integration (DeepEval + Autoevals)

stone-coding:
Major changes in this PR:

1. Adds a generic third-party evaluator adapter framework
- Introduces `BaseAdapter` under `evaluation/custom_code_based_evaluators/third_party/`.
- Standardizes the flow: extract fields from AgentCore spans → validate fields → execute third-party metric/scorer → return `EvaluatorOutput`.
- Supports `field_mapper` as an escape hatch for custom or unsupported span formats.

2. Adds DeepEval support
- Adds `DeepEvalAdapter` for running DeepEval `BaseMetric` implementations such as AnswerRelevancy, Faithfulness, Bias, Toxicity, and GEval.
- Converts extracted fields into DeepEval `LLMTestCase`.
- Handles `MissingTestCaseParamsError` and returns actionable `MISSING_REQUIRED_FIELD` errors.

3. Adds Autoevals support
- Adds `AutoevalsAdapter` for Autoevals scorers such as Factuality and ClosedQA.
- Maps AgentCore fields into Autoevals `eval(input, output, expected)` format.
- Supports configurable pass/fail threshold.

4. Adds span parsing support
- Adds parser layer for extracting `input`, `actual_output`, `retrieval_context`, `context`, and `expected_output` from AgentCore ADOT spans and `evaluationReferenceInputs`.
- Supports Strands, LangChain OTel, and OpenInference parser entry points, with shared Phase 1 extraction logic.
- Returns clear `FIELD_EXTRACTION_ERROR` when supported fields cannot be extracted.

5. Updates evaluator input model
- Adds `reference_inputs` to `EvaluatorInput` so `expected_output` can flow from evaluation reference inputs into third-party evaluators.

6. Adds tests
- Adds unit tests covering DeepEval, Autoevals, span parsing, field mapping, missing required fields, and error handling.
- All 42 tests pass.


src/bedrock_agentcore/evaluation/integrations/deepeval/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""DeepEval integration for AgentCore Evaluation."""
+
+from bedrock_agentcore.evaluation.integrations.deepeval.handler import DeepEvalHandler
```
jariy17: Since this is using eval's custom code evaluator please but this under custom_code_based_evaluators.

src/bedrock_agentcore/evaluation/integrations/deepeval/handler.py
```diff
@@ -0,0 +1,135 @@
+"""DeepEval handler that adapts AgentCore Lambda evaluation events to DeepEval metrics."""
+
+import logging
+import threading
+from typing import Any, Callable, Dict, Optional
+
+from deepeval.metrics import BaseMetric
```
jariy17: Please add an integ test for this. Look into tests_integ for examples.

jariy17: Also, let's add this in our pyproject as an optional dependency, so customer's know which deepeval version we support.

src/bedrock_agentcore/evaluation/integrations/deepeval/input_mapper.py
```diff
@@ -0,0 +1,253 @@
...
+    "HallucinationMetric": ["input", "actual_output", "context"],
+    "BiasMetric": ["input", "actual_output"],
+    "ToxicityMetric": ["input", "actual_output"],
+    "GEval": ["input", "actual_output"],
+    "SummarizationMetric": ["input", "actual_output"],
+}
+
+
+@dataclass
+class ParsedEvaluationEvent:
```
jariy17: Please use EvaluatorInput from our code_based_evaluator. No need to duplicate lambda logic.

src/bedrock_agentcore/evaluation/integrations/base.py
```diff
@@ -0,0 +1,302 @@
...
+            event: Either a raw Lambda event dict or an EvaluatorInput instance
+                from bedrock_agentcore.evaluation.custom_code_based_evaluators.models.
+            context: Lambda context object (unused).
+
+        Returns:
+            Success: {"value": float, "label": str, "explanation": str}
+            Error: {"errorCode": str, "errorMessage": str}
+        """
+        try:
+            if isinstance(event, EvaluatorInput):
```
jariy17: ParsedEvaluationEvent and EvaluatorInput look like they're doing the same job — both just turn the raw lambda event into a structured input. __call__ even copies one into the other field-for-field. Is there a reason we need a second type instead of reusing EvaluatorInput?

Proposal: make it a requirement that customers place these adapters within the @code_based_evaluators decorator. That way the adapter stops owning input/output validation and the decorator does it instead. Keeps the adapter focused on just running the eval.

src/bedrock_agentcore/evaluation/integrations/base.py
```diff
@@ -0,0 +1,302 @@
...
+        except _ExecutionTimeout:
+            return _error_response(
+                "METRIC_TIMEOUT",
+                f"{type(self).__name__} exceeded {self.timeout}s timeout.",
+            )
+        except Exception as e:
+            logger.error("Execution failed: %s", e, exc_info=True)
+            return _error_response("METRIC_ERROR", f"{type(self).__name__} failed: {e}")
+
+        return result
```
jariy17: Please return EvaluatorOutput from CodeBasedEvaluator's models in this repo.

src/bedrock_agentcore/evaluation/integrations/autoevals/adapter.py
```diff
@@ -0,0 +1,72 @@
...
+            "input": fields.get("input", ""),
+            "output": fields.get("actual_output", ""),
+        }
+        if fields.get("expected_output"):
+            kwargs["expected"] = fields["expected_output"]
+
+        result = self.scorer.eval(**kwargs)
+
+        score = result.score
+        label = "Pass" if score is not None and score >= 0.5 else "Fail"
```
jariy17: The score >= 0.5 pass threshold is hardcoded. DeepEval honors the metric's own threshold, but Autoevals scorers have no universal 0.5 cutoff, so this will mislabel non-binary scorers. Let's make it configurable.

src/bedrock_agentcore/evaluation/integrations/deepeval/adapter.py
```diff
@@ -0,0 +1,189 @@
...
+    "ContextualRecallMetric": ["input", "actual_output", "expected_output", "retrieval_context"],
+    "HallucinationMetric": ["input", "actual_output", "context"],
+    "BiasMetric": ["input", "actual_output"],
+    "ToxicityMetric": ["input", "actual_output"],
+    "GEval": ["input", "actual_output"],
+    "SummarizationMetric": ["input", "actual_output"],
+}
+
+
+def _get_required_params(metric: BaseMetric) -> List[str]:
```
jariy17: `metric.measure()` already calls `check_llm_test_case_params `with the metric's own `_required_params` and raises `MissingTestCaseParamsError`. 

So we can drop the registry: build the LLMTestCase with whatever fields we have, call measure(), and catch that error.
 
By doing this, we let customers use GEval too — its required fields aren't fixed on the class, they're whatever the customer passes to evaluation_params at construction, so a static registry can never cover it. Letting the metric validate itself handles that case for free.

src/bedrock_agentcore/evaluation/integrations/base.py
```diff
@@ -0,0 +1,302 @@
...
+                return str(val)
+    return ""
+
+
+def extract_fields_from_spans(
+    parsed: ParsedEvaluationEvent,
+) -> Dict[str, Any]:
+    """Extract evaluation fields from AgentCore session spans.
+
+    Parses _eval_log_records from span attributes, filters by target_trace_id,
```
jariy17: Can you tell me what otel agent semantic you are following here? Because I haven't seen any agent SDK emit this _eval_log_records?

src/bedrock_agentcore/evaluation/integrations/base.py
```diff
@@ -0,0 +1,302 @@
...
+                },
+                "evaluationReferenceInputs": parsed.reference_inputs,
+            }
+            return self.field_mapper(raw_event)
+
+        fields = extract_fields_from_spans(parsed)
+        self.validate_fields(fields)
+        return fields
+
+    def validate_fields(self, fields: Dict[str, Any]) -> None:
```
jariy17: Can you add @abstractmethod here please? The no-op default means a subclass that forgets to override it silently skips validation, and bad fields fail deeper in execute instead. Both adapters override it anyway, so abstract just makes each one declare its required fields on purpose.

src/bedrock_agentcore/evaluation/integrations/base.py
```diff
@@ -0,0 +1,302 @@
...
+
+        def target():
+            try:
+                result_holder.append(self.execute(fields))
+            except Exception as e:
+                exception_holder.append(e)
+
+        thread = threading.Thread(target=target, daemon=True)
+        thread.start()
+        thread.join(timeout=self.timeout)
```
jariy17: When the thread "times out" here, it doesn't actually end join just returns back to the caller while the worker keeps running. So if Lambda reuses the same container, we can have a background thread from a previous invocation still executing during the next one. I've heard this is a real failure case, so let's drop the thread machinery and let the AWS Lambda timeout handle it for us instead.

src/bedrock_agentcore/evaluation/integrations/base.py
```diff
@@ -0,0 +1,302 @@
...
+    evaluation logic and returns (score, label, explanation).
+
+    Never raises unhandled exceptions — always returns a valid response dict.
+    """
+
+    DEFAULT_TIMEOUT = 290
+
+    def __init__(
+        self,
+        field_mapper: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
```
jariy17: Can we make `extract_fields_from_spans` the default value of field_mapper in the constructor? Then we have one extraction path instead of the if-field_mapper-else branch.

jariy17 (changes requested):

PR #533, opened 2026-06-17
feat(evaluation): add KMS, tags, online data source, and updated_at to batch eval

jariy17:
## What

Expose batch-evaluation API parameters that `StartBatchEvaluation` / `GetBatchEvaluation` already support but the SDK did not surface:

- **`BatchEvaluationRunConfig.kms_key_arn`** → `StartBatchEvaluation.kmsKeyArn` — encrypt evaluation data at rest with a customer-managed KMS key.
- **`BatchEvaluationRunConfig.tags`** → `StartBatchEvaluation.tags`.
- **`OnlineEvaluationDataSourceConfig`** — new `DataSourceConfig` subclass emitting `onlineEvaluationConfigSource` (`onlineEvaluationConfigArn` + optional `sessionFilterConfig`), to evaluate sessions captured by an existing `OnlineEvaluationConfig`.
- **`BatchEvaluationResult.kms_key_arn`** and **`.updated_at`** — populated from the `GetBatchEvaluation` response.
- Bump `boto3`/`botocore` floor to **1.43.31**, the first release whose `bedrock-agentcore` model includes these parameters.
- Export `OnlineEvaluationDataSourceConfig` from `bedrock_agentcore.evaluation`.

## Why

These are existing, optional service parameters that the runner silently dropped. KMS support in particular is a common requirement for customers with data-at-rest encryption mandates. All additions are backward compatible: every new field is optional and omitted from the request when unset.

## Testing

- **Unit tests** added for each path (KMS/tags sent & omitted, result surfacing, online-source `to_data_source_config` with/without time range, ARN validation, `updated_at` surfacing). Full evaluation suite: **359 passed** (2 pre-existing collection errors in optional `strands-agents-evals` integration tests, unrelated to this change).
- **`ruff check` / `ruff format`**: clean.
- **Manual verification against live `bedrock-agentcore` (us-west-2)**:
  - `updated_at` — a real `GetBatchEvaluation` response was parsed by the result model; `updated_at` populated correctly. ✅
  - `kms_key_arn` — a real `StartBatchEvaluation` call **accepted and acted on** `kmsKeyArn`; it reached the service's KMS-encrypt step (request rejected only because the test key's policy lacks `kms:GenerateDataKey` for the service role — i.e. the parameter is wired through and honored end-to-end). ✅
  - online source / `botocore` param validation — verified via `botocore.Stubber` against the real service model. ✅

> Note: the live `kms_key_arn` test stopped at the KMS key-policy permission boundary by design — a full green run needs a key whose policy grants the AgentCore service role `kms:GenerateDataKey`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,8 +26,8 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
-    "boto3>=1.43.0",
-    "botocore>=1.43.0",
+    "boto3>=1.43.31",
+    "botocore>=1.43.31",
     "pydantic>=2.0.0,<2.41.3",
     "urllib3>=1.26.0",
     "starlette>=0.46.2",
--- a/src/bedrock_agentcore/evaluation/runner/batch/batch_evaluation_models.py
+++ b/src/bedrock_agentcore/evaluation/runner/batch/batch_evaluation_models.py
@@ -111,6 +111,46 @@ class CloudWatchDataSourceConfig(BaseModel, DataSourceConfig):
         }
 
 
+class OnlineEvaluationDataSourceConfig(BaseModel, DataSourceConfig):
+    """Online-evaluation data source — pulls spans from an existing OnlineEvaluationConfig.
+
+    .. warning::
+        This feature is in preview and may change in future releases.
+
+    Unlike :class:`CloudWatchDataSourceConfig`, this source does not filter by the
+    session IDs generated during agent invocation: the service reads sessions already
+    captured by the referenced OnlineEvaluationConfig, optionally narrowed to a time
+    window. As a result it is typically used to (re-)evaluate previously recorded
+    sessions rather than sessions produced by the current ``agent_invoker`` run.
+
+    Attributes:
+        online_evaluation_config_arn: ARN of the OnlineEvaluationConfig whose
+            captured sessions are evaluated.
+        use_invocation_time_range: When ``True`` (default), the runner supplies the
+            ``sessionFilterConfig`` time window from the earliest/latest session times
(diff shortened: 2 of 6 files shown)


notgitika (approved):
looks good to me, it is backwards compatible and just config changes. curious to see if you think adding integ tests here would be useful

pyproject.toml
```diff
@@ -26,8 +26,8 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
-    "boto3>=1.43.0",
-    "botocore>=1.43.0",
```
notgitika, line 30: can we remove the dependency bump from this PR?

src/bedrock_agentcore/evaluation/runner/batch/batch_evaluation_models.py
```diff
@@ -293,6 +347,8 @@ class BatchEvaluationRunConfig(BaseModel):
     polling_timeout_seconds: int = 1800
     polling_interval_seconds: int = 30
     simulation_config: Optional[SimulationConfig] = None
+    kms_key_arn: Optional[str] = None
```
notgitika, line 350: maybe adding some client side validation on the kms key shape could be good

pyproject.toml
jariy17: What's wrong with a dependency bump here? We need an updated boto3 so the new parameters shows up.

src/bedrock_agentcore/evaluation/runner/batch/batch_evaluation_models.py
jariy17: Good Idea but if a customers put an invalid arn here, the service will return a ValidationException.

PR #534, opened 2026-06-17
fix(test): update tests to be consistent with new model

Hweinstock:
### Problem 
There was a breaking change in strands evals https://github.com/strands-agents/evals/pull/241, this causes our tests to fail. 

Note: we never call it directly, so the implementation is safe. 

### Solution
- migrate tests to new shape.


Hweinstock:
This is an evals only change, unrelated to identity and memory.

tejaskash (approved):

PR #544, opened 2026-06-23
fix(memory): guard retrieve_customer_context against empty message content

### LINKED ISSUE #543: retrieve_customer_context raises IndexError on empty message content
**Describe the bug**

`AgentCoreMemorySessionManager.retrieve_customer_context` (a `MessageAddedEvent` callback) indexes the first content block of the last message without checking that the content list is non-empty:

```python
# src/bedrock_agentcore/memory/integrations/strands/session_manager.py
if not messages or messages[-1].get("role") != "user" or "text" not in messages[-1].get("content")[0]:
    return None
```

When the last message has an empty `content` list (`[]`), `messages[-1].get("content")[0]` raises `IndexError`. As this runs inside a `MessageAddedEvent` hook, it propagates out of the event loop and aborts the agent invocation (a `500` on AgentCore Runtime). The guard covers `not messages` but not an empty `content` list.

The callback is registered for every `MessageAddedEvent` in sync mode regardless of `retrieval_config`, and the failing line runs before the `retrieval_config` check — so any attached `AgentCoreMemorySessionManager` is affected. It triggers when a `user`-role message with empty `content` is the most recent message. (An empty-content `assistant` message is safe: the `role != "user"` check short-circuits before the index access.)

**To Reproduce**

An empty-`content` user message is reachable through the public agent API: Strands' `Agent._convert_prompt_to_messages` appends a `list[Message]` prompt verbatim, so `agent([{"role": "user", "content": []}])` fires a `MessageAddedEvent` whose `messages[-1]` is `{"role": "user", "content": []}`. (By contrast `agent([])` and an empty `list[ContentBlock]` map to no message and are unaffected.)

Runnable end-to-end repro against a real `strands.Agent` (raises on `main`, clean with the suggested fix):

```python
import asyncio
from unittest.mock import Mock, patch

from strands.agent.agent import Agent
from strands.hooks import MessageAddedEvent
from strands.models.model import Model
from strands.types.session import Session, SessionType

from bedrock_agentcore.memory.integrations.strands.config import AgentCoreMemoryConfig, RetrievalConfig


class FakeModel(Model):  # the crash happens at MessageAddedEvent time, before the model is ever called
    def get_config(self): return {}
    def update_config(self, **k): pass
    async def structured_output(self, *a, **k):
        if False:
            yield
    async def stream(self, *a, **k):
        yield {"messageStart": {"role": "assistant"}}
        yield {"messageStop": {"stopReason": "end_turn"}}


def build_manager():
    client = Mock()
    client.retrieve_memories.return_value = []
    cfg = AgentCoreMemoryConfig(
        memory_id="m", session_id="s", actor_id="a",
        retrieval_config={"prefs/{actorId}/": RetrievalConfig(top_k=5)},
    )
    with (
        patch("bedrock_agentcore.memory.integrations.strands.session_manager.MemoryClient", return_value=client),
        patch("boto3.Session"),
        patch("strands.session.repository_session_manager.RepositorySessionManager.__init__", return_value=None),
    ):
        m = AgentCoreMemorySessionManager(cfg)
        m.session_id = "s"
        m.session = Session(session_id="s", session_type=SessionType.AGENT)
    return m


async def main():
    manager = build_manager()
    agent = Agent(model=FakeModel())
    # exactly what register_hooks() wires up in sync mode (session_manager.py:941)
    agent.hooks.add_callback(MessageAddedEvent, lambda e: manager.retrieve_customer_context(e))
    await agent.invoke_async([{"role": "user", "content": []}])  # IndexError on main


asyncio.run(main())
```

**Expected behavior**

Return early when the last message has an empty `content` list, instead of raising.

**Suggested fix**

```python
if not messages or messages[-1].get("role") != "user":
    return None
content = messages[-1].get("content")
if not content or "text" not in content[0]:
    return None
```

**Environment**

- `bedrock-agentcore` 1.9.1 (line unchanged on `main`)
- `strands-agents` 1.39.0

dtaniwaki:
Thanks for maintaining this project! Would appreciate a review when you get a chance.

Fixes #543

## Problem

`AgentCoreMemorySessionManager.retrieve_customer_context` (registered as a `MessageAddedEvent` callback) indexes the first content block of the last message without checking that the content list is non-empty:

```python
if not messages or messages[-1].get("role") != "user" or "text" not in messages[-1].get("content")[0]:
    return None
```

When the last message has an empty `content` list (`[]`), `messages[-1].get("content")[0]` raises `IndexError`. Because this runs inside a `MessageAddedEvent` hook, the exception propagates out of the event loop and aborts the whole agent invocation (surfacing as a `500` on AgentCore Runtime). The existing guard handles `not messages` but not an empty `content` list. (See #543 for a runnable end-to-end reproduction.)

## Solution

Check that `content` is non-empty before indexing the first block, returning early otherwise. Added a regression test (`test_retrieve_customer_context_empty_content`) that fails without the guard and passes with it. All existing tests in the session manager suite still pass, and `ruff check` / `ruff format` are clean.

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/session_manager.py
@@ -848,7 +848,10 @@ class AgentCoreMemorySessionManager(RepositorySessionManager, SessionRepository)
             event (MessageAddedEvent): The message added event containing the agent and message data.
         """
         messages = event.agent.messages
-        if not messages or messages[-1].get("role") != "user" or "text" not in messages[-1].get("content")[0]:
+        if not messages or messages[-1].get("role") != "user":
+            return None
+        content = messages[-1].get("content")
+        if not content or "text" not in content[0]:
             return None
         if not self.config.retrieval_config:
             # Only retrieve LTM
--- a/tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
+++ b/tests/bedrock_agentcore/memory/integrations/strands/test_agentcore_memory_session_manager.py
@@ -1169,6 +1169,31 @@ class TestAgentCoreMemorySessionManager:
                     result = manager.retrieve_customer_context(event)
                     assert result is None
 
+    def test_retrieve_customer_context_empty_content(self, agentcore_config_with_retrieval, mock_memory_client):
+        """Empty content list on the last message must not raise IndexError."""
+        with patch(
+            "bedrock_agentcore.memory.integrations.strands.session_manager.MemoryClient",
+            return_value=mock_memory_client,
+        ):
+            with patch("boto3.Session") as mock_boto_session:
+                mock_session = Mock()
+                mock_session.region_name = "us-west-2"
+                mock_session.client.return_value = Mock()
+                mock_boto_session.return_value = mock_session
+
+                with patch(
+                    "strands.session.repository_session_manager.RepositorySessionManager.__init__", return_value=None
+                ):
+                    manager = AgentCoreMemorySessionManager(agentcore_config_with_retrieval)
+
+                    mock_agent = Mock()
+                    mock_agent.messages = [{"role": "user", "content": []}]
+
+                    event = MessageAddedEvent(agent=mock_agent, message={"role": "user", "content": []})
+                    result = manager.retrieve_customer_context(event)
+                    assert result is None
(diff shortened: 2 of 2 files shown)


nborges-aws (approved):

nborges-aws:
Breaking change bot failure is false alarm. Reported "No Breaking Changes Detected", but errored while attempting to post comment to PR, causing workflow to fail.

PR looks good to me. Thanks for the fix!

PR #545, opened 2026-06-23
fix(evaluation): make EvaluatorOutput.label optional for error responses

### LINKED ISSUE #541: Inconsistency between code and documentation for custom code-based evaluators
**Describe the bug**
The documentation for Amazon Bedrock AgentCore code-based evaluators says a Lambda may return one of two response shapes:

1. Success response:
   - `label`
   - optional `value`
   - optional `explanation`

2. Error response:
   - `errorCode`
   - `errorMessage`

However, when using the Python SDK/helper model `bedrock_agentcore.evaluation.EvaluatorOutput`, constructing an error-only response without `label` raises a Pydantic validation error because `label` is still required by the SDK model.

This creates a mismatch between the documented Lambda response contract and the actual SDK behavior. In practice, valid documented error responses can fail at runtime unless a placeholder `label` is included.

**To Reproduce**
1. Install `bedrock-agentcore` in Python.
2. Create a minimal code-based evaluator or simple repro script.
3. Construct an error response using the documented error-only shape, for example:
   ```python
   from bedrock_agentcore.evaluation import EvaluatorOutput

   EvaluatorOutput(
       errorCode="InvalidJudgeResponse",
       errorMessage="Example error"
   )
4. Observe that the SDK raises a validation error because label is required.

**Expected behavior**
A code-based evaluator error response documented as:
```
{
  "errorCode": "VALIDATION_FAILED",
  "errorMessage": "Input spans missing required tool call attributes."
}
```

should be accepted by the SDK/helper model and should not require label.

Either:
- the SDK should allow error-only responses without label, or
- the documentation should explicitly state that label is still required by the SDK model even for error responses.

**Additional context**
- Documentation referenced:
https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/code-based-evaluators.html
- Observed package version: bedrock-agentcore 1.14.1
- The validation failure appears to come from the Python SDK model layer, not from the documented wire contract itself.
- This mismatch can surface as test failures and may also cause evaluator runtime failures or generic internal server errors if the Lambda attempts to return a documented error-only response through EvaluatorOutput.
- As a workaround, we had to populate label="" on error responses even though that is not consistent with the documented error response schema.

citizen204:
## Summary

The documented Lambda contract for code-based evaluators allows error-only responses that omit `label`:

```json
{ errorCode: VALIDATION_FAILED, errorMessage: ... }
```

However, constructing such a response with the SDK helper model raised a Pydantic `ValidationError` because `label` was unconditionally required (`label: str`).

This PR makes `label` optional **only when `errorCode` is set**, preserving the existing validation for success responses:

- **Success response** (no `errorCode`): `label` is still required — a `ValidationError` is raised if omitted.
- **Error response** (`errorCode` set): `label` may be omitted.

Fixes #541

## Changes

- `src/bedrock_agentcore/evaluation/custom_code_based_evaluators/models.py`:
  - `label: str` → `label: Optional[str] = None`
  - Added `@model_validator(mode=after)` that raises `ValueError` when both `label` and `errorCode` are absent
  - Updated docstring to document the two response shapes

- `tests/bedrock_agentcore/evaluation/custom_code_based_evaluators/test_models.py`:
  - Updated `test_label_required` → `test_label_required_without_error_code` with match string
  - Added `test_error_response_without_label` and `test_error_response_with_label`

### DIFF
--- a/src/bedrock_agentcore/evaluation/custom_code_based_evaluators/models.py
+++ b/src/bedrock_agentcore/evaluation/custom_code_based_evaluators/models.py
@@ -2,7 +2,7 @@
 
 from typing import Any, Dict, List, Optional
 
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 
 class ReferenceInput(BaseModel):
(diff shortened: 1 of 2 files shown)


jariy17 (dismissed):
Thanks for fixing. Please rebase and I'll approve it again

jariy17:
The breaking change workflow flagged that label's type has changed. However, it went from str to Optional[str] which is not a breaking change.

aidandaly24 (approved):

PR #546, opened 2026-06-23
feat(payments): Add LangGraph integration for payment handling

ragsu43:
*Description of changes:*

Added LangGraph middleware and config files for developer integration with ACP.


src/bedrock_agentcore/payments/integrations/langgraph/config.py
```diff
@@ -0,0 +1,151 @@
+"""Configuration for AgentCorePaymentsMiddleware (LangGraph integration)."""
+
+from dataclasses import dataclass, field
+from typing import Any, Callable, Dict, List, Optional
+
+from ..handlers import PaymentResponseHandler
+
+
+@dataclass
+class AgentCorePaymentsConfig:
```
rajuans: As discussed, can we use the config defined in the integration folder.

src/bedrock_agentcore/payments/integrations/langgraph/errors.py
```diff
@@ -0,0 +1,45 @@
...
+
+class ErrorResolution(Enum):
+    """Return value from on_payment_error callback."""
+
+    RETRY = "retry"
+    PROPAGATE = "propagate"
+
+
+@dataclass
+class PaymentErrorContext:
```
rajuans, line 21: We already have error defined in the common package. Let's use the existing error class.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -0,0 +1,831 @@
...
+    make_get_payment_session_tool,
+    make_http_request_tool,
+    make_list_payment_instruments_tool,
+)
+
+logger = logging.getLogger(__name__)
+
+# Deterministic error messages per exception type.
+# The LLM sees these messages and should follow the "Do not retry" instruction.
+_ERROR_MESSAGES: Dict[type, str] = {
```
rajuans: Can we define at the root level of payments SDK instead of in langgraph specific package?

src/bedrock_agentcore/payments/integrations/langgraph/errors.py
ragsu43: Given how the callback function would be implemented by developers without PaymentErrorContext, it seems to me that passing in a context object with the exception info and all necessary state/context fields for error handling gives the simplest dev experience. Based on what I have found, there are some big blockers against deprecating PaymentErrorContext in favor of just passing in the exception:

there are 4 main fields that langgraph devs cannot access from the callback unless it is passed into the callback:
1. **Retry state** (number of retries on this tool call):

- can't track it directly in the config class because dev does not have access to "request" (field from wrap_tool_call), so no way to know what tool-name the retry is for unless we pass that in, so either way we would need to pass something into the callback.
- As mentioned above, middleware in langgraph doesn't receive graph state at all (which is the only equivalent of agent.state in strands), so even if we could add a custom field to the graph's State dict, the middleware can't read/write to it like "agent.state" in Strands.
- Only option is developer tracks it in some closure dict, but again we would need to pass in tool_call_id, in which case just having retry_count in the context object is much easier for the dev.

2. **tool_name**:

- as mentioned above can't be accessed by dev if we deprecate PaymentErrorContext unless we pass it in anyway, since the developer has no access to the tool call "request" itself.

3. **tool_args**: same problem as tool_name... lives in ToolCallRequest inside wrap_tool_call with no external visibility WITHOUT some arg passing into the callback

4. **payment_required_request** (the parsed 402 body):

- only exists as a local var in wrap_tool_call, developer doesn't see the 402 response from callback function so they can't parse it manually either so we would have to pass this in.


Based on these constraints, it seems to me that keeping the context object and passing it into the callback is the best option (developer just accesses everything needed for state in one object), but if anything seems off or something I haven't considered please let me know!

src/bedrock_agentcore/payments/integrations/langgraph/errors.py
```diff
@@ -0,0 +1,45 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import TYPE_CHECKING, Any, Dict, Optional
+
+if TYPE_CHECKING:
+    from .config import AgentCorePaymentsConfig
+
```
rajuans, line 11: **Wrong import path under TYPE_CHECKING**

This imports from `.config` (i.e. `langgraph/config.py`) but the config class lives at `integrations/config.py`. Should be:
```python
from ..config import AgentCorePaymentsConfig
```

Won't crash at runtime (guarded by `TYPE_CHECKING`), but will break mypy/pyright.

src/bedrock_agentcore/payments/integrations/config.py
```diff
@@ -65,51 +52,52 @@ class AgentCorePaymentsPluginConfig:
...
+    network_preferences_config: Optional[List[str]] = None
     auto_payment: bool = True
-    max_interrupt_retries: int = 5
     agent_name: Optional[str] = None
     bearer_token: Optional[str] = None
     token_provider: Optional[Callable[[], str]] = None
     payment_tool_allowlist: Optional[List[str]] = None
     provide_http_request: bool = True
     post_payment_retry_delay_seconds: float = 3.0
+    max_interrupt_retries: int = 5
```
rajuans, line 63: **Type annotation contradicts validation**

The field is typed `Dict[str, Any]` but validation at `__post_init__` enforces `isinstance(v, PaymentResponseHandler)` for all values. Consider:
```python
custom_handlers: Optional[Dict[str, "PaymentResponseHandler"]] = field(default=None)
```
This way the type signature matches the runtime contract and callers get proper type-checking.

tests/bedrock_agentcore/payments/integrations/langgraph/test_functional.py
```diff
@@ -0,0 +1,534 @@
...
+                return super().extract_status_code(result)
+
+            def extract_headers(self, result):
+                self.extract_called = True
+                return super().extract_headers(result)
+
+            def apply_payment_header(self, tool_input, payment_header):
+                self.inject_called = True
+                return super().apply_payment_header(tool_input, payment_header)
+
```
rajuans, line 349: **Duplicate test block (copy-paste leftover)**

Lines below this point duplicate the `test_custom_handler_full_flow` test verbatim — the `TrackingHandler` class, middleware setup, and assertions are repeated inside this method body. Appears to be a copy-paste accident. The duplicate code still runs (appended to this method), but adds confusion and redundant test time. Safe to remove.

src/bedrock_agentcore/payments/integrations/config.py
```diff
@@ -1,62 +1,49 @@
...
-            payment header before allowing the tool to be retried. The x402
-            EIP-3009 ``transferWithAuthorization`` contract requires
-            ``block.timestamp > validAfter`` (strict greater-than). Some signing
-            services set ``validAfter`` close to the current time, which can
-            cause the merchant facilitator to submit before ``validAfter``
-            elapses, producing a misleading "invalid_payload" response. A small
-            delay between signing and retry lets the chain advance one block so
-            the authorization is valid by the time the seller submits. Defaults
-            to 3.0 seconds (about one Base Sepolia block). Set to 0 to disable.
+        payment_connector_id: Payment connector ID (optional).
```
rajuans, line 24: Is this a new attribute? A manager can have many connectiors. What use case this attribute is going to solve?

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -0,0 +1,831 @@
...
+            user_id=self.config.user_id,
+            payment_instrument_id=self.config.payment_instrument_id,
+            payment_session_id=self.config.payment_session_id,
+            payment_required_request=payment_required_request,
+            network_preferences=self.config.network_preferences_config,
+            client_token=str(uuid.uuid4()),
+            payment_connector_id=self.config.payment_connector_id,
+        )
+
+    def _create_auto_session(self) -> None:
```
rajuans, line 375: Why are we creating autoSession? Agent may not have permission to create session. Creating session in agent will be risky as they can bypass the payments limit (budget) check.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -0,0 +1,831 @@
...
+                        retry_body.get("error", "unknown error")
+                        if isinstance(retry_body, dict)
+                        else "unknown error"
+                    )
+                    return self._error_tool_message(
+                        request,
+                        PaymentError(
+                            f"Payment was signed but rejected by the server ({error_detail})."
+                        ),
+                    )
```
rajuans: Duplicate code block. Can you make sure to remove duplicate code for the sake of maintainability?

rajuans (changes requested):

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -0,0 +1,791 @@
...
+                exception_message=str(current_exception),
+                tool_name=tool_name,
+                tool_args=tool_args,
+                payment_required_request=payment_required_request,
+                config=self.config,
+                retry_count=retry_count,
+            )
+
+            try:
+                resolution = self.config.on_payment_error(ctx)
```
aidandaly24, line 522: This drops async `on_payment_error` callbacks on the sync path. The async version at line 727 checks `inspect.iscoroutinefunction` before awaiting, but here we call the callback directly.

If someone registers an `async def` callback and runs the agent with `agent.invoke()`, this returns an un-awaited coroutine. It isn't a str and it isn't `ErrorResolution.RETRY`, so line 568 returns None and we fall through to PROPAGATE. The callback body never runs, the retry/fix is silently discarded, the LLM gets the default error message, and you leak a `coroutine was never awaited` RuntimeWarning.

Either run it to completion here (e.g. via an event loop), or detect a coroutine function up front and raise a clear error so it fails loudly instead of silently.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -0,0 +1,791 @@
...
+        from bedrock_agentcore.payments.integrations.handlers import GenericPaymentHandler as _GH
+
+        _rh = _GH()
+        retry_status = _rh.extract_status_code(retry_prepared)
+        if retry_status != 402:
+            fallback = self._fallback_detect_402(retry_result.content)
+            if fallback:
+                retry_status = 402
+        if retry_status == 402:
+            retry_body = _rh.extract_body(retry_prepared) or {}
```
aidandaly24, line 322: Copy-paste drift from `_check_post_payment_rejection`. When the fallback detects the 402 (lines 369-371) this never reassigns `_rh` to `_FallbackHandler(fallback)`, unlike the sibling method at line 343.

So `_rh` stays the `GenericPaymentHandler`, `extract_body` here looks for the `PAYMENT_REQUIRED:` marker, doesn't find it in raw JSON, and returns None which becomes `{}`. For a raw-JSON or MCP tool that returns `{"statusCode":402,"body":{"error":"budget exceeded"}}` on the recovery retry, the LLM sees `rejected after recovery (unknown)` instead of the real `budget exceeded`.

Reassign `_rh = _FallbackHandler(fallback)` inside the fallback branch, matching `_check_post_payment_rejection`.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -0,0 +1,791 @@
...
+            return None
+
+        has_custom_handler = self.config.custom_handlers is not None and tool_name in self.config.custom_handlers
+
+        if has_custom_handler:
+            detection_handler = self.config.custom_handlers[tool_name]
+        else:
+            from bedrock_agentcore.payments.integrations.handlers import GenericPaymentHandler
+
+            detection_handler = GenericPaymentHandler()
```
aidandaly24, line 224: Detection only ever uses `GenericPaymentHandler` on the non-custom path (marker only), plus the JSON-only `_fallback_detect_402`. The name-based registry (`http_request` maps to `HttpRequestPaymentHandler`) is only consulted on the injection side through `_get_handler`.

That means a tool returning 402 in the legacy `Status Code:` / `Headers:` / `Body:` text-block format, which is exactly what `HttpRequestPaymentHandler.extract_status_code` was written to parse (handlers.py:479), never gets detected here. Generic finds no marker and returns None, and `_fallback_detect_402` can't `json.loads` `Status Code: 402`. The 402 passes through as a normal result and payment never fires. This bites anyone who brings their own http_request-style tool with `provide_http_request=False`.

Consider routing detection through `get_payment_handler(tool_name, tool_args)` as well, not just injection.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -0,0 +1,791 @@
...
+
+        status_code = detection_handler.extract_status_code(prepared)
```
aidandaly24: Shape-contract gotcha for custom handlers. We hand the detection handler the `_prepare_for_handler` output, which is `{"content": [{"text": ...}]}`. That works for `GenericPaymentHandler`, but the exported `MCPRequestPaymentHandler` expects the raw result with top-level `structuredContent`/`httpStatus`.

Since config accepts any `PaymentResponseHandler`, `custom_handlers={"my_mcp_tool": MCPRequestPaymentHandler()}` silently never detects a 402: `extract_status_code` reads keys that aren't in the prepared shape and returns None, and because `has_custom_handler` is True the lenient fallback at line 272 is skipped.

Either feed custom handlers the raw content, or document and enforce that a custom handler must consume the prepared shape.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -0,0 +1,791 @@
...
+        if retry_status == 402:
+            retry_body = _retry_handler.extract_body(retry_prepared) or {}
+            error_detail = retry_body.get("error", "unknown error") if isinstance(retry_body, dict) else "unknown error"
+            return self._error_tool_message(
+                request,
+                PaymentError(f"Payment was signed but rejected by the server ({error_detail})."),
+            )
+        return None
+
+    def _check_post_recovery_rejection(
```
aidandaly24: Three chunks of near-duplicated logic in this file, and the drift called out at line 373 is a direct result. `_check_post_recovery_rejection` is basically `_check_post_payment_rejection`, `_ainvoke_error_handler` mirrors `_invoke_error_handler` (~90 lines each, and their inner retry block re-implements `_inject_payment_header`), and the `GenericPaymentHandler` plus `_fallback_detect_402` detect ladder shows up in three methods.

A single `_resolve_402(content)` helper plus a shared retry/inject body (the async version just wraps it with `to_thread`) would collapse all of this and stop the two rejection checks from diverging again.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -0,0 +1,791 @@
...
+        prepared = self._prepare_for_handler(result.content)
+        if prepared is None:
+            return None
+
+        has_custom_handler = self.config.custom_handlers is not None and tool_name in self.config.custom_handlers
+
+        if has_custom_handler:
+            detection_handler = self.config.custom_handlers[tool_name]
+        else:
+            from bedrock_agentcore.payments.integrations.handlers import GenericPaymentHandler
```
aidandaly24: Imports belong at the top of the file, not inside methods. This one (`from ...handlers import GenericPaymentHandler`) is repeated inside `_detect_402`, `_check_post_payment_rejection` (335), and `_check_post_recovery_rejection` (364). Same story for `import json as _json` (187), `from .errors import ...` (537, 709), `import asyncio` (635, 706), and `import inspect` (707).

None of these have a cycle risk (`.errors` doesn't import middleware), so hoist them to the module top. The repeated handler import is also part of what let the two rejection checks drift.

aidandaly24 (changes requested):
Nice work on this. The overall flow (detect, sign, inject, retry, across sync and async, plus the recovery callback and auto-session) reads well and the unit coverage is thorough. Requesting changes on a couple of correctness issues before this goes in, plus some cleanup.

Blocking:
- Sync path silently drops async `on_payment_error` callbacks (middleware.py:555).
- `_check_post_recovery_rejection` surfaces the wrong error detail because it never swaps in the fallback handler (middleware.py:373).

Should fix:
- Detection ignores the name-based handler registry, so legacy text-format 402s get missed (middleware.py:267).
- The exported `MCPRequestPaymentHandler` can't actually be used as a `custom_handlers` entry because of a shape mismatch (middleware.py:269).

Cleanup:
- The sync/async and the two rejection-check methods are near-duplicates and have already drifted, worth collapsing (middleware.py:353).
- Move the inline imports up to the top of the file per our style (middleware.py, several spots).

One process note: the header on test_functional.py says any change to this middleware has to be validated against a live Base Sepolia testnet before merge. Please confirm that run has happened.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
ragsu43: Great catch. Added coroutine detection in the sync path that raises TypeError so that we fail loudly. (c4ceaf6)

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
ragsu43: Great catch... by not also trying the name-based handler result (HttpRequestPaymentHandler, which was made specifically for the Strands-native http_request tool output format), we miss the case where a developer used the Strands-native tool for the opt-out. Although this is a narrow edge case it's risk-free and doesn't hurt to cover. Added a third fallback to run the result through the get_payment_handler() result as well if still no 402 detected.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
ragsu43: Great point, devs will expect to handle whatever shape is outputted by their tool, not our specific prepared shape. Custom handlers now get fed raw content in detect_402 (field added to _DetectionResult). Also updated extract_payment_request() to make sure if there is a custom handler it gets fed detection.raw_content instead of detection.prepared. (076d6ed)

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
ragsu43: Great points. Collapsed duplicated code with a refactor for most (wrap and awrap methods seem to be at the practical minimum, duplication is just Python sync/async split):

  1. Two rejection checks → one: _check_post_payment_rejection + _check_post_recovery_rejection → _check_retry_rejection(request, result, error_context) with a context string parameter
  
  2. Repeated inject logic in error handlers → one: _inject_for_error_retry(request, tool_name, tool_args, payment_header)
  
3. Repeated context building → one: _build_error_context(...)

4. Repeated resolution handling → one: _handle_callback_resolution(resolution, request)

aidandaly24 (approved):
Thanks for making the fixes, approved.

PR #568, opened 2026-07-06
feat: third-party eval metrics adapter (DeepEval + Autoevals) with strands-evals mappers

stone-coding:
## Summary

  Add DeepEvalAdapter and AutoevalsAdapter that integrate third-party evaluation metrics with
  AgentCore's code-based evaluator framework.

  - **Span mapping via strands-evals:** Uses `detect_otel_mapper()` from `strands-agents-evals` for
    auto-detection of Strands, OpenInference LangChain, and OpenTelemetry LangChain span formats.
    Bridge function converts `Session` → `SpanMapResult` for adapter consumption.
  - **custom_mapper** parameter lets customers bypass built-in mappers for unsupported frameworks:
    - DeepEval: `custom_mapper: Callable[[EvaluatorInput], LLMTestCase]`
    - Autoevals: `custom_mapper: Callable[[EvaluatorInput], Dict[str, Any]]`
  - **tools_called + expected_tools** extraction from execute_tool spans — enables
    ToolCorrectnessMetric and ArgumentCorrectnessMetric
 - assertions from reference_inputs mapped to LLMTestCase.context → should now say assertions from reference_inputs mapped to LLMTestCase.context (None when no assertions)
  - **reference_inputs** full pipeline: expectedResponse → expected_output,
    expectedTrajectory → expected_tools, assertions → context
  - Never raises unhandled exceptions — all error paths return valid EvaluatorOutput with
    structured errorCode/errorMessage
  - Pinned dependency: `strands-agents-evals>=1.0.0,<2.0.0`

  ## Test plan

  - 38 unit tests passing (`pytest tests/.../third_party/ -v`)
  - 21 E2E tests passing across:
    - DeepEval: 12 metrics (RAG, Agentic, Custom/GEval, Contextual, Non-LLM, Safety + OpenInference +
  OpenTelemetry)
    - Autoevals: 6 metrics (LLM-Judge, LLM-Scorer, Deterministic + OpenInference + OpenTelemetry)
    - Custom mapper: 3 tests (Google ADK, OpenAI Agents, Autoevals path)
  - E2E validated against live AgentCore service (evaluate() API → Lambda → metric → score)


src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Example (customer mapper returning eval kwargs)::
+
+        adapter = AutoevalsAdapter(
+            scorer=Factuality(),
+            customer_mapper=lambda ev: {
+                "input": ev.session_spans[0]["attributes"]["question"],
+                "output": ev.session_spans[0]["attributes"]["answer"],
+                "expected": "the expected answer",
+            },
```
chianhw: Do we need to accept aws lambda?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+    """
+
+    def __init__(
+        self,
+        scorer: Any,
+        customer_mapper: Optional[Callable[[EvaluatorInput], Dict[str, Any]]] = None,
```
chianhw: 1. We have not updated this type Dict[str, Any]?
2. Regarding naming, "Custom mapper" would be a better choice for describing an arbitrary mapper provided by a customer?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
```diff
@@ -0,0 +1,86 @@
...
+        threshold: float = 0.5,
```
chianhw: `label: Optional[str] = None` label in EvaluatorOutput is optional. 
We don't need to have a default value to generate a label when a metric doesn't have a label and the user doesn't provide any threshold. Can we set default as None?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
```diff
@@ -0,0 +1,86 @@
...
+        """Initialize the adapter.
+
+        Args:
+            scorer: An Autoevals scorer instance (e.g. Factuality(), ClosedQA()).
+            customer_mapper: Optional callable that receives the EvaluatorInput and
+                returns a dict of kwargs for scorer.eval(). Bypasses default span
+                mapping when provided. Expected keys: input, output, expected (optional).
+            threshold: Score threshold for Pass/Fail determination. Defaults to 0.5.
+        """
+        self.scorer = scorer
```
chianhw: this naming is not aligned with DeepEval adaptor. Looks like you haven't updated AutoevalsAdapter.
If you haven't completed end to end tests for AutoevalsAdapter, you should not include it in your PR.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
```diff
@@ -0,0 +1,101 @@
...
+        adapter = DeepEvalAdapter(metric=metric)
+
+    Example (customer mapper returning LLMTestCase)::
+
+        adapter = DeepEvalAdapter(
+            metric=AnswerRelevancyMetric(threshold=0.7),
+            customer_mapper=lambda ev: LLMTestCase(
+                input=ev.session_spans[0]["attributes"]["user_query"],
+                actual_output=ev.session_spans[0]["attributes"]["response"],
+            ),
```
chianhw: using lambda? same as above

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+    """
+
+    def __init__(
+        self,
+        metric: BaseMetric,
+        customer_mapper: Optional[Callable[[EvaluatorInput], LLMTestCase]] = None,
```
chianhw: naming? same as above

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
```diff
@@ -0,0 +1,101 @@
...
+        Args:
+            metric: A DeepEval BaseMetric instance (e.g. AnswerRelevancyMetric).
+            customer_mapper: Optional callable that receives the EvaluatorInput and
+                returns a LLMTestCase. Bypasses default span mapping when provided.
+            model: Optional model override for the metric's LLM.
+        """
+        self.metric = metric
+        self.customer_mapper = customer_mapper
+        if model is not None:
+            self.metric.model = model
```
chianhw: Where do we use `self.metric.model`?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/base.py
```diff
@@ -0,0 +1,51 @@
...
+        session_spans: Raw ADOT span dicts from the evaluation service.
+        reference_inputs: Optional ReferenceInput list for expected_output.
+
+    Returns:
+        SpanMapResult with extracted fields.
+
+    Raises:
+        ValueError: If no mapper can extract data from the spans.
+    """
+    result = _strands_mapper.map(session_spans)
```
chianhw: We can check the span scope name and then decide which mapper to use.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
stone-coding: Changed threshold default to None. When no threshold is set,
  no Pass/Fail judgment is made. Note: EvaluatorOutput's
  validator currently requires a label for success responses —
  I'm returning the score as the label string when threshold is
  None. Let me know if you'd rather relax the validator to
  allow label=None.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
stone-coding: 1.The type Dict[str, Any] is correct — Autoevals has no typed
  input class like DeepEval's LLMTestCase. Scorers take plain
  kwargs (input, output, expected) and the dict gets unpacked
  as metric.eval(**kwargs). Different scorers need different
  keys, so a generic dict is the right contract.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
stone-coding: Replaced lambda examples with named functions in docstrings

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/strands.py
```diff
@@ -0,0 +1,158 @@
...
+        for span in session_spans:
+            scope = span.get("scope", {})
+            if not isinstance(scope, dict):
+                continue
+            if scope.get("name") != SCOPE_STRANDS:
+                continue
+            attributes = span.get("attributes", {})
+            if attributes.get("gen_ai.operation.name") == "invoke_agent":
+                return span
+        return None
```
chianhw: Does this mapper only take the first agent span? For strands, the OtelSpanMapper in OpenTelemetryMapper package use the last agent span.  That is, the agent span with the latest end_time per trace is the one retained.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
```diff
@@ -0,0 +1,101 @@
...
+                    errorMessage=f"Field(s) {missing} required by {metric_name} but not found in evaluation event. "
+                    f"Provide a customer_mapper or ensure spans contain the necessary data.",
+                )
+            test_case = LLMTestCase(
+                input=result.input,
+                actual_output=result.actual_output,
+                expected_output=result.expected_output,
+                context=result.context,
+                retrieval_context=result.retrieval_context,
+            )
```
chianhw, line 90: why don't we have tools_called and expected_tools?
Do we want to support ToolCorrectnessMetric, ArgumentCorrectnessMetric in DeepEval?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/common.py
```diff
@@ -0,0 +1,100 @@
...
+
+@dataclass
+class SpanMapResult:
+    """Extraction result from spans — only what metrics consume."""
+
+    input: Optional[str] = None
+    actual_output: Optional[str] = None
+    retrieval_context: Optional[List[str]] = None
+    context: Optional[List[str]] = None
+    system_prompt: Optional[str] = None
```
chianhw, line 19: Are we using system_prompt for any metric?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/base.py
```diff
@@ -0,0 +1,51 @@
...
+    if result is not None:
+        if reference_inputs:
+            ref = reference_inputs[0]
+            expected = getattr(ref, "expected_response_text", None)
+            if expected:
+                result.expected_output = expected
+        return result
```
chianhw: We only take the data for `expected_output` here, but why don't we use the expected_trajectory field and assertions in reference_inputs?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/strands.py
```diff
@@ -0,0 +1,158 @@
...
+            return None
+
+        if system_prompt is None:
+            system_prompt = agent_span.get("attributes", {}).get("system_prompt")
+
+        return SpanMapResult(
+            input=input_text,
+            actual_output=actual_output,
+            retrieval_context=tool_outputs if tool_outputs else None,
+            context=tool_outputs if tool_outputs else None,
```
chianhw: Why do we use tool_output as context?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
stone-coding: fixed

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
stone-coding: fixed

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/common.py
stone-coding: Currently no DeepEval or Autoevals metric consumes system_prompt, LLMTestCase doesn't have this field. I include it in SpanMapResult for two reasons: (1) alignment with the server-side AgentInvocationSpan which also extracts it, and (2) forward-compatibility, customers using custom_mapper can access it from the spans if a future  metric needs it. Happy to remove if you feel it adds unnecessary surface area.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/strands.py
stone-coding: DeepEval uses context and retrieval_context for different metrics: HallucinationMetric requires context (ground-truth facts to check against), while FaithfulnessMetric requires  retrieval_context (retrieved documents to verify grounding). Tool outputs serve both purposes — they're what the agent retrieved AND the factual basis for its response. Setting  both ensures both metrics work without requiring a custom_mapper. Note: when assertions from reference_inputs are present, they override context (explicit ground truth takes 
 precedence over tool outputs). This matches the server-side pattern — the AgentInvocationSpan exposes tool results, and the service's built-in metrics use them as both retrieval context and grounding truth.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/base.py
stone-coding: updated. Now all three fields from reference_inputs are extracted:
 
  - expectedResponse.text → SpanMapResult.expected_output → LLMTestCase.expected_output
  - expectedTrajectory.toolNames → SpanMapResult.expected_tools → LLMTestCase.expected_tools (enables ToolCorrectnessMetric)
  - assertions[].text → SpanMapResult.assertions → LLMTestCase.context (enables HallucinationMetric to check agent response against ground-truth claims)
 For the Autoevals adapter, assertions are joined as a fallback for the expected kwarg when no explicit expectedResponse is provided.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
stone-coding: Added both. tools_called is now extracted from execute_tool spans (name + input_parameters + output) and mapped to LLMTestCase.tools_called. expected_tools comes from reference_inputs[0].expectedTrajectory.toolNames → LLMTestCase.expected_tools. This enables ToolCorrectnessMetric and ArgumentCorrectnessMetric.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/strands.py
stone-coding: Fixed, now takes the last agent span to match server-side OtelSpanMapper behavior. Applied to all three mappers (Strands, OpenInference LangChain, OpenTelemetry LangChain).

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/base.py
stone-coding: Applied to all three mappers (Strands, OpenInference LangChain, OpenTelemetry LangChain).

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/common.py
chianhw: No, I agree with you. We should keep it.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
stone-coding: Working on E2E test plan now. Considering how to strcuture parameterized Lambda handlers and how to obtains span from strands, openinference, opentelemetry, and customer mappers.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Raises:
+        ValueError: If no mapper can extract data from the spans.
+    """
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", category=DeprecationWarning, module="strands_evals")
+        mapper = _detect_mapper(session_spans)
+
+    try:
+        session = mapper.map_to_session(session_spans, session_id="eval")
```
chianhw: Why is session_id="eval" hardcoded?

StrandsInMemorySessionMapper uses it the filter spans.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
```diff
@@ -0,0 +1,98 @@
...
+            if result.retrieval_context:
+                kwargs["context"] = "\n".join(result.retrieval_context)
+
+        eval_result = self.metric.eval(**kwargs)
+
+        score = eval_result.score
+        label = None
+        if self.threshold is not None:
+            label = "Pass" if score is not None and score >= self.threshold else "Fail"
+        explanation = getattr(eval_result, "metadata", {}).get("rationale", "") if hasattr(eval_result, "metadata") else ""
```
chianhw: metadata could be None?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
```diff
@@ -0,0 +1,160 @@
...
+from strands_evals.types.trace import AgentInvocationSpan, Session, ToolExecutionSpan
+
+from bedrock_agentcore.evaluation.custom_code_based_evaluators.third_party.span_mappers.common import (
+    SpanMapResult,
+)
+
+logger = logging.getLogger(__name__)
+
+# Amazon ADOT distro scope not recognized by strands-evals detect_otel_mapper
+SCOPE_AMAZON_OTEL_LANGCHAIN = "amazon.opentelemetry.distro.instrumentation.langchain"
```
chianhw: Does the LangChainOtelSessionMapper in strands_evals support it? any reference?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
```diff
@@ -0,0 +1,160 @@
...
+            has_strands_scope = True
+        if get_body(span) is not None:
+            has_body_entry = True
+
+    # CloudWatch split format: Strands scope on metadata entries, body on log entries
+    if has_strands_scope and has_body_entry:
+        return CloudWatchSessionMapper()
+
+    # Fallback to strands-evals auto-detection
+    return detect_otel_mapper(session_spans)
```
chianhw: when we call `detect_otel_mapper(session_spans)`, do we still need to check the scope name before using it?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
stone-coding: Fixed, now extracts session.id from span attributes via _extract_session_id(). Falls back to "default" if no session.id found.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
stone-coding: Fixed, now handles metadata=None safely: metadata = getattr(eval_result, "metadata", None); explanation = metadata.get("rationale", "") if isinstance(metadata, dict) else ""

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
stone-coding: No — LangChainOtelSessionMapper only filters for opentelemetry.instrumentation.langchain scope. However, the Amazon ADOT distro scope (amazon.opentelemetry.distro.instrumentation.langchain) uses the same span format, so LangChainOtelSessionMapper works for it once routed correctly. Added comment explaining this is a workaround until strands-evals adds native support.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
stone-coding: Yes — detect_otel_mapper has two gaps: (1) doesn't recognize the Amazon ADOT distro scope, and (2) misidentifies CloudWatch split format as StrandsInMemorySessionMapper (which expects ReadableSpan objects, not dicts). Our pre-check handles both cases, then falls back for standard formats. Added docstring explaining why.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
chianhw: I don't think the spans with scope name `amazon.opentelemetry.distro.instrumentation.langchain` are exactly the same.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Autoevals adapter for AgentCore code-based evaluators."""
+
+from bedrock_agentcore.evaluation.custom_code_based_evaluators.third_party.autoevals.adapter import AutoevalsAdapter
+
+__all__ = ["AutoevalsAdapter"]
```
jariy17: nit: It should be AutoEvals

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
```diff
@@ -0,0 +1,99 @@
...
+
+        Args:
+            metric: An Autoevals scorer instance (e.g. Factuality(), ClosedQA()).
+            custom_mapper: Optional callable that receives the EvaluatorInput and
+                returns a dict of kwargs for metric.eval(). Bypasses default span
+                mapping when provided. Expected keys: input, output, expected (optional).
+            threshold: Optional score threshold for Pass/Fail label. If None, label
+                is omitted from the output.
+        """
+        self.metric = metric
```
jariy17, line 56: Can this support custom evaluators?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
```diff
@@ -0,0 +1,177 @@
...
+    from strands_evals.mappers.utils import get_body
+
+    has_strands_scope = False
+    has_body_entry = False
+
+    for span in session_spans:
+        scope_name = get_scope_name(span)
+        if scope_name == SCOPE_AMAZON_OTEL_LANGCHAIN:
+            return LangChainOtelSessionMapper()
+        if scope_name == "opentelemetry.instrumentation.langchain":
```
jariy17: nit: you can combine this and the above if statements together

tests_integ/evaluation/test_third_party_adapters.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def test_with_field_mapper(self):
+        from deepeval.metrics import BiasMetric
+
+        from bedrock_agentcore.evaluation.custom_code_based_evaluators.third_party.deepeval import DeepEvalAdapter
+
+        metric = BiasMetric(threshold=0.5)
+        adapter = DeepEvalAdapter(
+            metric=metric,
+            field_mapper=lambda ev: {
```
jariy17: does this exist? This should be custom_mapper.

tests_integ/evaluation/test_third_party_adapters.py
```diff
@@ -0,0 +1,171 @@
...
+        """Skip if autoevals is not installed."""
+        pytest.importorskip("autoevals")
+
+    def test_factuality_scorer(self):
+        from autoevals import Factuality
+
+        from bedrock_agentcore.evaluation.custom_code_based_evaluators.third_party.autoevals import AutoevalsAdapter
+
+        scorer = Factuality()
+        adapter = AutoevalsAdapter(scorer=scorer)
```
jariy17: I think this scorer is suppose to be metric.

tests_integ/evaluation/test_third_party_adapters.py
```diff
@@ -0,0 +1,171 @@
...
+        assert result.value is not None
+        assert result.label in ("Pass", "Fail")
+
+    def test_custom_threshold(self):
+        from autoevals import Factuality
+
+        from bedrock_agentcore.evaluation.custom_code_based_evaluators.third_party.autoevals import AutoevalsAdapter
+
+        scorer = Factuality()
+        adapter = AutoevalsAdapter(scorer=scorer, threshold=0.9)
```
jariy17: scorer should be metric

tests_integ/evaluation/test_third_party_adapters.py
```diff
@@ -0,0 +1,171 @@
...
+        assert result.value is not None
+
+    def test_with_field_mapper(self):
+        from autoevals import Factuality
+
+        from bedrock_agentcore.evaluation.custom_code_based_evaluators.third_party.autoevals import AutoevalsAdapter
+
+        scorer = Factuality()
+        adapter = AutoevalsAdapter(
+            scorer=scorer,
```
jariy17: scorer should be metric

tests_integ/evaluation/test_third_party_adapters.py
```diff
@@ -0,0 +1,171 @@
...
+            field_mapper=lambda ev: {
```
jariy17: field_mapper should be custom_mapper

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
```diff
@@ -0,0 +1,121 @@
...
+                if not result.actual_output:
+                    missing.append("actual_output")
+                metric_name = type(self.metric).__name__
+                return EvaluatorOutput(
+                    label="Error",
+                    errorCode="MISSING_REQUIRED_FIELD",
+                    errorMessage=f"Field(s) {missing} required by {metric_name} but not found in evaluation event. "
+                    f"Provide a custom_mapper or ensure spans contain the necessary data.",
+                )
+            # assertions from reference_inputs take precedence over span-extracted context
```
jariy17: It seems like context should be ground truth. Why are we setting it to tool call outputs? It should be None if result.assertions is null.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
```diff
@@ -0,0 +1,177 @@
...
+    Extracts the last AgentInvocationSpan for input/output and all
+    ToolExecutionSpans for retrieval_context and tools_called.
+    """
+    agent_span = None
+    tool_spans: List[ToolExecutionSpan] = []
+
+    for trace in session.traces:
+        for span in trace.spans:
+            if isinstance(span, AgentInvocationSpan):
+                agent_span = span
```
jariy17, line 195: This only keeps the last AgentInvocationSpan. This mean session-level evaluators like ConversationCompletenessEvaluator will evaluate the last turn only. 
Are we only scoping this project to trace/tool level evaluators only? If so, we need to document this and make sure customers are aware of this.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/base.py
```diff
@@ -0,0 +1,84 @@
...
+        Args:
+            evaluator_input: Parsed EvaluatorInput from the code-based evaluator flow.
+            context: Lambda context object (unused).
+
+        Returns:
+            EvaluatorOutput with score, label, and explanation or error fields.
+        """
+        try:
+            return self._run(evaluator_input)
+        except ValueError as e:
```
jariy17: Is every ValueError going to be a field-extraction error?  If not, can we scope this down? Either wrap only the map_spans() call in the try, or have registry.py raise a custom FieldExtractionError(ValueError) and catch that specifically.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
```diff
@@ -0,0 +1,121 @@
...
+                context=context,
+                retrieval_context=result.retrieval_context,
+                tools_called=self._build_tool_calls(result.tools_called) if result.tools_called else None,
+                expected_tools=self._build_tool_calls(result.expected_tools) if result.expected_tools else None,
+            )
+
+        try:
+            self.metric.measure(test_case)
+        except Exception as e:
+            if type(e).__name__ == "MissingTestCaseParamsError":
```
jariy17: `type(e).__name__ == "MissingTestCaseParamsError"` - we already hard-import deepeval at the top of this file, so just import the exception class and use isinstance. String-matching on class name is fragile.

 Also, pyproject.toml pins `deepeval>=2.0.0` with no upper bound. If DeepEval renames or moves this class in a future release, this check silently returns False, falls through to the base except Exception, and every MISSING_REQUIRED_FIELD becomes METRIC_ERROR.

jariy17: `type(e).__name__ == "MissingTestCaseParamsError"` - we already hard-import deepeval at the top of this file, so just import the exception class and use isinstance. String-matching on class name is fragile.

 Also, pyproject.toml pins `deepeval>=2.0.0` with no upper bound. If DeepEval renames or moves this class in a future release, this check silently returns False, falls through to the base except Exception, and every MISSING_REQUIRED_FIELD becomes METRIC_ERROR.

pyproject.toml
```diff
@@ -164,12 +164,18 @@ strands-agents = [
...
 simulation = [
     "jinja2>=3.1.0",
-    "strands-agents-evals>=0.1.0",
+    "strands-agents-evals>=1.0.0,<2.0.0",
 ]
 datasets = [
     "requests>=2.31.0",
 ]
+deepeval = [
+    "deepeval>=2.0.0",
```
jariy17, line 187: `registry.py` imports `strands_evals` but neither the deepeval nor autoevals extras include `strands-agents-evals`. A customer following the advertised install path (pip install bedrock-agentcore[deepeval]) gets an ImportError at cold start. `Add strands-agents-evals>=1.0.0,<2.0.0` to both extras.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/common.py
```diff
@@ -0,0 +1,51 @@
...
+    expected_output: Optional[str] = None
+    tools_called: Optional[List[Dict[str, Any]]] = None
+    expected_tools: Optional[List[Dict[str, Any]]] = None
+    assertions: Optional[List[str]] = None
+
+    def to_dict(self) -> Dict[str, Any]:
```
jariy17: is this being used? Please remove if it's not

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/common.py
```diff
@@ -0,0 +1,51 @@
...
+        if self.tools_called is not None:
+            result["tools_called"] = self.tools_called
+        if self.expected_tools is not None:
+            result["expected_tools"] = self.expected_tools
+        if self.assertions is not None:
+            result["assertions"] = self.assertions
+        return result
+
+
+def _try_parse_json(val: str) -> Any:
```
jariy17: is this dead code? PLease remove

jariy17 (changes requested):
Lefts some comments but its a good addition to our SDK.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/__init__.py
stone-coding: Fixed — renamed to AutoEvalsAdapter throughout.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/adapter.py
stone-coding: Yes, metric accepts any object with an .eval(output, expected, **kwargs) method. Customers can pass their own scorer class alongside the built-in ones.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
stone-coding: Fixed, combined into if scope_name in (SCOPE_AMAZON_OTEL_LANGCHAIN, "opentelemetry.instrumentation.langchain").

tests_integ/evaluation/test_third_party_adapters.py
stone-coding: Fixed, renamed to custom_mapper and updated return type to LLMTestCase.

tests_integ/evaluation/test_third_party_adapters.py
stone-coding: Fixed, rename to metric

tests_integ/evaluation/test_third_party_adapters.py
stone-coding: Fixed, rename to metric

tests_integ/evaluation/test_third_party_adapters.py
stone-coding: Fixed, rename to metric

tests_integ/evaluation/test_third_party_adapters.py
stone-coding: Fixed.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
stone-coding: You are right, fixed. Now context = result.assertions if result.assertions else None. Tool outputs stay in retrieval_context only.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
stone-coding: Correct, currently scoped to trace/tool level evaluators only. ConversationalTestCase (multi-turn) is documented as future work. I will add a docstring/comment noting this limitation so customers are aware.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/base.py
stone-coding: Fixed, created FieldExtractionError(ValueError) in common.py. Registry raises it, base.py catches only that. Other ValueErrors now fall through to the general exception handler.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
stone-coding: Fixed, imported MissingTestCaseParamsError from deepeval.errors and using except MissingTestCaseParamsError directly. No more string matching.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/common.py
stone-coding: removed.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/common.py
stone-coding: removed.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/deepeval/adapter.py
stone-coding: Fixed, imported MissingTestCaseParamsError from deepeval.errors and using except MissingTestCaseParamsError directly. No more string matching.

pyproject.toml
stone-coding: Fixed, added strands-agents-evals>=1.0.0,<2.0.0 to both deepeval and autoevals extras.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
chianhw: The original design doc includes `ConversationalTestCase`. Can you help me understand the rationale behind excluding it from the project scope?

pyproject.toml
```diff
@@ -164,12 +164,20 @@ strands-agents = [
...
     "jinja2>=3.1.0",
-    "strands-agents-evals>=0.1.0",
+    "strands-agents-evals>=1.0.3,<2.0.0",
 ]
 datasets = [
     "requests>=2.31.0",
 ]
+deepeval = [
+    "deepeval>=2.0.0",
+    "strands-agents-evals>=1.0.3,<2.0.0",
```
jariy17, line 188: Please regenerate uv.lock after this dependency change. It still resolves strands-agents-evals==0.1.0, which does not provide detect_otel_mapper. Frozen installs will fail at import time.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
```diff
@@ -0,0 +1,148 @@
...
+    except Exception as e:
+        raise FieldExtractionError(
+            f"Could not extract evaluation fields from spans using {type(mapper).__name__}: {e}. "
+            f"Provide a custom_mapper for custom or unsupported span formats."
+        ) from e
+
+    result = _session_to_span_map_result(session)
+
+    if reference_inputs:
+        ref = reference_inputs[0]
```
jariy17: When you use reference_inputs[0], you are only reading one scoped reference. A TRACE evaluation can contain both a session-level reference and a matching trace-level reference, so this can drop either expectedResponse or the session-level assertions and trajectory. Please inspect and combine the relevant entries instead of selecting one by position.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
```diff
@@ -0,0 +1,226 @@
...
+        if service_result:
+            result = service_result
+        elif result is None:
+            raise FieldExtractionError(
+                "No AgentInvocationSpan found in session and service format extraction failed. "
+                "Provide a custom_mapper for custom or unsupported span formats."
+            )
+
+    if reference_inputs:
+        ref = reference_inputs[0]
```
chianhw: > Attributes:
        context: Span context for the entry, e.g. {"spanContext": {"sessionId", "traceId"}}.

Why do we use the first reference_input? If all the reference_inputs have an empty context or an empty spanContext, using the first one as the default makes sense. But what if the sessionId/traceId has been explicitly specified?
reference for SpanContext : https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_SpanContext.html
```
    "evaluationReferenceInputs": [{
            "context": {
                "spanContext": {
                    "sessionId": "a9f5f0c7-3c94-43cc-933d-7e191caf9d8d"
                }
            },
            "assertions": [{
                    "text": "search_flights tool call args are origin=SEA and destination=NYC"
                },
                {
                    "text": "search_hotels is called with city=NYC"
                }
            ],
            "expectedTrajectory": {
                "toolNames": ["search_flights", "book_flight", "search_hotels", "book_hotel"]
            }
        }, {
            "context": {
                "spanContext": {
                    "traceId": "69b0693f0ed8ff777d364fe4265fe2a4",
                    "sessionId": "a9f5f0c7-3c94-43cc-933d-7e191caf9d8d"
                }
            },
            "expectedResponse": {
                "text": "Booked flight DL420"
            }
        },
        {
            "context": {
                "spanContext": {
                    "traceId": "69b0696718e27bea18ac5a74148a81d8",
                    "sessionId": "a9f5f0c7-3c94-43cc-933d-7e191caf9d8d"
                }
            },
            "expectedResponse": {
                "text": "Booked the plaza hotel for $1350"
            }
        }
    ],
```

tests_integ/evaluation/test_third_party_adapters.py
```diff
@@ -0,0 +1,172 @@
...
+        session_spans=spans,
+        target_trace_id="integ-trace-1",
+    )
+
+
+class TestDeepEvalAdapterIntegration:
+    """Integration tests for DeepEvalAdapter with real DeepEval metrics."""
+
+    @pytest.fixture(autouse=True)
+    def check_deepeval(self):
```
jariy17, line 57: I would remove this skip. Otherwise, CI can pass without making it clear that the test never ran.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/base.py
```diff
@@ -0,0 +1,87 @@
...
+        - SESSION: all spans (no filtering) — one call for the entire session
+        - TRACE: only spans matching target_trace_id (service sends exactly one)
+        - TOOL_CALL: only the span matching target_span_id (service sends exactly one)
+        """
+        spans = evaluator_input.session_spans
+
+        if evaluator_input.evaluation_level == "TRACE" and evaluator_input.target_trace_id:
+            spans = [s for s in spans if s.get("traceId") == evaluator_input.target_trace_id]
+        elif evaluator_input.evaluation_level == "TOOL_CALL" and evaluator_input.target_span_id:
+            spans = [s for s in spans if s.get("spanId") == evaluator_input.target_span_id]
```
jariy17: Filtering a TOOL_CALL evaluation to only the target span drops the enclosing AgentInvocationSpan. DeepEval's Tool Correctness documentation lists input, actual_output, tools_called, and expected_tools as required arguments. The agent span provides input and actual_output, while the targeted tool span provides the data for tools_called.

For example, suppose a trace contains an invoke_agent span with the input "What's the weather?" and output "72 F", plus an execute_tool span for get_weather. When target_span_id points to the tool span, this filter removes the agent span. map_spans() then raises No AgentInvocationSpan found in session, and the adapter returns FIELD_EXTRACTION_ERROR instead of running ToolCorrectnessMetric.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/models.py
```diff
@@ -80,8 +80,9 @@ class EvaluatorOutput(BaseModel):
 
     @model_validator(mode="after")
     def _require_label_or_error_code(self) -> "EvaluatorOutput":
-        if not self.errorCode and self.label is None:
+        if not self.errorCode and self.label is None and self.value is None:
```
jariy17: Successful AgentCore code-based evaluator responses cannot omit label. This change permits EvaluatorOutput(value=0.75, label=None), which is the default path in AutoEvalsAdapter when no threshold is provided.

I verified this against the live service. The Lambda completed successfully and returned:

```
{"value": 0.75, "explanation": "test"}
```


AgentCore rejected it with:

```
InvalidLambdaResponse: LambdaEvaluationSuccessResponse.label: Field required
```


The same Lambda succeeded when it included "label": "Pass". Please restore the existing label validation and make AutoEvalsAdapter always produce a label, such as by requiring a threshold or restoring the 0.5 default. The existing test_label_required_without_error_code also currently fails because this invalid response is accepted by the SDK model.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""AutoEvals adapter for AgentCore code-based evaluators."""
```
jariy17, line 1: The branch does not pass the repository's required lint and formatting checks. Using the exact Ruff 0.12.0 version pinned by CI, ruff check . reports 10 errors (F401, G201, E501, and F841), and ruff format --check . reports four files that require formatting. The CI workflow runs pre-commit run --all-files, so the lint job will fail as submitted. Please run pre-commit run --all-files, address the remaining errors, and rerun it until the branch is clean.

jariy17 (changes requested):

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
stone-coding: Fixed — now matches reference_input to the target trace by spanContext.traceId. If a reference input has a matching traceId, it's used. If no traceId is specified (empty context), it's treated as the default. Falls back to first entry if nothing matches.

jariy17 (dismissed):

jariy17 (dismissed):

jariy17 (approved):

PR #570, opened 2026-07-07
fix(payments): address langgraph middleware review follow-ups

aidandaly24:
## Description of changes

Follow-up fixes to the LangGraph payments middleware merged in #546. These came out of a post-merge review of `middleware.py`. All are scoped to the LangGraph integration; no public API changes.

### Correctness

1. **Async `on_payment_error` callback on the sync path now fails loudly.**
   The guard that detects an async callback used with sync `.invoke()` previously raised a `TypeError` *inside* the callback `try`, whose `except Exception` immediately caught it, logged it, and returned `None` (PROPAGATE). The developer's misconfiguration was swallowed into a log line and an error message indistinguishable from a real payment failure. The check now runs before the try and propagates, matching the "fail loudly" intent stated in the original commit and class docstring.

2. **Async callable objects (`async def __call__`) are now detected.**
   `inspect.iscoroutinefunction` returns `False` for a callable instance whose `__call__` is async, so such a callback slipped past the guard on the sync path (leaked an un-awaited coroutine, then degraded to PROPAGATE) and was never awaited on the async path. Detection now inspects the bound `__call__`; the sync path raises and the async path awaits it.

3. **Post-payment retry-rejection honors custom handlers.**
   `_check_retry_rejection` hardcoded `GenericPaymentHandler` + JSON fallback, so when a tool with a registered custom handler returned a bespoke-format 402 *after* payment (server rejected the signed payment), the still-402 went undetected and was handed back to the LLM as a normal result. It now routes through the custom handler (raw content), mirroring `_detect_402`, and surfaces the real error detail.

4. **Payment header injection preserved when a tool call omits `args`.**
   `_detect_402` used `request.tool_call.get("args", {})`, which returns a throwaway dict when there is no `args` key, so header injection mutated a dict the retried handler never read. Switched to `setdefault("args", {})` so the header reaches the retried call.

### Docs

5. **Documented the custom-handler input contract.**
   Custom handlers receive the raw `ToolMessage.content` (str or list of blocks), not the middleware's internal wrapped shape. The built-in handlers expect the normalized shape, so passing one of them directly as a `custom_handlers` value silently fails to detect 402s. Clarified this in the config docstring and README. This is docs-only on purpose: a type-based config guard would wrongly reject legitimate `PaymentResponseHandler`/`GenericPaymentHandler` subclasses (the existing `TrackingHandler` functional test is exactly that).

### Behavior change to flag for reviewers

Fix #1 changes async-callback-on-sync from "log + PROPAGATE" to "raise `TypeError`", and updates the two tests that asserted the swallowed behavior. If the graceful fallthrough was intentional, the alternative is to keep it and only correct the misleading "fail loudly" wording instead. Happy to switch.

### Deferred (noted, not addressed here)

Non-correctness cleanups left for a separate change: the `"PROPAGATE"` magic-string sentinel in `_handle_callback_resolution`, the remaining near-duplication between `_inject_payment_header` and `_inject_for_error_retry`, the redundant name-based detection fallback for generic tools, and `_FallbackHandler` not subclassing `PaymentResponseHandler`.

## Testing

- Added regression tests for all four correctness fixes; each was verified to **fail against the merged `middleware.py` and pass with these changes** (temporarily swapped the merged file back in and ran the new tests: 6 failed / 1 guard-passed, then all pass once restored).
- Full LangGraph unit suite green: `146 passed` (up from 141).
- Full payments suite green: `648 passed`.
- `ruff check` and `ruff format --check` clean on all changed files.

### DIFF
--- a/src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
+++ b/src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
@@ -1,6 +1,7 @@
 """AgentCorePaymentsMiddleware for LangGraph agents."""
 
 import asyncio
+import functools
 import inspect
 import json
 import logging
@@ -209,9 +210,7 @@ class AgentCorePaymentsMiddleware(AgentMiddleware):
     def _detect_402(self, request: ToolCallRequest, result: ToolMessage) -> Optional[_DetectionResult]:
         """Run 402 detection. Returns detection context if 402 found, None otherwise."""
         tool_name = request.tool_call["name"]
-        # setdefault (not get): when a tool call arrives without an "args" key, write the
-        # empty dict back into request.tool_call so a later payment-header injection is
-        # visible to the retried handler instead of mutating a throwaway dict.
+        # Store args back on the tool call so an injected payment header reaches the retried handler.
         tool_args = request.tool_call.setdefault("args", {})
         prepared = self._prepare_for_handler(result.content)
         if prepared is None:
@@ -402,15 +401,19 @@ class AgentCorePaymentsMiddleware(AgentMiddleware):
 
     @staticmethod
     def _is_async_callback(callback: Any) -> bool:
-        """True if the callback is a coroutine function or a callable object with async __call__.
+        """True if the callback ultimately resolves to an async coroutine function.
 
-        ``inspect.iscoroutinefunction`` alone misses stateful callbacks implemented as an
-        object with an ``async def __call__``, so we inspect the bound ``__call__`` too.
-        (``callable()`` — ruff B004's suggestion — cannot distinguish sync from async.)
+        Covers the wrappings a plain ``inspect.iscoroutinefunction`` can miss: callable
+        objects with an ``async def __call__``, ``functools.partial`` (whose async-ness is
+        not reliably visible through the partial across Python versions), and partials of
+        either. (``callable()`` — ruff B004's suggestion — cannot distinguish sync from async.)
         """
-        if inspect.iscoroutinefunction(callback):
+        target = callback
+        while isinstance(target, functools.partial):
+            target = target.func
+        if inspect.iscoroutinefunction(target):
             return True
-        call = getattr(callback, "__call__", None)  # noqa: B004
+        call = getattr(target, "__call__", None)  # noqa: B004
         return inspect.iscoroutinefunction(call)
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -209,7 +209,10 @@ def _check_guards(
     def _detect_402(self, request: ToolCallRequest, result: ToolMessage) -> Optional[_DetectionResult]:
         """Run 402 detection. Returns detection context if 402 found, None otherwise."""
         tool_name = request.tool_call["name"]
-        tool_args = request.tool_call.get("args", {})
+        # setdefault (not get): when a tool call arrives without an "args" key, write the
```
jariy17: nit: remove comments. I would follow this mantra: "Only write comments for why we are doing this not how"

jariy17: For example, "We set a empty dictionary so the payment-header injection is visible to the retried handler"

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
```diff
@@ -387,6 +400,19 @@ def _create_auto_session(self) -> None:
         self.config.payment_session_id = session["paymentSessionId"]
         logger.info("auto_session: created session %s", self.config.payment_session_id)
 
+    @staticmethod
+    def _is_async_callback(callback: Any) -> bool:
```
jariy17, line 403: Nice catch on the __call__ inspection — callable() can't distinguish sync from async, so the two-step check is the right approach. Docstring explains the rationale well.
One thing to consider: inspect.iscoroutinefunction won't see through functools.partial on Python < 3.12. If we ever pass partial-wrapped async callbacks (e.g. functools.partial(some_async_handler, tenant_id=...)), this silently returns False and we'd schedule it on the wrong loop/executor.

jariy17 (changes requested):
Just need to catch all types of async calls.

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
aidandaly24: Trimmed to a single why-only line in 9b9fc88:
`# Store args back on the tool call so an injected payment header reaches the retried handler.`

src/bedrock_agentcore/payments/integrations/langgraph/middleware.py
aidandaly24: Checked this empirically on 3.10.20: `inspect.iscoroutinefunction(functools.partial(async_fn))` returns `True` there, so a partial-wrapped async *function* is already caught. The gap that does exist even on 3.10 is a partial wrapping an async *callable object* (`async def __call__`), which returns `False`. Rather than depend on per-version partial behavior, `_is_async_callback` now unwraps `functools.partial` before inspecting, covering `partial(async_fn)`, `partial(async_object)`, and nested partials. Regression tests added for the sync (raise) and async (await) partial paths in 9b9fc88.

jariy17 (approved):

PR #581, opened 2026-07-17
fix: tighten package specifier validation in install_packages()

notgitika:
## Summary

Tightens package specifier validation in `install_packages()`:

- Restricts the extras group in the package-name pattern to valid extra names (comma-separated identifiers) instead of allowing arbitrary characters.
- Quotes each package argument with `shlex.quote()` when building the `pip install` command so each is passed as a single literal token.

## Tests

- Adds a test that extras groups with characters outside valid extra names are rejected.
- Adds a test confirming valid extras (e.g. `pandas[security]`, `celery[redis,auth]`) still work.
- Updates the version-specifier test for the quoting change.

### DIFF
--- a/src/bedrock_agentcore/tools/code_interpreter_client.py
+++ b/src/bedrock_agentcore/tools/code_interpreter_client.py
@@ -6,6 +6,7 @@ applications to start, stop, and invoke code execution in a managed sandbox envi
 
 import logging
 import re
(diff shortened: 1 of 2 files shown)


notgitika:
Breaking change expected. We changed the regex

padmak30 (approved):

PR #584, opened 2026-07-21
Add LLM issue auto-triage: type and high-severity labelers

jesseturner21:
## What

Adds automatic issue triage on `issues: opened` using [strands-agents/devtools/issue-labeler](https://github.com/strands-agents/devtools/tree/main/issue-labeler) (the same setup strands-agents runs on their SDK repos), pinned by commit SHA.

Two parallel jobs:
- **Type** (`.github/labelers/type.yml`): applies exactly one of `bug` / `enhancement` / `question`.
- **Severity** (`.github/labelers/severity.yml`): applies `high-severity` or nothing. Criteria are deliberately conservative — production outage, data loss, security vulnerability, crash with no workaround, or regression in a released version, and only with concrete evidence. Feature requests, usage questions, and vague reports never qualify. Body truncation is raised to 4000 chars for this job so stack traces/repro steps are visible to the classifier.

The `high-severity` label has been created on the repo.

## Security model

The classifier LLM has **no tools, no shell, and no GitHub access**. It returns a structured object whose label field is an enum built from the config allowlist, so out-of-allowlist labels are rejected by the schema. Issue content is sanitized and truncated before reaching Bedrock. IAM is scoped to `bedrock:InvokeModel*` via an inline session policy; the workflow needs only `issues: write`. Worst case from prompt injection in an issue body is a mislabel.

## Notes for reviewers

- Reuses `secrets.BEDROCK_SECURITY_REVIEW_ROLE_ARN` (same OIDC role as the PR security review). **If that role's trust policy pins the `sub` claim to pull-request refs, it will reject `issues`-triggered runs** — needs a check before merge, or a dedicated minimal role.
- The action is pinned to `5e50593` (strands devtools main as of 2026-07-17); bump deliberately.
- Possible follow-up: Slack notification when `high-severity` lands, via the existing `github-slack-notifications.yml` webhook.

### DIFF
--- a/.github/workflows/issue-auto-triage.yml
+++ b/.github/workflows/issue-auto-triage.yml
@@ -1,0 +1,55 @@
+name: Issue Auto Triage
+
+# Classifies new issues with an LLM (Bedrock via strands-agents/devtools
+# issue-labeler) and applies labels from a hardcoded allowlist. The model has
+# no tools, no shell, and no GitHub access — it returns a structured object
+# whose label field is an enum built from the config allowlist, so the worst
+# a prompt injection in an issue body can achieve is a mislabel.
+#
+# Labeler configs live in .github/labelers/. Each job is an independent
+# classification concern and they run in parallel.
+
+on:
+  issues:
+    types: [opened]
+
+permissions:
+  issues: write
+  id-token: write
+  contents: read
+
+jobs:
+  label-type:
+    name: "Label: Type"
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: .github/labelers
+          sparse-checkout-cone-mode: false
+      # strands-agents/devtools/issue-labeler pinned to a reviewed commit
+      # (main as of 2026-07-17). Bump deliberately, not automatically.
+      - uses: strands-agents/devtools/issue-labeler@5e50593798905abfbb31cc709f5ac58a2992b03a
+        with:
+          aws_role_arn: ${{ secrets.BEDROCK_SECURITY_REVIEW_ROLE_ARN }}
+          config_path: '.github/labelers/type.yml'
+          max_labels: '1'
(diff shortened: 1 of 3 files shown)


.github/workflows/issue-auto-triage.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: .github/labelers
+          sparse-checkout-cone-mode: false
+      # strands-agents/devtools/issue-labeler pinned to a reviewed commit
+      # (main as of 2026-07-17). Bump deliberately, not automatically.
+      - uses: strands-agents/devtools/issue-labeler@5e50593798905abfbb31cc709f5ac58a2992b03a
```
tejaskash, line 33: Why are we pinning this?

.github/workflows/issue-auto-triage.yml
```diff
@@ -0,0 +1,55 @@
...
+  label-severity:
+    name: "Label: Severity"
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: .github/labelers
+          sparse-checkout-cone-mode: false
+      - uses: strands-agents/devtools/issue-labeler@5e50593798905abfbb31cc709f5ac58a2992b03a
```
tejaskash, line 48: Same as above

tejaskash (changes requested):
Looks good, but not sure why we're pinning the strands reviewer to a certain commit.

.github/workflows/issue-auto-triage.yml
jesseturner21: Without this there could be a breaking change. By pinning to the commit we protect against that. A sort of "Versioning" for this

tejaskash (approved):

Issue #187, opened 2025-12-10
[Feature Request] Support for Claude Agent SDK


austinmw:
**Is your feature request related to a problem? Please describe.**
It would be cool to add support for the Claude Agent SDK

**Describe the solution you'd like**
Allow wrapping [claude-agent-sdk-python](https://github.com/anthropics/claude-agent-sdk-python)

PR #588, opened 2026-07-21
feat(memory): add AgentCoreMemoryStore Strands integration

### LINKED ISSUE #586: Port AgentCoreMemoryStore to Python
**Is your feature request related to a problem? Please describe.**
Support AgentCoreMemoryStore for strands in python.

**Describe the solution you'd like**
AgentCoreMemoryStore for strands is currently built in TS.

**Additional context**
Loooking to port this over to Python so customers have feature parity in Strands.

strandly-the-agent:
> **TL;DR:** Ports the TypeScript `AgentCoreMemoryStore` Strands integration to Python as a contained package. Revision `3de6915` is on current `main` (which clears the a2a breaking-change report) and the only client callers can pass is this SDK's `MemoryClient` — the scoped-down `AgentCoreDataPlaneClient` Protocol and `AgentCoreClient` union are gone.

## Review responses

| Ask | Response |
|---|---|
| Breaking change: `A2A_CONTRACT_PORT` / `A2A_PORT_ENV` "removed" | **Stale-branch artifact, not a removal.** Both were *added on `main`* by #615 (`207adb7`) after this branch's merge-base, so griffe (which diffs `main` → PR tree) saw them missing from the older tree. This PR never touched `src/bedrock_agentcore/runtime/a2a.py`. Merged current `main`; griffe now reports **0 breaking changes**. |
| Drop `AgentCoreDataPlaneClient` / `AgentCoreClient`; only accept `MemoryClient` | Done. Both names and `resolve_data_plane_client()` are deleted. `AgentCoreMemoryStore`, `create_agentcore_memory_stores` and `AgentCoreEventSender` now take `client: MemoryClient`, and the default is a `MemoryClient(integration_source="strands")`. They call the `bedrock-agentcore` data plane it vends, because `MemoryClient.create_event` cannot carry the `clientToken` the sender needs for idempotent retries and `MemoryClient.retrieve_memories` swallows errors `MemoryManager` expects to see. |
| Fold the message helpers into `AgentCoreEventSender` ([r3707303752](https://github.com/aws/bedrock-agentcore-sdk-python/pull/588#discussion_r3707303752)) | `_format.py` deleted; `_map_role` / `_extract_text` / `_is_user_or_assistant_with_text` are private members of the sender. |
| `_assert_writable_topology` ([r3708398590](https://github.com/aws/bedrock-agentcore-sdk-python/pull/588#discussion_r3708398590)) | Renamed and removed from the package `__all__`. |
| Integration-test timeout ([r3660049235](https://github.com/aws/bedrock-agentcore-sdk-python/pull/588#discussion_r3660049235)) | Hunk dropped; `.github/workflows/integration-testing.yml` is byte-identical to `main`. |

Source: TypeScript PR [#187](https://github.com/aws/bedrock-agentcore-sdk-typescript/pull/187) at `c7cb2bca47258108771e87966901d651aee3cef8` · Target: `3de6915` · Closes #586

<details>
<summary><strong>Scope: what this PR changes</strong></summary>

```text
pyproject.toml                                    (strands-agents floor >= 1.46.0, boto3/botocore >= 1.43.35)
uv.lock
src/.../strands/memorystore/{__init__,factory,sender,store,types}.py, README.md   (new)
tests/.../strands/memorystore/**                                                   (new)
tests_integ/memory/integrations/test_memory_store.py                               (new)
```

No existing module, test, doc, or workflow is modified. `griffe` reports **0 breaking changes** vs `main`, and a runtime probe of the pre-existing Strands modules (`dir()`, `__all__`, `logger.name`, owning module per public symbol) matches a `main` worktree — customer import paths, patch targets and logger names are untouched.
</details>

<details>
<summary><strong>Client contract after this revision</strong></summary>

```python
from bedrock_agentcore.memory import MemoryClient
from bedrock_agentcore.memory.integrations.strands.memorystore import AgentCoreMemoryStore

# default: the integration builds MemoryClient(integration_source="strands") itself
store = AgentCoreMemoryStore(memory_id=..., actor_id=..., session_id=..., namespace="/facts/{actorId}")

# or reuse your own connection
store = AgentCoreMemoryStore(..., client=MemoryClient(region_name="us-west-2"))
```

`create_agentcore_memory_stores(..., client=MemoryClient(...))` shares one client across the whole topology. Nothing else is accepted — there is no Protocol or union in the public surface any more.
</details>

<details>
<summary><strong>Behavior traceability matrix (TypeScript test → Python test)</strong></summary>

Source paths are relative to `src/memory/integrations/strands/__tests__/`; target paths to `tests/bedrock_agentcore/memory/integrations/strands/memorystore/`.

| Source test (`file:line`) | Behavior | Target test (`file:line`) |
|---|---|---|
| `format.test.ts:13,16` | `mapRole` user/assistant → `USER`/`ASSISTANT` | `test_message_formatting.py:19` |
| `format.test.ts:22,25,28` | join text blocks, ignore non-text, empty when none | `test_message_formatting.py:24`, `:54` |
| `format.test.ts:31` | trim surrounding whitespace | `test_message_formatting.py:40` |
| `format.test.ts:37,40,43,46` | only user/assistant with extractable text | `test_message_formatting.py:68` |
| `sender.test.ts:57` | whole batch → one `createEvent`, role-tagged, ordered | `test_sender.py:68` |
| `sender.test.ts:70` | split at `maxTurnsPerEvent` | `test_sender.py:82` |
| `sender.test.ts:79,238` | skip tool-only/empty; no-op batch | `test_sender.py:95` |
| `sender.test.ts:86,136` | omit `clientToken` without complete sequence numbers | `test_sender.py:116` |
| `sender.test.ts:92,105` | deterministic per-chunk seq-range token, stable across a re-fire | `test_sender.py:128` |
| `sender.test.ts:111,122` | token anchored on run id; default fresh UUID | `test_sender.py:144`, `:152` |
| `sender.test.ts:142,153` | new event when metadata changes; consecutive-only grouping | `test_sender.py:169` |
| `sender.test.ts:166,198` | map constant metadata to `stringValue`; omit empty bag | `test_sender.py:190`, `:423` |
| `sender.test.ts:179` | charset violation raises before any `createEvent` | `test_sender.py:214`, `:232`, `:240` |
| `sender.test.ts:188` | nullish/non-finite metadata rejected | `test_sender.py:249`, `:360` |
| `sender.test.ts:205,213,222` | attempt every event, then aggregate error folding the first reason | `test_sender.py:261`, `:283` |
| `sender.test.ts:244` | reject non-positive `maxTurnsPerEvent` | `test_sender.py:293` |
| `sender.test.ts:248,255` | `extractionMode` passthrough only when configured | `test_sender.py:305` |
| `store.test.ts:67,401` | resolve `{actorId}`/`{sessionId}`, query exact `namespace` | `test_store.py:65` |
| `store.test.ts:75` | subtree mode uses `namespacePath` | `test_store.py:74` |
| `store.test.ts:83` | `MemoryRecordSummary` → `MemoryEntry` mapping | `test_store.py:88` |
| `store.test.ts:105,137` | `topK == want` without a score floor | `test_store.py:114`, `:402` |
| `store.test.ts:112` | over-fetch, filter by floor, trim to want | `test_store.py:121` |
| `store.test.ts:128,144,155` | custom factor, ceil to int, cap at 100 | `test_store.py:141`, `:412` |
| `store.test.ts:172` | unscored records treated as 0 under a floor | `test_store.py:162` |
| `store.test.ts:179,195,201` | non-text → empty, `[]` on empty, errors propagate | `test_store.py:169`, `:177`, `:182` |
| `store.test.ts:211,225,237` | writable store forwards messages + sequence numbers, batched | `test_store.py:190` |
| `store.test.ts:251` | non-writable store rejects `addMessages` | `test_store.py:213` |
| `store.test.ts:259,268,277,309` | extraction only on the writer; warn only when dropped | `test_store.py:219`, `:232` |
| `store.test.ts:285,299` | self-name from namespace; slug fallback for a blank name | `test_store.py:240`, `:260` |
| `store.test.ts:292` | `writable` defaults to false | `test_store.py:275` |
| `store.test.ts:317,341` | stands alone with flat identity; identity validation | `test_store.py:288`, `:309` |
| `store.test.ts:369,373,380` | empty namespace / unresolved placeholder (exact + subtree) | `test_store.py:325`, `:346` |
| `store.test.ts:395` | `$`-sequences in `{actorId}` inserted verbatim | `test_store.py:352`, `:389` |
| `factory.test.ts:29,34,41` | one store per namespace, one default writer, extraction on writer | `test_factory.py:50` |
| `factory.test.ts:47,54` | custom cadence/filter reach the writer | `test_factory.py:61` |
| `factory.test.ts:61,76` | explicit `writable` flag / opt-out selection | `test_factory.py:70`, `:87` |
| `factory.test.ts:91,105` | all opted out, or two writers → error | `test_factory.py:104`, `:118` |
| `factory.test.ts:118,126,131` | recall-only modes; `extraction: true` passthrough | `test_factory.py:132`, `:142` |
| `factory.test.ts:140,152` | derived/explicit names, default fallback | `test_factory.py:148` |
| `factory.test.ts:162,166,180` | namespace validation | `test_factory.py:174`, `:181`, `:199` |
| `factory.test.ts:188` | actor/session bound per call | `test_factory.py:257` |
| `factory.test.ts:208,212,216,222,226` | writable-topology assertions | `test_factory.py:227`, `:233`, `:241` |
| `tests_integ/memory.test.ts` (live) | idempotent batched writes, extraction modes, exact/subtree recall, namespace isolation, real-agent round trip | `tests_integ/memory/integrations/test_memory_store.py:82`, `:129`, `:177`, `:193`, `:228`, `:270` — **not executed** (needs live AWS) |

**Deliberate divergences from the source**, both requested in review: the `topK` clamp on every retrieval path (`test_store.py:412`, `:428`, `:435`, `:446`) and Python-native `str.strip()`/`json.dumps` metadata handling instead of ECMAScript emulation (`test_message_formatting.py:40`, `test_sender.py:383`, `:408`).

Python-only tests cover Python-runtime concerns rather than new product behavior: `MemoryClient` construction/acceptance (`test_store.py:465`, `:550`, `test_factory.py:270`), region/session resolution (`test_store.py:482`, `:510`, `:525`, `:586`), blocking boto3 call kept off the event loop (`test_sender.py:316`), cancellation settlement (`test_sender.py:432`, `:454`), `TypedDict` runtime metadata (`test_types.py:18`), package exports (`test_package_exports.py`), and static typing fixtures (`test_static_typing.py`).
</details>

<details>
<summary><strong>Validation at <code>3de6915</code></strong></summary>

```text
pytest tests/bedrock_agentcore/memory -q                       → 774 passed
pytest tests/bedrock_agentcore/memory/.../memorystore -q       → 163 passed
pytest tests/unit tests/integration -q                         → 372 passed, 1 skipped
pytest tests/bedrock_agentcore/{runtime,evaluation,gateway,services,tools,config_bundle} -q
                                                               → 1134 passed, 4 xpassed, 2 failed*
ruff check . / ruff format --check .                            → clean
mypy src/.../strands/memorystore                                → Success: no issues found in 5 source files
griffe find_breaking_changes vs origin/main                     → 0
pytest --collect-only tests_integ/memory/integrations           → 18 collected
git diff --check                                                → clean
```

\* The two failures are `runtime/test_a2a.py::TestBuildA2AApp::test_auto_builds_card_when_none_provided` and `::TestServeA2A::test_explicit_port_overrides_environment`, reproduced identically at the merged base with this PR's commit stashed — environmental in my sandbox (a2a-sdk version), not caused by this branch. `tests/bedrock_agentcore/identity` also attempts a live `bedrock-agentcore` call here and hangs. CI is the authority for both. Live AWS memory tests need `MEMORY_PREPOPULATED_ID` plus Bedrock access.

An independent review pass ran over the client-contract commit; its finding (a now-duplicated mypy fixture block) is folded in.
</details>

<details>
<summary><strong>Open items</strong></summary>

- **Retrieval pagination** (`maxResults`/`nextToken`) stays a fast follow, per `arielnabavian` on the `store.py` thread; a request above 100 records returns the first 100.
- **Full `CreateEvent` request preflight** (payload/text/metadata service bounds) also matches the source and is in the same fast-follow bucket.
- **CI wiring for the live suite** is not in this PR: the memory matrix group would need `tests_integ/memory/integrations/test_memory_store.py`, `pytest-asyncio`, and a timeout that accommodates eventual-consistency polling.
</details>


arielnabavian:
Meant to set this as draft, doing some work to validate parity, ignore for now.

src/bedrock_agentcore/memory/integrations/strands/memorystore/factory.py
```diff
@@ -0,0 +1,145 @@
+"""Factory helpers for multi-namespace AgentCore memory topologies."""
```
arielnabavian, line 1: port lgtm

src/bedrock_agentcore/memory/integrations/strands/memorystore/_format.py
```diff
@@ -0,0 +1,62 @@
+"""Internal Strands-message formatting helpers."""
```
arielnabavian: port lgtm

src/bedrock_agentcore/memory/integrations/strands/README.md
```diff
@@ -26,6 +26,147 @@ Advanced memory with multiple strategies for user preferences, facts, and sessio
 
 ---
 
+## Native Strands long-term memory stores
+
+`AgentCoreMemoryStore` plugs AgentCore long-term memory directly into Strands' `MemoryManager`.
```
arielnabavian: This is meant to replace AgentCoreMemorySessionManager implementation long term. Lets remove mentions of AgentCoreMemorySessionManager since it long term will be on deprecation path

.github/workflows/integration-testing.yml
```diff
@@ -198,6 +198,7 @@ jobs:
           MEMORY_KINESIS_ARN: ${{ secrets.MEMORY_KINESIS_ARN }}
           MEMORY_ROLE_ARN: ${{ secrets.MEMORY_ROLE_ARN }}
           MEMORY_PREPOPULATED_ID: ${{ secrets.MEMORY_PREPOPULATED_ID }}
+          MEMORY_STORE_TEST_ID: ${{ secrets.MEMORY_STORE_TEST_ID }}
```
arielnabavian: Lets verify there is no existing MEMORY_STORE_TEST_ID we can re-use from the legacy AgentCoreMemorySessionManager.

TESTING.md
```diff
@@ -30,6 +30,22 @@ uv run pytest tests_integ/memory/test_controlplane.py -xvs -k "integration"
 uv run pytest tests_integ/memory/test_memory_client.py -xvs
 ```
 
+### Strands memory-store integration
```
arielnabavian: Same comment from above,

Let's verify there is no existing MEMORY_STORE_TEST_ID we can re-use from the legacy AgentCoreMemorySessionManager. 

If this is available.

TESTING.md
```diff
@@ -77,9 +93,15 @@ uv run pytest tests_integ/tools -xvs
 
 ## CI
 
-Integration tests run via `integration-testing.yml` on PRs to main. The workflow runs runtime, memory (stream delivery only), and evaluation tests in parallel.
+Integration tests run via `integration-testing.yml` on PRs to main. The workflow runs runtime,
+control-plane/client/session-manager/native memory-store memory tests, evaluation, services, policy,
```
arielnabavian: Please do not put paths like this in any comments or markdown documentation. Reference the class or primitive implementation as needed

TESTING.md
```diff
@@ -77,9 +93,15 @@ uv run pytest tests_integ/tools -xvs
...
+gateway, identity, payments, and tools in parallel. Stream delivery tests require
+`MEMORY_KINESIS_ARN` and `MEMORY_ROLE_ARN`; native memory-store tests require
+`MEMORY_STORE_TEST_ID` or fall back to `MEMORY_PREPOPULATED_ID`.
 
-Stream delivery tests fail in CI unless `MEMORY_KINESIS_ARN` and `MEMORY_ROLE_ARN` secrets are configured on the repo. Other memory integration tests are not yet run in CI due to provisioning times and flaky LLM-dependent assertions — CI support is planned once test stability is addressed.
+Changes to `.github/workflows/` trigger the workflow security gate and require maintainer security
```
arielnabavian: Valid to add this for agents to read, but not relevant to this port. We can add this in a separate PR.

src/bedrock_agentcore/memory/integrations/strands/memorystore/types.py
```diff
@@ -0,0 +1,211 @@
+"""Public types and namespace helpers for the Strands AgentCore memory store."""
```
arielnabavian, line 1: Can we have all these new memorystore related changes under /strands/memorystore so they are more contained

arielnabavian: Move AgentCoreMemorySessionManager related changes seperate /strands/  a /strands/memorysessionmanager. Do not mention this is legacy/deprecated yet, will have this in a follow up.

src/bedrock_agentcore/memory/integrations/strands/__init__.py
```diff
@@ -1,5 +1,145 @@
-"""Strands integration for Bedrock AgentCore Memory."""
+"""Strands integrations for Bedrock AgentCore Memory."""
```
arielnabavian: This runtime check is overkill

we have raised the minor version requirement to "strands-agents>=1.46.0". So client will just upgrade their strands version on the next agentcore release.

.github/workflows/integration-testing.yml
```diff
@@ -124,8 +124,8 @@ jobs:
             ignore: ""
```
arielnabavian, line 113: General high level. 

We missed a lot of the existing inline code comments from the typescript package during the port. 

Make sure we are bringing over any comments that are relevant to the python version

src/bedrock_agentcore/memory/integrations/strands/sender.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def _token_for_sequence_numbers(self, sequence_numbers: list[int | None]) -> str | None:
+        if not sequence_numbers or any(number is None for number in sequence_numbers):
+            return None
+        first = sequence_numbers[0]
+        last = sequence_numbers[-1]
+        return f"{self._memory_id}-{self._actor_id}-{self._run_id}-{first}-{last}"
+
+
+def _ecmascript_number_string(value: int | float) -> str:
```
arielnabavian: We are hyper-optimizing with these ecmascript helper functions. Please just rely use json.dumps(metadata, sort_keys=True)

src/bedrock_agentcore/memory/integrations/strands/memorystore/sender.py
```diff
@@ -0,0 +1,298 @@
...
+            encoded_value = "true" if value else "false"
+        else:
+            # Reuse wire validation so unsupported/non-finite values keep the
+            # public error contract while finite numbers use the JS formatter.
+            encoded_value = _metadata_scalar_string(key, value)
+        entries.append(f"{encoded_key}:{encoded_value}")
+    return "{" + ",".join(entries) + "}"
+
+
+def _metadata_scalar_string(key: str, value: object) -> str:
```
arielnabavian, line 254: this is _metadata_scalar_string is overkill aswell  lets maintain parity with a logic like

if value is None or (isinstance(value, float) and not math.isfinite(value)):
    raise ValueError(...)
string_value = value if isinstance(value, str) else json.dumps(value)

src/bedrock_agentcore/memory/integrations/strands/store.py
```diff
@@ -0,0 +1,259 @@
...
+from typing import TYPE_CHECKING, Any
+
+import boto3
+from botocore.config import Config as BotocoreConfig
+from strands.memory import AddMessagesContext, ExtractionConfig, MemoryEntry, MemoryStore, SearchOptions
+from strands.types.content import Message
+
+from bedrock_agentcore._utils.user_agent import build_user_agent_suffix
+
+from ._format import _ecmascript_trim
```
arielnabavian: in general for this whole port. this is a python package. lets not try to reach super optimized parity with ecmascript.

arielnabavian:
@strandly-the-agent review this

src/bedrock_agentcore/memory/integrations/strands/memorystore/store.py
```diff
@@ -0,0 +1,259 @@
...
+            else self.max_search_results or DEFAULT_MAX_SEARCH_RESULTS
+        )
+        if type(want) is not int or want < 1:
+            raise ValueError(f"AgentCoreMemoryStore.search: max_search_results must be a positive integer, got {want}")
+        top_k = want
+        if self._min_score is not None:
+            top_k = min(math.ceil(want * self._over_fetch_factor), MAX_TOPK)
+        kwargs: dict[str, Any] = {
+            "memoryId": self._memory_id,
+            "searchCriteria": {"searchQuery": query, "topK": top_k},
```
strandly-the-agent, line 247: 🔴 **Retrieval silently truncates the requested count.** `max_search_results=50` sets `topK=50`, but this omits top-level `maxResults` (service default: 20), and the response path below ignores `nextToken`. Two cold local runs returned 20 entries after one call. Values above 100 also emit an invalid `topK` because AgentCore caps it at 100.

Please set `maxResults`, follow pagination until the requested count/no token, and reject or otherwise define behavior for requests above the service cap.

src/bedrock_agentcore/memory/integrations/strands/memorystore/sender.py
```diff
@@ -0,0 +1,298 @@
...
+            session_id: Session identifier.
+            metadata_provider: Optional per-message event metadata callback.
+            run_id: Run-unique idempotency-token component. Defaults to a UUID.
+            max_turns_per_event: Maximum role-tagged turns in one event.
+            extraction_mode: Optional AgentCore long-term extraction mode.
+
+        Raises:
+            ValueError: If ``max_turns_per_event`` is not a positive integer.
+        """
+        if type(max_turns_per_event) is not int or max_turns_per_event < 1:
```
strandly-the-agent, line 129: 🔴 **Preflight every `CreateEvent` group before dispatching any of them.** Cold repros passed a 101-item payload, 100,001-character text, 16 metadata entries, a 129-character key, and a 257-character value into the client; botocore did not reject them locally. AgentCore limits these to 100 payload items, 100,000 text characters, 15 metadata entries, 1–128-character patterned keys, and 0–256-character patterned values.

Because `_send_all` dispatches sibling groups concurrently, an invalid group can fail only after another group has committed. Please fully materialize and validate every group before `gather()`, cap `max_turns_per_event` at 100, and update the tests to cover exact boundaries plus zero client calls when a later group is invalid. The README’s “any positive integer” claim must change with that cap.

strandly-the-agent (commented):
**Superseded — reconciled at `8f2fac4`.** This review was written against `8a9543ee`; the verdict below no longer reflects the current state.

- The two documentation findings (`flush()` semantics, non-runnable first example) are **fixed** in the contained `memorystore/README.md`; their original threads pointed at the old root README, which is now byte-identical to `main`, so I deleted those two stale comments rather than leave them dangling.
- The two service-contract findings (retrieval pagination/`maxResults`, full `CreateEvent` request preflight) are **intentionally not implemented**: they are behavior the TypeScript source does not have, and `arielnabavian` confirmed on the pagination thread that this matches TypeScript and belongs in a fast follow. Those two threads stay open as that record — no further nagging from me.
- Nothing here is a blocker on the current head. Verification evidence for `8f2fac4` is in the PR description.

arielnabavian:
@strandly-the-agent read all the comments on this pr. Read the comments thoroughly and makes sure we are filling the requirements of the given task, (port typescript AgentCoreMemoryStore to python AgentCoreMemoryStore). Make sure not to add any logic changes outside of whats required for 1-1 parity. Make sure the code comments are all address, make a revision to the PR once you are done.

strandly-the-agent:
Revision pushed at `bac5ffee`: all current maintainer comments are addressed while keeping the MemoryStore behavior aligned with the TypeScript source.

- Contained the APIs under `strands.memorystore` / `strands.memorysessionmanager`; preserved existing SessionManager imports and patch paths.
- Removed runtime version probing and ECMAScript-emulation helpers; retained the source retrieval/write semantics without adding pagination or service preflight.
- Reused `MEMORY_PREPOPULATED_ID`, restored unrelated `TESTING.md` content, transferred relevant source rationale comments, and corrected the new docs/examples.
- Verified: **2952 passed, 10 skipped, 4 xpassed**; Ruff and lock checks passed; 18 live tests collect; wheel build passed. Live AWS was not run.

<details><summary>Review disposition</summary>

The two earlier Strandly implementation blockers requested behavior beyond the TypeScript source. Per the explicit 1:1-parity direction, this revision intentionally does not implement them. The two documentation findings are fixed. Full traceability and validation evidence are in the updated PR description.

</details>

strandly-the-agent:
Final parity follow-up pushed at `56dcff8`.

The independent correctness pass found one Python-only overflow edge in over-fetch calculation: a huge finite factor overflowed to `inf`, where Python `math.ceil(inf)` raised instead of matching TypeScript's `Math.min(Math.ceil(...), 100)`. The calculation now clamps before integer conversion and has a regression test. This preserves source behavior; it does not add retrieval pagination or service preflight.

Current focused validation: **386 passed**; Ruff, format, lock, and diff checks pass. The PR description now points to the final head and updated traceability evidence. Live AWS remains unrun.

src/bedrock_agentcore/memory/integrations/strands/memorystore/store.py
arielnabavian: This is the same implementation in typescript. I suggest we fix this as a fast follow

arielnabavian:
For sanity, ran a manual audit as well [full_suite_interval_results.xlsx](https://github.com/user-attachments/files/30421197/full_suite_interval_results.xlsx)

arielnabavian:
[full_suite_interval_report.docx](https://github.com/user-attachments/files/30422131/full_suite_interval_report.docx)

.github/workflows/integration-testing.yml
```diff
@@ -124,8 +124,8 @@ jobs:
...
           # TODO: expand to full tests_integ/memory once test stability is addressed
           - group: memory
-            path: tests_integ/memory/test_controlplane.py tests_integ/memory/test_memory_client.py tests_integ/memory/integrations/test_session_manager.py
-            timeout: 15
+            path: tests_integ/memory/test_controlplane.py tests_integ/memory/test_memory_client.py tests_integ/memory/integrations/strands/memorysessionmanager/test_session_manager.py tests_integ/memory/integrations/strands/memorystore/test_memory_store.py
+            timeout: 30
```
jariy17: Why are we increasing this 30 minutes now?

arielnabavian:
@strandly-the-agent make a revision. 

Make sure we are not messing up any imports for existing clients. specifically these 1s
Run actions/github-script@v8
RequestError [HttpError]: Resource not accessible by integration
    at /home/runner/work/_actions/actions/github-script/v8/dist/index.js:9537:21
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async eval (eval at callAsyncFunction (/home/runner/work/_actions/actions/github-script/v8/dist/index.js:36187:16), <anonymous>:27:3)
    at async main (/home/runner/work/_actions/actions/github-script/v8/dist/index.js:36285:20) {
  status: 403,
  response: {
    url: 'https://api.github.com/repos/aws/bedrock-agentcore-sdk-python/issues/588/comments',
    status: 403,
    headers: {
      'access-control-allow-origin': '*',
      'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset, Warning',
      'content-encoding': 'gzip',
      'content-security-policy': "default-src 'none'",
      'content-type': 'application/json; charset=utf-8',
      date: 'Thu, 23 Jul 2026 21:25:52 GMT',
      'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
      server: 'github.com',
      'strict-transport-security': 'max-age=31536000; includeSubdomains; preload',
      'transfer-encoding': 'chunked',
      vary: 'Accept-Encoding, Accept, X-Requested-With',
      'x-accepted-github-permissions': 'issues=write; pull_requests=write',
      'x-content-type-options': 'nosniff',
      'x-frame-options': 'deny',
      'x-github-api-version-selected': '2022-11-28',
      'x-github-media-type': 'github.v3; format=json',
      'x-github-request-id': '0440:130C5F:E9623B:ED2DED:6A6286E0',
      'x-ratelimit-limit': '15000',
      'x-ratelimit-remaining': '14992',
      'x-ratelimit-reset': '1784843964',
      'x-ratelimit-resource': 'core',
      'x-ratelimit-used': '8',
      'x-xss-protection': '0'
    },
    data: {
      message: 'Resource not accessible by integration',
      documentation_url: 'https://docs.github.com/rest/issues/comments#create-an-issue-comment',
      status: '403'
    }
  },
  request: {
    method: 'POST',
    url: 'https://api.github.com/repos/aws/bedrock-agentcore-sdk-python/issues/588/comments',
    headers: {
      accept: 'application/vnd.github.v3+json',
      'user-agent': 'actions/github-script octokit-core.js/5.0.1 Node.js/24',
      authorization: 'bearer [REDACTED]',
      'content-type': 'application/json; charset=utf-8'
    },
    body: '{"body":"\\n## ⚠️ Breaking Change Warning\\n\\nFound **3** potential breaking change(s) in this PR:\\n\\n\\u001b[1msrc/bedrock_agentcore/memory/integrations/strands/session_manager.py\\u001b[0m:0: logger: \\u001b[33mPublic object was removed\\u001b[39m\\n\\u001b[1msrc/bedrock_agentcore/memory/integrations/strands/bedrock_converter.py\\u001b[0m:0: logger: \\u001b[33mPublic object was removed\\u001b[39m\\n\\u001b[1msrc/bedrock_agentcore/memory/integrations/strands/converters/openai.py\\u001b[0m:0: logger: \\u001b[33mPublic object was removed\\u001b[39m\\n\\n---\\n> **Note:** This is an automated static analysis check. Some flagged changes may be intentional.\\n> Please confirm each item is expected and, if so, add a migration note to `CHANGELOG.md`."}',
    request: {
      agent: [Agent],
      fetch: [Function: proxyFetch],
      hook: [Function: bound bound register]
    }
  }
}
Error: Unhandled error: HttpError: Resource not accessible by integration


I also recommended moving our converters back under strands folder rather than memorysessionmanager folders since these are more reusable.


Also please investigate this bug we found in a bug bash:

 1. max_search_results > 100 throws a raw service error — the clamp only runs on the min_score path
Category: recall · Severity: medium — plausible input, opaque failure, inconsistent with sibling path

store.py:208-213 computes top_k:

top_k = want
if self._min_score is not None:
    over_fetch = want * self._over_fetch_factor
    top_k = MAX_TOPK if over_fetch >= MAX_TOPK else math.ceil(over_fetch)
MAX_TOPK (100, the real service cap) is applied only inside the min_score branch. With no min_score, want goes to the wire verbatim. Construction validates max_search_results >= 1 but never <= 100.

Verified against the live service:

config	topK sent	result
max_search_results=100, no min_score	100	OK
max_search_results=101, no min_score	101	ValidationException
max_search_results=200, no min_score	200	ValidationException
search(opts={"max_search_results": 200})	200	ValidationException
max_search_results=200, min_score=0.1	100	OK — clamped
ValidationException: An error occurred (ValidationException) when calling the
RetrieveMemoryRecords operation: 1 validation error detected: Value at
'searchCriteria.topK' failed to satisfy constraint: Member must have value less
than or equal to 100
Adding a min_score makes a previously-broken call start working, which is a confusing signal. Everything else in this port validates eagerly with an AgentCoreMemoryStore:-prefixed message; this one path defers to a raw botocore error.

Fix: hoist the clamp out of the if — top_k = min(top_k, MAX_TOPK) — or reject max_search_results > MAX_TOPK at construction and in search.

Revise the CR.

strandly-the-agent:
Revision pushed at `8c06a2c`. All three items from your comment are handled.

**1. Imports for existing clients.** The griffe step in *Detect Breaking Changes* exits 0 — the job died on its `actions/github-script` step (`403 Resource not accessible by integration`; `GITHUB_TOKEN` is read-only for fork PRs), so the report you pasted never made it onto the PR. It was right: 3 × `logger: Public object was removed`, caused by moving the SessionManager into `strands/memorysessionmanager/` behind re-export shims. That move also silently renamed the module loggers (`...strands.session_manager` → `...memorysessionmanager.session_manager`), which would break customer logging config keyed on those names. Reverted: `session_manager.py`, `config.py`, `bedrock_converter.py`, `converters/*` and the package root are byte-identical to `main`. griffe now reports **0 breaking changes**, and a runtime probe of every pre-existing module (`dir()`, `__all__`, `logger.name`, owning module per symbol) matches a `main` worktree exactly.

**2. Converters back under `strands/`.** Done — `strands/converters/{__init__,openai,protocol}.py` are byte-identical to `main` again, so the reusable converters stay where they were and nothing under `memorysessionmanager/` exists in this PR.

**3. Bug-bash defect: `max_search_results > 100`.** Fixed — `MAX_TOPK` is now applied on every retrieval path instead of only inside the `min_score` branch, and a one-time warning reports the reduced cap so it is not silent. Your matrix, reproduced offline at `8c06a2c`:

| Config | `topK` before | `topK` now |
|---|---|---|
| `max_search_results=100`, no `min_score` | 100 | 100 |
| `max_search_results=101`, no `min_score` | 101 → `ValidationException` | **100** |
| `max_search_results=200`, no `min_score` | 200 → `ValidationException` | **100** |
| `search(options={"max_search_results": 200})` | 200 → `ValidationException` | **100** |
| `max_search_results=200`, `min_score=0.1` | 100 | 100 |

This is a **deliberate divergence from the TypeScript source** (which clamps only the over-fetch path) — flagged as such in the PR description so it does not read as a parity slip. I went with hoisting the clamp rather than rejecting `>100` eagerly, because rejecting would also break `max_search_results=200, min_score=0.1`, which works today; one line to flip if you prefer the hard error. The tests that asserted `topK == 101` are gone, replaced by 100/101/200 coverage with and without a score floor, the per-call `search_memory` option, and warn-once behaviour.

<details><summary>Evidence & open items</summary>

- At `8c06a2c`: **386 passed** focused, **3024 passed** full suite (the 2 remaining failures and the `deepeval` collection error reproduce unchanged on `main` with the same venv). `ruff`, `ruff format`, `mypy` on the new package, `griffe` (0), and `git diff --check` clean; 18 live tests collect (live AWS not run).
- Branch also merges current `main` — it was `CONFLICTING`, now `MERGEABLE`; `pyproject`/`uv.lock` keep main's version and evals pins plus the `strands-agents>=1.46.0` floor.
- Only pre-existing file still modified: two `TestAsyncMode` cases moved to the public `HookRegistry.get_callbacks_for` API — required by the 1.46 floor; they fail on `main` too under 1.46.
- Pagination stays the fast follow you called on the `store.py` thread (a request for >100 records returns the first 100). The `CreateEvent` request-bound preflight is in the same bucket — happy to pull either into this PR.
- `AgentCoreMemorySessionManager` containment is the one item I did not carry over, since it is what broke the import surface; ready to open it as a standalone PR.
- `safety-gate` stays red by design: this PR still touches `.github/workflows/integration-testing.yml` to add the new live test to the memory matrix and install `pytest-asyncio`. Happy to drop that hunk if you would rather wire CI separately.

</details>

src/bedrock_agentcore/memory/integrations/strands/memorystore/store.py
strandly-the-agent: Half of this is fixed at `8c06a2c`; the other half is the fast follow you called.

**Fixed — the `topK` cap.** `MAX_TOPK` now applies on every retrieval path, not just inside the `min_score` branch, so `max_search_results` of 101/200 (constructor or per-call `search_memory` option) sends `topK=100` instead of failing with a raw `ValidationException`; a one-time warning names the store and the reduced cap so it is not silent. Tests cover 100/101/200 with and without a score floor, plus warn-once. This is a deliberate divergence from the TypeScript source, noted as such in the PR description.

**Still open by design — pagination.** `maxResults`/`nextToken` are not sent, so a request for more than 100 records returns the first 100, matching the source. Left for the fast follow.

arielnabavian (approved):

src/bedrock_agentcore/memory/integrations/strands/memorystore/_format.py
```diff
@@ -0,0 +1,52 @@
...
+
+from typing import Literal
+
+from strands.types.content import Message
+
+AgentCoreRole = Literal["USER", "ASSISTANT"]
+
+
+def map_role(message: Message) -> AgentCoreRole:
```
jariy17: It seems like these functions can be part of the AgentCoreSender class as private function _map_roles(). Let's shift these functions into AgentCoreSender if its only used by that class.

src/bedrock_agentcore/memory/integrations/strands/memorystore/types.py
```diff
@@ -0,0 +1,221 @@
...
+DEFAULT_OVERFETCH_FACTOR = 4
+# Bound score-filter over-fetching even when callers configure a large multiplier.
+MAX_TOPK = 100
+# Packing turns lets extraction cadence control API volume instead of writing one event per message.
+DEFAULT_MAX_TURNS_PER_EVENT = 50
+# Store-supplied record fields use this prefix to avoid colliding with user metadata.
+RESERVED_METADATA_PREFIX = "_"
+
+
+class AgentCoreDataPlaneClient(Protocol):
```
jariy17: Cant we use the MemoryClient we already vend in this repo?

src/bedrock_agentcore/memory/integrations/strands/memorystore/factory.py
```diff
@@ -0,0 +1,152 @@
...
+from .store import AgentCoreMemoryStore, _create_data_plane_client
+from .types import (
+    AgentCoreDataPlaneClient,
+    AgentCoreExtractionConfig,
+    AgentCoreNamespaceConfig,
+    MetadataProvider,
+)
+
+
+def assert_writable_topology(stores: Sequence[MemoryStore], expect_extraction: bool = False) -> None:
```
jariy17: _assert_writable_topology

jariy17 (changes requested):

arielnabavian:
@strandly-the-agent address these comments and revise the pr.

src/bedrock_agentcore/memory/integrations/strands/memorystore/_format.py
strandly-the-agent: Checked usage first, and you're right that they were sender-exclusive: `map_role`, `extract_text` and `is_user_or_assistant_with_text` had exactly three call sites, all in `sender.py` (message filtering and payload construction). `store.py`'s only `_format`-looking symbol was its unrelated local `_format_created_at`, so nothing was shared.

Done in `7bc3595`: `_format.py` is deleted and the three are private members of `AgentCoreEventSender` (`_map_role`, `_extract_text`, `_is_user_or_assistant_with_text`), with the role literal now `_AgentCoreRole` inside `sender.py`. Pure moves — no behavior change. `test_format.py` became `test_message_formatting.py` and targets the private helpers directly.

src/bedrock_agentcore/memory/integrations/strands/memorystore/types.py
strandly-the-agent: Yes — and as of `5c5c0b0` we now use it for the **default** path too, not just accept one: `_create_data_plane_client()` builds `MemoryClient(region_name=..., boto3_session=..., integration_source="strands")` and calls the data-plane client it vends, so AgentCore client construction and the user agent live in one place instead of a second hand-rolled boto3 client. Passing your own works as well — `client=MemoryClient(...)` or a boto3 `bedrock-agentcore` client, both accepted by the store, factory and sender.

What the store still can't call is `MemoryClient`'s higher-level methods, which is why this `Protocol` describes the two data-plane operations rather than `MemoryClient` itself:

- `MemoryClient.create_event(memory_id, actor_id, session_id, messages=[(text, role)], ...)` has no `clientToken` parameter. That token is how a Strands batch re-fire dedupes; without it retries would create duplicate events.
- `MemoryClient.retrieve_memories(...)` catches `ClientError` and returns `[]`. `MemoryManager` does its own per-store failure isolation and this store's contract is that retrieval errors propagate, so a swallowed error would look like "no memories".
- `retrieve_memory_records` isn't defined on `MemoryClient` at all — `__getattr__` forwards it to `gmdp_client`, i.e. exactly the client we use.

Tradeoff worth naming: `MemoryClient.__init__` also builds a control-plane client we never call. It's one construction per store set (the factory shares one), no network, so I took that over duplicating client setup — easy to revert if you'd rather not pay it.

Covered by `test_store.py::test_default_client_is_built_through_the_sdk_memory_client`, `::test_store_accepts_the_sdk_memory_client_for_reads_and_writes`, `test_factory.py::test_factory_accepts_the_sdk_memory_client_as_shared_client`, and a mypy fixture case.

src/bedrock_agentcore/memory/integrations/strands/memorystore/factory.py
strandly-the-agent: Renamed in `7bc3595`: `_assert_writable_topology`. Also removed from the package `__all__` (and from the export test), since it's an internal guard the factory calls — it was never a shipped name, so no compatibility concern.

.github/workflows/integration-testing.yml
strandly-the-agent: Fair question, and the answer is that this PR shouldn't be raising it — the hunk is gone in `7bc3595`. `.github/workflows/integration-testing.yml` is now byte-identical to `main`, so this PR touches no workflow (which also means `safety-gate` no longer blocks integration tests for it).

For the record, the reason I'd bumped it: the new live suite polls long-term extraction, which is eventually consistent — three of its cases wait up to 240s each — on top of the existing control-plane/client/session-manager files already budgeted at 15m for that matrix group. Rather than defend a number, wiring the live suite into CI is better as a maintainer-owned change; when you want it, the memory group needs `tests_integ/memory/integrations/test_memory_store.py` added to `path`, `pytest-asyncio` in the install line, and whatever timeout you're comfortable with (the polling budget is tunable in the test if you'd rather keep 15).

arielnabavian:
@strandly-the-agent address these comments from 
jariy17 and revise the pr

strandly-the-agent (commented):
Replies to `jariy17`'s review, published at `5c5c0b0`. These were drafted yesterday but sat in an unsubmitted review — my mistake, they were never visible. Per-thread detail is inline; the short version:

1. **Format helpers** — all three were sender-exclusive, so `_format.py` is gone and they are private members of `AgentCoreEventSender`.
2. **`MemoryClient`** — now used to build the default client, and accepted when you pass one; the `Protocol` stays because the store needs the raw data-plane calls (`clientToken`, propagating errors).
3. **`_assert_writable_topology`** — renamed, and dropped from the package `__all__`.
4. **30-minute integration timeout** — hunk removed; this PR no longer touches any workflow.

strandly-the-agent:
Revised at `5c5c0b0`, and my earlier replies to `jariy17` are now actually visible — they had been left in an unsubmitted review, which is why this looked unanswered. Sorry for the round trip.

All four of `jariy17`'s points, with the code:

1. **Format helpers** — all three were sender-exclusive (only call sites were in `sender.py`), so `_format.py` is deleted and they are private members of `AgentCoreEventSender`.
2. **`MemoryClient`** — went further than "you can pass one": the default connection is now built through `MemoryClient(..., integration_source="strands")` and the store calls the data-plane client it vends, so there's no second hand-rolled boto3 client. Its higher-level `create_event`/`retrieve_memories` still can't be used (no `clientToken`, and swallowed `ClientError`), which is what the two-method `Protocol` documents.
3. **`_assert_writable_topology`** — renamed and dropped from the package `__all__`.
4. **30-minute integration timeout** — hunk removed; this PR no longer touches any workflow, so `safety-gate` doesn't block the integration matrix.

Also rebased onto current `main`. The PR now touches no existing module, test, doc or workflow — only `pyproject.toml`/`uv.lock` plus the new `memorystore` package and its tests.

<details><summary>Validation at <code>5c5c0b0</code></summary>

`pytest tests/bedrock_agentcore/memory` → **774 passed** · focused Strands suite → **389 passed** · `tests/unit` → 344 · `tests/integration` → 28 passed, 1 skipped · config_bundle/evaluation/gateway/payments/services/tools → all passed · `ruff check` + `ruff format --check` clean · `mypy` on the new package clean · griffe **0 breaking changes** vs `main`.

Two caveats I won't paper over: the whole-`tests/` run doesn't finish in my sandbox — `runtime/test_a2a.py` fails twice and `identity` attempts a live `bedrock-agentcore` call — but both reproduce at the merged base with my changes stashed, so they're environmental; CI is the authority. Live AWS memory tests still need `MEMORY_PREPOPULATED_ID` and Bedrock access.

An independent review pass over the client-construction change is folded in (documented that `MemoryClient` re-checks the session by truthiness). The one tradeoff worth your call: `MemoryClient.__init__` also builds a control-plane client we never use — one construction per store set, no network. Happy to go back to a direct boto3 client if you'd rather not pay it.

</details>

jariy17 (dismissed):

arielnabavian:
@strandly-the-agent Make sure this breaking change is address    body: '{"body":"\\n## ⚠️ Breaking Change Warning\\n\\nFound **2** potential breaking change(s) in this PR:\\n\\n\\u001b[1msrc/bedrock_agentcore/runtime/a2a.py\\u001b[0m:0: A2A_CONTRACT_PORT: \\u001b[33mPublic object was removed\\u001b[39m\\n\\u001b[1msrc/bedrock_agentcore/runtime/a2a.py\\u001b[0m:0: A2A_PORT_ENV: \\u001b[33mPublic object was removed\\u001b[39m\\n\\n---\\n> **Note:** This is an automated static analysis check. Some flagged changes may be intentional.\\n> Please confirm each item is expected and, if so, add a migration note to `CHANGELOG.md`."}',

We don't want to break our customers imports.

Additionally get rid of this scoped down AgentCoreDataPlaneClient and AgentCoreClient. users should only be able to pass in the normal MemoryClient

strandly-the-agent:
Revised at `3de6915`. Both points handled.

**1. The breaking-change report is a stale-branch artifact, not a removal.** `A2A_CONTRACT_PORT` and `A2A_PORT_ENV` were *added on `main`* by #615 (`207adb7`), after this branch's merge-base. griffe diffs `main` → PR tree, so it saw them missing from the older tree and reported "Public object was removed". This PR has never touched `src/bedrock_agentcore/runtime/a2a.py` — no customer import is affected. Merged current `main`; griffe now reports **0 breaking changes**, and the pre-existing Strands modules are still byte-identical to `main` (import paths, patch targets, logger names untouched).

**2. `AgentCoreDataPlaneClient` and `AgentCoreClient` are gone.** The public contract is now just `MemoryClient`:

```python
# default — the integration builds MemoryClient(integration_source="strands") itself
store = AgentCoreMemoryStore(memory_id=..., actor_id=..., session_id=..., namespace="/facts/{actorId}")

# or reuse your own connection
store = AgentCoreMemoryStore(..., client=MemoryClient(region_name="us-west-2"))
```

`create_agentcore_memory_stores` and `AgentCoreEventSender` take the same thing, and `create_agentcore_memory_stores` shares one client across the topology. `resolve_data_plane_client()` is deleted too; each entry point simply uses the `bedrock-agentcore` data plane the client vends — needed because `MemoryClient.create_event` can't carry the `clientToken` the sender uses for idempotent retries, and `MemoryClient.retrieve_memories` swallows `ClientError` that `MemoryManager` expects to surface.

<details><summary>Validation at <code>3de6915</code></summary>

`pytest tests/bedrock_agentcore/memory` → **774 passed** · memorystore package → 163 · `tests/unit`+`tests/integration` → 372 passed, 1 skipped · runtime/evaluation/gateway/services/tools/config_bundle → 1134 passed, 4 xpassed, 2 failed\* · `ruff` + `ruff format` clean · `mypy` on the new package clean · griffe **0** · 18 live tests collect.

\* Those two are `runtime/test_a2a.py::TestBuildA2AApp::test_auto_builds_card_when_none_provided` and `::TestServeA2A::test_explicit_port_overrides_environment`; they reproduce identically at the merged base with my commit stashed, so they're environmental in my sandbox (a2a-sdk version), not this branch. `tests/bedrock_agentcore/identity` also attempts a live AWS call here and hangs. CI is the authority for both.

An independent review pass ran over the client-contract commit; its one finding (a mypy fixture block that became a duplicate once the Protocol was gone) is folded in. Test doubles now wrap a data-plane mock in `Mock(spec=MemoryClient)`, and the live suite swaps the vended `gmdp_client` on a copied `MemoryClient` to count/capture calls.

One mechanical note: my sandbox token can't push `.github/workflows/**`, so the workflow file `main` updated came in through the API as `3de6915`. The PR still shows no workflow diff.

</details>

jariy17 (approved):

PR #591, opened 2026-07-22
feat(a2a): migrate runtime integration to a2a-sdk v1

notgitika:
## Summary
- migrate `serve_a2a` and `build_a2a_app` from removed 0.3 app APIs to the v1 route factories
- update agent cards, call context handling, and the A2A dependency range to `>=1.0.1,<2.0`
- retain v0.3 JSON-RPC compatibility while serving the v1 protocol and agent-card endpoint

The lower bound is 1.0.1 because a2a-sdk 1.0.0 imports `grpc` from its route module without including it in the `http-server` extra; 1.0.1 is the first HTTP-only release that starts successfully.

## Validation
- `uv run pytest -q` (2,794 passed, 10 skipped, 4 xpassed)
- focused A2A suite against a2a-sdk 1.0.1 (36 passed)
- focused A2A suite against a2a-sdk 1.1.2 (36 passed)
- Ruff and MyPy pass

Closes #509

PR #593, opened 2026-07-23
fix(a2a): honor PORT when serving locally

notgitika:
## Summary

- Resolve `serve_a2a`'s port from the explicit argument, then `PORT`, then the existing `9000` default.
- Preserve explicit `port=` precedence for backwards compatibility.
- Document the environment fallback and cover default, environment, and explicit-port behavior.

This is stacked on #591 because the current `main` branch does not yet contain the A2A SDK v1 `serve_a2a` implementation. It enables the CLI to assign distinct ports to multiple local CodeZip A2A runtimes.

## Testing

- `pytest tests/bedrock_agentcore/runtime/test_a2a.py -q` (28 passed)
- `ruff check` and `ruff format --check` on changed Python files
- Full suite: 2,795 passed, 10 skipped, 4 xpassed; the sole sandbox-blocked localhost integration test passed separately with socket access

### DIFF
--- a/src/bedrock_agentcore/runtime/a2a.py
+++ b/src/bedrock_agentcore/runtime/a2a.py
@@ -287,7 +287,7 @@ def serve_a2a(
     executor: Any,
     agent_card: Any = None,
     *,
-    port: int = 9000,
+    port: Optional[int] = None,
     host: Optional[str] = None,
     task_store: Any = None,
     context_builder: Any = None,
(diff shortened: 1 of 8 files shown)


src/bedrock_agentcore/runtime/a2a.py
jariy17: PORT is used for uvicorn, but the generated agent card still defaults to localhost:9000. with PORT=9001, the server binds to 9001 while the card advertises 9000. can we pass the resolved port into the card URL and add a test for it?

jariy17 (commented):
pretty good, just some edge cases to solve for

src/bedrock_agentcore/runtime/a2a.py
notgitika: Addressed in 2b06b3b. `serve_a2a()` now passes its resolved local port into auto-generated card construction while preserving `AGENTCORE_RUNTIME_URL` precedence and explicit-card behavior. The updated test verifies that `PORT=9001` drives both uvicorn and `supportedInterfaces[0].url`. Validated with 38 unit/integration tests, Ruff, and mypy.

jariy17 (approved):

PR #594, opened 2026-07-23
fix: convert RST admonition to ADOC

nborges-aws:
*Description of changes:*
Update `render_adoc.py` script to handle two more conversion cases from RST to adoc:
- admonitions (note, warning, etc.) formatting. `.. warning::` → `[WARNING]\n====\n...\n====`
- RST class formatting. `:py:meth:bar` → `bar` monospace

Applied in relevant call sites


notgitika (approved):
LGTM but as a note maybe we should add tests/regression tests here? Could be a follow up or just a backlog item imo

PR #604, opened 2026-07-28
fix(memory-integ): address capacity cap and update failures in tests

Hweinstock:
Failing integ tests on main: https://github.com/aws/bedrock-agentcore-sdk-python/actions/runs/30303039785/job/90288106089. 

### Problem

Failing integ tests on main: https://github.com/aws/bedrock-agentcore-sdk-python/actions/runs/30303039785/job/90288106089 with two root causes:

- `add_strategy(wait_for_active=True)` only waited for the strategy to reach ACTIVE, but the memory itself could still be in `UPDATING` state. Consecutive `add_strategy` calls (as in `test_add_strategy_immediately_after_another`) would hit `ValidationException: Memory is in transitional state UPDATING`.
- CI account hit memory limit with 150. 

## Solution

  After the strategy is confirmed `ACTIVE`, `add_strategy` now also checks the memory status and waits for it to return to `ACTIVE` using the existing `_wait_for_memory_active` utility.

## Testing
  - All 6 integ tests pass against dev account (previously test 4 failed consistently)

### Notes
there appears to be an issue with PRs from forks: https://github.com/aws/bedrock-agentcore-sdk-python/pull/603 so make this from origin branch.

### DIFF
--- a/src/bedrock_agentcore/memory/controlplane.py
+++ b/src/bedrock_agentcore/memory/controlplane.py
@@ -379,9 +379,13 @@ class MemoryControlPlaneClient:
                     break
 
             if strategy_id:
-                return self._wait_for_strategy_active(memory_id, strategy_id, max_wait, poll_interval)
+                memory = self._wait_for_strategy_active(memory_id, strategy_id, max_wait, poll_interval)
             else:
                 logger.warning("Could not identify newly added strategy %s to wait for activation", strategy_name)
+
+            # Also ensure the memory itself has returned to ACTIVE
+            if memory.get("status") != MemoryStatus.ACTIVE.value:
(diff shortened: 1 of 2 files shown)


Hweinstock:
```
FAILED tests_integ/evaluation/test_dataset_client.py::TestDatasetCrud::test_create_dataset_and_wait_inline - botocore.errorfactory.ConflictException: An error occurred (ConflictException) when calling the CreateDataset operation: A dataset with this name already exists in your account. Please choose a different name and try again.
```

evaluator failure is likely a flaky test, can track as a follow-up.

notgitika (approved):

nborges-aws (approved):

PR #620, opened 2026-08-04
feat(payments): add MPP (Machine Payments Protocol) support

rajuans:
Add MPP alongside the existing x402 protocol in ProcessPayment. MPP servers answer 402 with WWW-Authenticate: Payment challenges; the SDK selects the one challenge the payment instrument can satisfy and mints an Authorization header.

Core:
- mpp.py: parse WWW-Authenticate challenges and select one (charge-intent, unexpired, method the instrument satisfies: evm/tempo->ETHEREUM, solana->SOLANA; ordered by NETWORK_PREFERENCES). The selected challenge is forwarded verbatim so the challenge HMAC stays valid.
- generate_payment_header auto-detects protocol from the 402 and returns {"Authorization": "Payment <token>"} for MPP; x402 path unchanged.
- Optional buyer_pays_gas_fees passthrough (MPP buyerPaysGasFees).
- _model_patch.py: inject MPP shapes into the loaded botocore model when the installed release predates them; idempotent, no-op once shipped natively, never mutates installed botocore.

Integrations:
- Strands plugin and LangGraph middleware detect MPP 402s (WWW-Authenticate) and settle them; buyer_pays_gas_fees exposed on the shared config.

Tests: unit coverage for parsing, selection, model patch, manager routing, and both integrations; integ tests gated on TEST_* env vars.

Verified against the live mpp.dev 402 and prod ProcessPayment (request shape accepted; end-to-end settlement pending account MPP entitlement).

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/bedrock_agentcore/payments/constants.py
+++ b/src/bedrock_agentcore/payments/constants.py
@@ -34,6 +34,14 @@ class PaymentConnectorType(Enum):
     STRIPE_PRIVY = "StripePrivy"
 
 
+class PaymentType(Enum):
+    """Payment protocols supported by ProcessPayment."""
+
+    CRYPTO_X402 = "CRYPTO_X402"
+    # MPP does not differentiate between crypto and fiat — one value covers both.
+    MPP = "MPP"
+
+
 class PaymentsAuthorizerType(Enum):
     """Payment manager authorizer types."""
 
@@ -70,3 +78,37 @@ NETWORK_PREFERENCES = [
     "eip155:84532",  # Base Sepolia (testnet)
     "eip155:11155111",  # Ethereum Sepolia (Test)
 ]
+
+# ─────────────────────────────────────────────────────────────
+# MPP (Machine Payments Protocol) — https://mpp.dev
+# ─────────────────────────────────────────────────────────────
+
+# Default MPP protocol version sent to ProcessPayment. The service model constrains
+# this to a bare numeric string (^[0-9]+$).
+MPP_DEFAULT_VERSION = "1"
+
+# The `Payment` auth-scheme name used in `WWW-Authenticate` / `Authorization` headers.
+MPP_AUTH_SCHEME = "Payment"
+
+# Only the `charge` intent is implemented. Challenges advertising `session` or
+# `subscription` are filtered out during selection.
+MPP_SUPPORTED_INTENT = "charge"
+
+# Payment method identifiers, mapped to the blockchain family of the payment
+# instrument that can satisfy them. Tempo is an EVM chain, so it maps to ETHEREUM.
+MPP_METHOD_BLOCKCHAIN = {
+    "evm": "ETHEREUM",
+    "tempo": "ETHEREUM",
+    "solana": "SOLANA",
+}
+
+# Maps a Solana `methodDetails.network` value to the network identifier used in
+# NETWORK_PREFERENCES. Per draft-solana-charge-00 the field is optional and
+# defaults to mainnet.
+MPP_SOLANA_NETWORK_ALIASES = {
+    "mainnet": "solana-mainnet",
+    "devnet": "solana-devnet",
+    "localnet": "solana-testnet",
+    "testnet": "solana-testnet",
+}
+MPP_SOLANA_DEFAULT_NETWORK = "mainnet"
--- a/src/bedrock_agentcore/payments/constants.py
+++ b/src/bedrock_agentcore/payments/constants.py
@@ -105,10 +105,15 @@ MPP_METHOD_BLOCKCHAIN = {
 # Maps a Solana `methodDetails.network` value to the network identifier used in
 # NETWORK_PREFERENCES. Per draft-solana-charge-00 the field is optional and
 # defaults to mainnet.
+#
+# `localnet` is deliberately absent. It denotes a distinct local RPC/Surfpool
+# environment, not Solana testnet; aliasing it would let a local-only challenge
+# satisfy a solana-testnet preference and outrank a genuinely payable devnet
+# challenge. Unmapped values are left unranked rather than misranked — they remain
+# selectable when they are the only option, but never win on preference order.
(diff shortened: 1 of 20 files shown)


src/bedrock_agentcore/payments/manager.py
```diff
@@ -1088,6 +1111,169 @@ def generate_payment_header(
...
+        mpp_input: Dict[str, Any] = {
+            "version": MPP_DEFAULT_VERSION,
+            "wwwAuthenticateHeaders": [selected["raw"]],
+        }
+
+        # Only send buyerPaysGasFees when the caller expressed an intent. Omitting it
+        # leaves the protocol default in place rather than asserting a choice the
+        # caller did not make.
+        if buyer_pays_gas_fees is not None:
+            mpp_input["buyerPaysGasFees"] = bool(buyer_pays_gas_fees)
```
aidandaly24: `bool()` treats every non-empty string as true, so `buyer_pays_gas_fees="false"` sends `buyerPaysGasFees: true` and authorizes additional wallet charges. Please require `None` or an actual bool and forward it unchanged, matching the strict integration-config validation.

src/bedrock_agentcore/payments/manager.py
```diff
@@ -1012,31 +1039,27 @@ def generate_payment_header(
...
+                logger.debug("Detected MPP payment required request")
+                return self._generate_mpp_payment_header(
+                    payment_instrument_id=payment_instrument_id,
+                    payment_session_id=payment_session_id,
+                    payment_required_request=payment_required_request,
+                    user_id=user_id,
+                    network_preferences=network_preferences,
+                    client_token=client_token,
+                    buyer_pays_gas_fees=buyer_pays_gas_fees,
+                )
```
aidandaly24: Inline MPP/x402 endpoints intentionally advertise both protocols. This unconditional branch chooses MPP whenever any Payment challenge exists, then fails if that challenge does not match the instrument without inspecting a satisfiable x402 option. Should we fall back to x402 when MPP selection is unsatisfiable and an x402 requirement is present?

src/bedrock_agentcore/payments/constants.py
```diff
@@ -70,3 +78,37 @@ class PaymentsAuthorizerType(Enum):
...
+    "solana": "SOLANA",
+}
+
+# Maps a Solana `methodDetails.network` value to the network identifier used in
+# NETWORK_PREFERENCES. Per draft-solana-charge-00 the field is optional and
+# defaults to mainnet.
+MPP_SOLANA_NETWORK_ALIASES = {
+    "mainnet": "solana-mainnet",
+    "devnet": "solana-devnet",
+    "localnet": "solana-testnet",
```
aidandaly24: Solana MPP uses localnet as a distinct local RPC/Surfpool environment, not Solana testnet. This alias lets a local-only challenge satisfy a solana-testnet preference and beat a payable devnet challenge. Should we keep localnet distinct or unranked unless explicitly supported?

src/bedrock_agentcore/payments/manager.py
rajuans: Great finding!

src/bedrock_agentcore/payments/constants.py
rajuans: yes, it should be just devnet/testnet.

src/bedrock_agentcore/payments/manager.py
rajuans: Fixed in 41da888 — good catch, this was a real money-moving bug.

`bool()` is gone. The value must now be `None` or an actual `bool` and is forwarded unchanged; anything else raises `PaymentError` rather than being coerced. This matches the strict integration-config validation as you suggested.

Added a regression test for the exact case you named (`buyer_pays_gas_fees="false"` must never authorize charges), plus a parametrized test covering `"false"`, `"true"`, `"no"`, `1`, `0`, `[]`, `{}` — all rejected, and `process_payment` is never called.

src/bedrock_agentcore/payments/manager.py
rajuans: Agreed, and fixed in 41da888 — MPP now falls back to x402 when no advertised challenge is satisfiable and the same 402 carries a usable x402 requirement.

Behavior:

| 402 contents | Outcome |
|---|---|
| Satisfiable MPP challenge (± x402) | MPP → `Authorization` |
| Unsatisfiable MPP **+** usable x402 `accepts` | Falls back to x402 → `X-PAYMENT` |
| Unsatisfiable MPP, no usable x402 | `PaymentError` |

One deliberate constraint worth flagging: the fallback catches **only** `MppChallengeSelectionError`, which is raised before anything is submitted. A failure *after* the payment reaches the service (insufficient budget, malformed output, verification failure) still propagates — falling back at that point could charge the buyer twice. There is a test asserting exactly one `process_payment` call in that case.

To make this work, `_generate_mpp_payment_header` now lets `MppChallengeSelectionError` propagate instead of converting it to `PaymentError`, so the dispatcher can tell a selection failure apart from a submission failure. Callers still see a `PaymentError` when no fallback exists.

`_has_x402_payment_required` is read-only and non-raising, so a malformed x402 payload reports "no fallback available" rather than masking the original MPP error. Empty `accepts`, missing `x402Version`, and absent body are all treated as no-fallback.

src/bedrock_agentcore/payments/constants.py
rajuans: Fixed in 41da888 — the `localnet` alias is removed, so the map is now `mainnet`/`devnet`/`testnet` only (matching @rajuans' confirmation).

`localnet` is now left **unranked rather than misranked**: `challenge_network()` returns `None`, so it sorts last on preference order but stays selectable when it is the only advertised option. That keeps a local Surfpool setup usable for development without ever letting it outrank a payable public-network challenge.

Two tests cover the regression you identified:
- a `localnet` + `devnet` pair must select `devnet`
- `localnet` alone is still selectable

PR #622, opened 2026-08-05
fix: deep eval adapter to extract multi-turn conversations from service-normalized SESSION format

ybdarrenwang:
### Background: PR #568

[PR #568](https://github.com/aws/bedrock-agentcore-sdk-python/pull/568) (`feat: third-party eval metrics adapter (DeepEval + Autoevals) with strands-evals mappers`) introduced the `DeepEvalAdapter` — a generic wrapper that takes any DeepEval metric and runs it inside an AgentCore Lambda evaluator. The adapter uses strands-evals mappers to auto-detect span formats, extract fields, and construct the appropriate DeepEval test case (`LLMTestCase` for single-turn, `ConversationalTestCase` for multi-turn).

### The Bug
1. At SESSION level, the AgentCore eval service collapses all spans sharing a session.id into one span with multiple span_events (one per turn):
```
Session
 └─ Trace 1 → ONE span
                └─ span_events: [ {body: turn1}, {body: turn2}, {body: turn3} ]
```
2. strands-evals CloudWatchSessionMapper folds that one span into a single AgentInvocationSpan (first user prompt, last assistant response),
3. so `_session_to_span_map_result()` yields only 2 turn entries
4. the `len(turns) > 2` guard nulls it
5. conversational metrics raise "only a single turn was found."

PR #568's `_extract_from_service_format()` fallback couldn't help: it only parsed single-turn gen_ai.* events, and map_spans() only invokes it when the mapper yields no usable input/output — which never happens here, so any multi-turn logic added to it was unreachable.

**Impact:** Every multi-turn metric failed at SESSION level because no turns were extracted at all, forcing users to bypass the adapter and write custom conversational handlers.

### The Fix
Two parts, in `registry.py`:
1. Reachability: `map_spans()` now detects the collapsed multi-event shape via a new `_has_multi_event_span()` helper (any span with >1 span_events) and runs the service-format extraction to populate turns even when the mapper already produced a valid single-turn input/actual_output.
2. Correct decoding: `_extract_from_service_format()` gained a multi-turn branch that iterates `span_events[*].body`, and its `_extract_message_text()` helper now reuses CloudWatchSessionMapper's text parsing (_extract_text_from_content). This decodes the real double-encoded content (a JSON string like '[{"text": ...}]') into plain text instead of returning it literally, and selects the latest message per role, consistent with the CloudWatch single-turn path (PR #454).

Backward-compatible: the single-turn gen_ai.* events path is unchanged; the multi-turn path activates only when >1 span_event is present.

### Scope

This PR restores turn extraction (`role` + `content`) from the service-normalized SESSION format. That is sufficient for the multi-turn metrics that only need the conversation turns:

- GoalAccuracy, RoleAdherence, ConversationCompleteness, KnowledgeRetention, TopicAdherence, TurnRelevancy, ConversationalGEval

#### Out of Scope / Known Limitation (follow-up)

Metrics that need **per-turn side data** are NOT fixed by this PR, because the service-normalized `span_events[*].body` format has no slot for that data — only `input.messages` and `output.messages`:
- **retrieval_context** → TurnContextualPrecision, TurnContextualRecall, TurnFaithfulness
- **tools_called** → ToolUse

`_build_conversational_test_case()` constructs `Turn(role, content)` only, so even if the body carried these fields, they would be dropped. Fixing this requires a format convention for where per-turn `retrieval_context` / `tools_called` live in `span_events[*].body` (e.g. `body.retrieval_context`, `body.tools`), then threading them through `SpanMapResult.turns[i]` into the `Turn(...)` constructor. Tracked as a follow-up.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
+++ b/src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
@@ -1,5 +1,6 @@
 """Span mapping orchestration — uses strands-evals mappers with auto-detection."""
 
+import json
 import logging
 import warnings
 from typing import Any, Dict, List, Optional
@@ -83,7 +84,28 @@ def map_spans(
         # Mapper couldn't find AgentInvocationSpan — try service format fallback
         result = None
 
-    # Fallback: extract from service-normalized format (gen_ai events)
+    # Multi-turn reachability: when the service collapses a session into a single
+    # span with multiple span_events (one per turn), the CloudWatch mapper still
+    # produces a valid single-turn input/actual_output from the first/last event,
+    # so the `not result.input` guard below would never fire and the multi-turn
+    # turns would be lost. Detect the collapsed multi-event shape up front and run
+    # the service-format extractor so `turns` is populated even when the mapper
+    # already found a valid single-turn pair.
+    if _has_multi_event_span(session_spans):
+        service_result = _extract_from_service_format(session_spans)
+        if service_result is not None and service_result.turns:
+            if result is None:
+                result = service_result
+            else:
+                # Keep the mapper's richer fields (tools, retrieval_context, etc.)
+                # but adopt the multi-turn conversation extracted from span_events,
+                # including its input/actual_output (the latest turn) so single-turn
+                # fields stay consistent with the extracted conversation.
+                result.turns = service_result.turns
+                result.input = service_result.input
+                result.actual_output = service_result.actual_output
+
+    # Fallback: extract from service-normalized format (span_events / gen_ai events)
     if result is None or not result.input or not result.actual_output:
         service_result = _extract_from_service_format(session_spans)
         if service_result:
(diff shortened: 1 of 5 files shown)


stone-coding (commented):
Reviewed PR. The fix correctly handles the SESSION-level service-normalized format where multiple turns are collapsed into span_events[*].body. My original _extract_from_service_format() only covered single-turn gen_ai events.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
```diff
@@ -125,15 +125,63 @@ def map_spans(
...
-    events (gen_ai.user.message, gen_ai.choice) instead of body with input/output.
-    This handles that format as a fallback when strands-evals mappers can't parse it.
+    Handles two service formats:
+    1. SESSION format with span_events[*].body (multi-turn conversations where the
+       service collapses all ADOT spans into one span with multiple span_events)
+    2. gen_ai semantic convention events (single-turn Strands spans)
     """
     import json as _json
 
+    # --- Multi-turn: extract from span_events[*].body ---
```
aidandaly24, line 241: **[P0]** I think this still misses the exact one-span/many-event case this PR is trying to fix. When the CloudWatch mapper recognizes the span, it returns valid input/output from the first event, so the fallback in `map_spans()` never calls this new path. I reproduced this using real Strands `invoke_agent` bodies collapsed into the target shape: `turns` stayed `None` and DeepEval returned `FIELD_EXTRACTION_ERROR`. Could we run the service-format extraction when multiple body events are present even if the mapper found one valid pair, and retain the real scope/attributes in the regression test?

tests/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/test_error_handling.py
```diff
@@ -107,7 +107,7 @@ def test_02_unrecognized_scope(self):
         ]
         adapter = AutoEvalsAdapter(metric=_mock_scorer())
         result = adapter(_make_evaluator_input(spans=spans))
-        _assert_error_response(result, "FIELD_EXTRACTION_ERROR")
+        _assert_error_response(result, "MISSING_REQUIRED_FIELD")
```
aidandaly24: **[P0]** I do not think this expected error code matches the current behavior. The full third-party suite has `2 failed, 115 passed`: this case and the equivalent DeepEval test both still return `FIELD_EXTRACTION_ERROR`. These spans have no body or `span_events`, so the new extractor cannot reach `MISSING_REQUIRED_FIELD`. The base tests pass with the original expectation. Could we keep `FIELD_EXTRACTION_ERROR` unless the runtime behavior is intentionally changing too?

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
```diff
@@ -125,15 +125,63 @@ def map_spans(
...
+    """Extract text content from service message format.
+
+    Handles the nested structure: [{role: ..., content: {content: [{text: ...}]}}]
+    as well as the variant: [{role: ..., content: {message: [{text: ...}]}}]
+    """
+    for msg in messages:
+        content = msg.get("content", msg.get("message", {}))
+        if isinstance(content, dict):
+            # Unwrap nested content/message key
+            content = content.get("content", content.get("message", []))
```
aidandaly24: **[P0]** Could we reuse the existing CloudWatch message parsing here? Real Strands bodies encode user content as `{"content": "[{\"text\": ...}]"}`, and multi-turn input can contain interleaved user/assistant history. This helper returns the serialized JSON literally and selects the first historical message. Against three real `invoke_agent` events, it produced six turns but every user turn still contained JSON syntax. PR #454 also established the chronological-role/latest-user behavior, so sharing that parsing logic would keep this path consistent.

tests/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/autoevals/test_error_handling.py
ybdarrenwang: I reverted back to `FIELD_EXTRACTION_ERROR`. But this seems have to do with the upstream strands evals version. I just checked the latest `strands-agents-evals 1.1.1`, and it's indeed `MISSING_REQUIRED_FIELD`.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
ybdarrenwang: Added the correct fork in `map_spans` in latest revision.

src/bedrock_agentcore/evaluation/custom_code_based_evaluators/third_party/span_mappers/registry.py
ybdarrenwang: Mirrored #454 and inlined parsing logic.

PR #627, opened 2026-08-08
fix: strip reasoningContent blocks in filter_restored_tool_context

### LINKED ISSUE #621: [BUG] filter_restored_tool_context is incompatible with extended thinking
### Version

`bedrock-agentcore` 1.19.0

### Model

Claude Sonnet 4.6 via Amazon Bedrock, with extended thinking enabled (`{"type": "adaptive", "display": "summarized"}`)

### What happens

Setting `filter_restored_tool_context=True` on `AgentCoreMemoryConfig` makes every request fail once there is history to restore:

```
ValidationException: messages.1.content.6: `thinking` or `redacted_thinking` blocks in
the latest assistant message cannot be modified. These blocks must remain as they were
in the original response.
```

The option is unusable for any agent with extended thinking enabled. Turning the flag off restores normal operation.

### Expected

Either the option works alongside extended thinking, or the incompatibility is documented and the flag rejected at configuration time rather than failing on every request at runtime.

### Context

We hit this while trying to work around https://github.com/strands-agents/harness-sdk/issues/3651, where an interrupted turn leaves a session permanently unusable.

gingeekrishna:
## Summary

- `_filter_restored_tool_context()` stripped `toolUse`/`toolResult` blocks from restored session history but left `reasoningContent` (extended thinking) blocks intact.
- Bedrock rejects an assistant message where `reasoningContent` blocks are present without their companion `toolUse` blocks, raising a `ValidationException`.
- Fix: add `"reasoningContent" not in content` to the filter predicate so thinking blocks are stripped together with the tool calls they preceded.

## Root cause

When extended thinking is enabled, an assistant turn produces a message like:

```json
{
  "role": "assistant",
  "content": [
    {"reasoningContent": {"reasoningText": {"text": "…"}}},
    {"toolUse": {"toolUseId": "t1", "name": "my_tool", "input": {}}}
  ]
}
```

`reasoningContent` and `toolUse` are semantically coupled — the Bedrock API requires them to appear together. `filter_restored_tool_context` was removing `toolUse` while keeping `reasoningContent`, producing a partial message that the API refuses.

## Test plan

- [x] `test_strips_reasoning_content_alongside_tool_use` — assistant message with reasoningContent+toolUse → both stripped, only plain-text assistant message survives
- [x] `test_message_with_only_reasoning_and_tool_use_is_dropped_entirely` — message becomes empty after stripping → excluded from output
- [x] `test_text_alongside_reasoning_and_tool_use_is_preserved` — text content survives after stripping reasoning+toolUse
- [x] `test_messages_without_tool_context_are_unchanged` — plain text messages pass through unchanged (no regression)
- [x] `test_existing_tool_use_filtering_still_works` — original toolUse/toolResult behaviour preserved

```
5 passed in 0.XXs
```

Fixes #621

### DIFF
--- a/src/bedrock_agentcore/memory/integrations/strands/config.py
+++ b/src/bedrock_agentcore/memory/integrations/strands/config.py
@@ -55,8 +55,12 @@ class AgentCoreMemoryConfig(BaseModel):
             Default is None (disabled).
         context_tag: XML tag name used to wrap retrieved memory context injected into messages.
             Default is "user_context".
-        filter_restored_tool_context: When True, strip historical toolUse/toolResult blocks from
-            restored messages before loading them into Strands runtime memory. Default is False.
+        filter_restored_tool_context: When True, strip historical toolUse, toolResult, and
+            reasoningContent (extended thinking) blocks from restored messages before loading them
+            into Strands runtime memory. Default is False.
+            Note: reasoningContent blocks are always stripped together with their companion toolUse
+            blocks because Bedrock rejects assistant messages that contain reasoning blocks without
+            the tool calls that followed them.
         default_metadata: Optional default metadata key-value pairs to attach to every message event.
             Merged with any per-call metadata. Maximum 15 total keys per event (including internal keys).
             Accepts plain strings (auto-wrapped) or explicit MetadataValue dicts.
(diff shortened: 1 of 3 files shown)


gingeekrishna:
Done - addressed Copilot review: updated the filter_restored_tool_context docstring in AgentCoreMemoryConfig (commit d4f9b5d) to clarify that it strips toolUse, toolResult, and reasoningContent (extended thinking) blocks, with a note explaining that reasoning blocks must always be removed alongside the tool calls that followed them.

PR #642, opened 2026-08-11
fix(runtime): update shell session wire protocol

priyasingh2111:
*Issue #, if available:*

*Description of changes:*
Updates the shell session wire protocol to simplify the connection handshake and remove deprecated features:

- **Remove metadata frame handshake**: Connection is ready immediately after WebSocket 101 upgrade — shellId is now read from the response header instead of waiting for a STATUS frame. Eliminates the blocking \`_read_metadata_frame()\` step and the \`_pending_frames\` queue.
- **Remove \`encode_close()\` frame**: \`close()\` no longer sends a CLOSE frame — the shell detaches and stays alive for the reconnect window.
- **Simplify \`on_reconnect\` callback**: No longer passes a \`reconnected: bool\` argument — callback takes no parameters.
- **Remove \`reconnected\` and \`bytes_dropped\` attributes**: No longer needed without the metadata frame flow.
  

### DIFF
--- a/src/bedrock_agentcore/runtime/shell/config.py
+++ b/src/bedrock_agentcore/runtime/shell/config.py
@@ -6,7 +6,6 @@ from typing import Awaitable, Callable, Optional, Union
 _DEFAULT_MAX_RETRIES = 5
 _DEFAULT_BASE_DELAY = 1.0
 _DEFAULT_MAX_DELAY = 15.0
-_DEFAULT_METADATA_TIMEOUT = 10.0
 _DEFAULT_RECONNECT_WINDOW = 900.0  # ~15 min — matches server-side KARP idle timeout
 _DEFAULT_OUTER_LOOP_DELAY = 30.0  # wait between inner loop exhaustion and next outer attempt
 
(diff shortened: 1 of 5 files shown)


src/bedrock_agentcore/runtime/shell/config.py
```diff
@@ -58,4 +52,4 @@ async def on_reconnect(reconnected: bool) -> None:
     max_delay: float = _DEFAULT_MAX_DELAY
     reconnect_window: Optional[float] = _DEFAULT_RECONNECT_WINDOW
     outer_loop_delay: float = _DEFAULT_OUTER_LOOP_DELAY
-    on_reconnect: Optional[Callable[[bool], Union[Awaitable[None], None]]] = field(default=None, repr=False)
+    on_reconnect: Optional[Callable[[], Union[Awaitable[None], None]]] = field(default=None, repr=False)
```
aidandaly24, line 55: Could we update the `open_shell()` docstring in `src/bedrock_agentcore/runtime/agent_core_runtime_client.py`? It still shows `on_reconnect(reconnected: bool)` and says that exiting the context sends a CLOSE frame. The callback now takes no arguments, and exiting the context now detaches the shell.

I can also make this change after this is merged, but will reapprove if you update. Thanks.

aidandaly24 (approved):

src/bedrock_agentcore/runtime/shell/config.py
priyasingh2111: Hey @aidandaly24  — thanks for the review and for offering to handle the docstring follow-up. Since it's just a docstring change, I'll go ahead and merge this now. If you don't get to the docstring update, let me know and I'll send a small follow-up PR for it.

Issue #654, opened 2026-09-01
[FEATURE] Add a WebSearch client to bedrock_agentcore.tools and a create_web_search_target() helper


sundargthb:
**Is your feature request related to a problem? Please describe.**

`bedrock_agentcore.tools` ships first-class clients for two of the built-in tools:

- `browser_client.py` provides `BrowserClient`
- `code_interpreter_client.py` provides `CodeInterpreterClient`

Web Search has no equivalent, even though it is reachable today as a Gateway connector target. To run one search from Python a developer has to create a gateway, create a target whose `targetConfiguration.mcp.connector.source.connectorId` is `web-search` with the `WebSearch` operation enabled, attach a `GATEWAY_IAM_ROLE` credential provider, wait for the target to reach READY, and then call it over MCP with SigV4-signed requests. The comparable third-party search tools are an install and a key.

**Describe the solution you'd like**

Two additions, both following patterns that already exist in this repo.

1. `src/bedrock_agentcore/tools/web_search_client.py` with a `WebSearchClient`, shaped like the existing two clients in that package, so a search is one constructor and one call. The client owns the transport and the request signing, and keeps both as internal details that callers do not configure.

2. A `create_web_search_target()` helper on `GatewayClient`, directly analogous to the existing `create_knowledge_base_target()` at `src/bedrock_agentcore/gateway/client.py`. That method builds the `connectorId: "bedrock-knowledge-bases"` connector with `enabled: ["Retrieve"]` and a `GATEWAY_IAM_ROLE` credential provider, then delegates to `create_gateway_target_and_wait()`. The web search equivalent would build `connectorId: "web-search"` with `enabled: ["WebSearch"]` the same way. Generic target creation already works through `create_gateway_target_and_wait()`, so this is about a typed one-call helper rather than new capability.

The helper should also expose the filtering the connector supports. For reference, `aws/agentcore-cli` currently models only domain exclusion: `WebSearchTranslatorInput` in `src/cli/operations/connectors/translators.ts` has a single `excludeDomains` field and builds `parameterValues.domainFilter = { exclude: [...] }`. Domain inclusion is not exposed there.

**Describe alternatives you've considered**

[...]

PR #656, opened 2026-09-01
feat(gateway): add create_web_search_target() helper

sundargthb:
## Description

Adds `GatewayClient.create_web_search_target()`, a typed one-call helper for wiring the managed `web-search` connector to a gateway.

`GatewayClient` already has helpers for the other two connectors it supports: `create_knowledge_base_target()` (`bedrock-knowledge-bases`) and `create_agentic_retrieve_target()` (`bedrock-agentic-retrieve`). Web search has no equivalent, so callers have to hand-assemble the `targetConfiguration.mcp.connector` block and the credential provider configuration themselves. This closes that gap and follows the same shape as its two siblings.

This is typed convenience, not new capability. `create_gateway_target_and_wait()` can already create the same target; the helper removes the need to know the connector ID, the tool name, and the parameter path.

Part of #654. That issue also asks for a `WebSearchClient` under `bedrock_agentcore.tools`; this PR does not include that, since the invoke-side details are still open (see Out of scope).

## What it produces

```python
client.create_web_search_target(
    gateway_identifier="gw-123",
    exclude_domains=["example.com"],
)
```

sends:

```json
{
  "gatewayIdentifier": "gw-123",
  "name": "web-search",
  "targetConfiguration": {
    "mcp": {
      "connector": {
        "source": {"connectorId": "web-search"},
        "enabled": ["WebSearch"],
        "configurations": [
          {
            "name": "WebSearch",
            "parameterValues": {"domainFilter": {"exclude": ["example.com"]}}
          }
        ]
      }
    }
  },
  "credentialProviderConfigurations": [{"credentialProviderType": "GATEWAY_IAM_ROLE"}]
}
```

`parameterValues` is omitted entirely when `exclude_domains` is not passed or is empty, so the minimal call sends only `{"name": "WebSearch"}` as the configuration entry.

## Where the constants come from

The two values that are not inferable from this repo are the connector ID and the tool name. Both are taken from the CLI, which already ships a web-search path:

- `connectorId: 'web-search'` and the tool name `WebSearch`: [`aws/agentcore-cli` `src/cli/operations/connectors/translators.ts`](https://github.com/aws/agentcore-cli/blob/main/src/cli/operations/connectors/translators.ts)
- `parameterValues.domainFilter = { exclude: [...] }`: same file, `translateWebSearch()`

The surrounding wire shape (`enabled`, `configurations`, `GATEWAY_IAM_ROLE`) mirrors `create_knowledge_base_target()` in this file.

Note on capitalization: the `ConnectorConfiguration` shape in the service model writes its example tool names in lower camel case (`retrieve`, `webSearch`), while the shipped `create_knowledge_base_target()` here sends `Retrieve` and the CLI sends `WebSearch`. This PR follows the shipped code rather than the doc example. This is the one thing the live run could not confirm, because validation stopped at the entitlement check before reaching tool names.

## Testing

- `tests/unit/gateway/test_gateway_web_search_targets.py`, 8 tests asserting the exact wire shape, the default name, `parameterValues` omission, `**kwargs` override behavior, the default credential provider, and `wait_config` pass-through. Modeled on the existing `test_gateway_kb_targets.py`.
- `tests_integ/gateway/test_gateway_web_search_targets.py`, 3 opt-in tests marked `@pytest.mark.integration`, requiring `GATEWAY_ROLE_ARN`, creating and tearing down a real gateway and targets. Modeled on the existing `tests_integ/gateway/test_gateway_kb_targets.py`.

Full unit suite: 3412 passed, 10 skipped, 4 xpassed. `ruff check` and `ruff format --check` clean on all three files.

### What the live run showed

The `Test (gateway)` CI job runs `tests_integ/gateway` against a real account in `us-west-2`, so the first push did exercise this against a live `CreateGatewayTarget`. All three tests failed with the same error:

```
ValidationException: An error occurred (ValidationException) when calling the
CreateGatewayTarget operation: Connector integration web-search is not available
for this account.
```

The request the SDK built was accepted structurally and rejected at the account entitlement check, with the service naming `web-search` back. The rest of the gateway suite, including the KB target tests, passed in the same run.

Two things follow, and the second one is a real limit on this PR:

1. The connector ID is recognized by the service.
2. Validation stops at entitlement, so this run does **not** confirm `enabled: ["WebSearch"]` or the `domainFilter.exclude` path. Those are still only verified against the CLI's implementation.

The second push makes the integration tests `pytest.skip` on that specific message and still fail on anything else, so the job reports honestly instead of red. To get the remaining verification, the integ test account needs the connector enabled, or someone on an enabled account should run:

```
GATEWAY_ROLE_ARN=<role> pytest tests_integ/gateway/test_gateway_web_search_targets.py
```

## Out of scope

- **`WebSearchClient` in `bedrock_agentcore.tools`.** The other built-in tools have data-plane clients because they have data-plane APIs (`StartBrowserSession`, `StartCodeInterpreterSession`). Web search is reached over MCP through a gateway, so an equivalent client needs a decision on how it signs and transports requests. Tracked in #654.
- **Connector `version`.** `ConnectorSource` accepts an optional `version`; neither existing connector helper exposes it, so this one does not either. Omitting it means the service picks the latest.
- **Filters other than domain exclusion.** Only `domainFilter.exclude` is set here, because that is the only parameter with a verified path. Anything else a caller needs can go through `parameter_overrides` or by passing `targetConfiguration` directly, both of which still work.
- **Runtime parameters.** Values the agent supplies per call are not admin-configured, so they are not part of this helper.

## Checklist

- [x] Tests added
- [x] Lint and format clean
- [x] No public API removed or changed; this is additive
- [ ] Wire shape confirmed end to end on an account with the `web-search` connector enabled

## Unrelated CI failures

None of the red jobs on this PR touch gateway code:

- `Compat (evaluation)` errors with `ModuleNotFoundError: No module named 'deepeval'` and `'autoevals'` in that job's own install step. It fails the same way on #653, #626 and #625, including the two baselined against v1.21.0.
- `Compat (memory)` and `Test (memory)` fail with `ServiceQuotaExceededException: Memory count limit exceeded for account ...: Current: 150, Max allowed: 150`, plus a `DeleteMemory` timeout on a memory stuck in `UPDATING`. That is integ-account state: `Compat (memory)` reproduces it while running v1.22.0's own tests, and it also fails on #653, #626 and #625.
- `Test Python 3.13` reports `1 failed, 3411 passed`. The one failure is `tests/bedrock_agentcore/payments/integrations/langgraph/test_stage3.py::TestRetryDelay::test_zero_delay_skips_sleep`, with `Expected 'sleep' to not have been called. Called 762 times.` That test patches `...langgraph.middleware.time.sleep`, which sets the attribute on the shared `time` module object rather than a module-local alias, so any other `time.sleep` call in the same interpreter is counted against the mock while the patch is active. The full suite passes locally on this branch (`3412 passed, 10 skipped, 4 xpassed`), and the same job passes on #653, #626 and #625.
- `Test Python 3.10` has no failing step. Every step, including `Upload coverage to Codecov`, is green; the job is red only because of matrix fail-fast, annotated `The strategy configuration was canceled because "test._3_13" failed`.

`Test (gateway)`, the job that covers this change, passes.


src/bedrock_agentcore/gateway/client.py
```diff
@@ -381,6 +381,101 @@ def create_agentic_retrieve_target(
...
+        if domain_filter:
+            tool_config["parameterValues"]["domainFilter"] = domain_filter
+        if description:
+            tool_config["description"] = description
+        if parameter_overrides:
+            tool_config["parameterOverrides"] = parameter_overrides
+
+        source: Dict[str, Any] = {"connectorId": "web-search"}
+        if connector_version:
+            source["version"] = connector_version
```
tejaskash, line 454: If `include_domains` is set and `connector_version` is not, this sends `domainFilter.include` against the connector default (1.1.0), which only 1.2.0+ accepts. The service rejects it with a ValidationException. Since the point of the helper is to hide connector details, either default `source["version"]` to `"1.2.0"` when `include_domains` is set, or raise a `ValueError` here so the caller gets a clear message instead of a server-side error.

tests/unit/gateway/test_gateway_web_search_targets.py
```diff
@@ -0,0 +1,164 @@
...
+        assert config["description"] == "Search the public web"
+        assert config["parameterValues"] == {
+            "domainFilter": {
+                "include": ["allowed.example"],
+                "exclude": ["example.com", "spam.example"],
+            }
+        }
+        assert config["parameterOverrides"] == [{"path": "$.maxResults", "visible": True}]
+
+    def test_include_domains_only(self):
```
tejaskash, line 72: This asserts a request the service will reject (see comment on `client.py`). Once the version default or validation is added, this test needs to change with it.

tests/unit/gateway/test_gateway_web_search_targets.py
```diff
@@ -0,0 +1,164 @@
...
+        client.create_web_search_target(
+            gateway_identifier="gw-123",
+            include_domains=["docs.aws.amazon.com"],
+        )
+
+        call_kwargs = client.create_gateway_target_and_wait.call_args[1]
+        config = call_kwargs["targetConfiguration"]["mcp"]["connector"]["configurations"][0]
+        assert config["parameterValues"] == {"domainFilter": {"include": ["docs.aws.amazon.com"]}}
+
+    def test_connector_version_omitted_by_default(self):
```
tejaskash, line 84: `test_minimal` already asserts the full request with exact equality, so this test, `test_no_domain_filter_when_no_exclude_domains`, `test_parameter_values_always_present`, and `test_default_credential_provider` add no new coverage. Suggest dropping them. `test_empty_exclude_domains_is_omitted` is the only one in this group with a distinct input and should stay.

tests_integ/gateway/test_gateway_web_search_targets.py
```diff
@@ -0,0 +1,106 @@
...
+from bedrock_agentcore.gateway.client import GatewayClient
+
+
+@pytest.mark.integration
+class TestGatewayWebSearchTarget:
+    """Integration tests for create_web_search_target."""
+
+    @classmethod
+    def setup_class(cls):
+        cls.region = os.environ.get("BEDROCK_TEST_REGION", "us-east-1")
```
tejaskash, line 24: Every other gateway integ test defaults to `us-west-2`, and the CI workflow sets `AWS_REGION=us-west-2` without setting `BEDROCK_TEST_REGION`. This file defaults to `us-east-1`, so the gateway gets created in a different region than the rest of the suite. If the account is not entitled in us-east-1, all three tests skip on every CI run. Worth confirming which region CI should actually use here.

tests_integ/gateway/test_gateway_web_search_targets.py
```diff
@@ -0,0 +1,106 @@
...
+        cls.gateway_role_arn = os.environ.get("GATEWAY_ROLE_ARN")
+        if not cls.gateway_role_arn:
+            pytest.fail("GATEWAY_ROLE_ARN must be set")
+
+        cls.gateway_client = GatewayClient(region_name=cls.region)
+        cls.test_prefix = f"sdk-integ-ws-tgt-{int(time.time())}"
+        cls.gateway_id = None
+        cls.target_ids = []
+
+        gw = cls.gateway_client.create_gateway_and_wait(
```
tejaskash, line 34: When the connector is not available, all three tests skip but `setup_class` still creates a real gateway and `teardown_class` deletes it on every run. Consider checking availability here, before `create_gateway_and_wait`, and skipping the whole class instead.

tests_integ/gateway/test_gateway_web_search_targets.py
```diff
@@ -0,0 +1,106 @@
...
+        if cls.gateway_id:
+            try:
+                cls.gateway_client.delete_gateway_and_wait(gatewayIdentifier=cls.gateway_id)
+            except Exception as e:
+                print(f"Failed to delete gateway {cls.gateway_id}: {e}")
+
+    def _create_target(self, **kwargs):
+        """Create a web search target, skipping the test if the account is not entitled.
+
+        The web-search connector is enabled per account. When it is not, CreateGatewayTarget
```
tejaskash, line 62: This says the connector is enabled per account, but the module docstring says it is only offered in three regions. If the skip in CI is really a region mismatch, a reader will look at the wrong thing. Worth reconciling the two explanations.

tejaskash (commented):
A few comments inline. The main one is the `include_domains` default version. Also the PR description still describes an earlier revision: it says `parameterValues` is omitted when there are no exclusions (it is always sent as `{}`), lists `include_domains` and connector version as out of scope (both are implemented), says 8 tests (there are 11), and shows the default name as `web-search` (code uses `amazon-web-search`). Worth refreshing before merge.

src/bedrock_agentcore/gateway/client.py
sundargthb: Fixed. An include list now pins connector version 1.2.0 when the caller did not pin one, and if the caller pinned something older the helper raises a ValueError naming the minimum instead of building a request that cannot validate. A version string the SDK cannot parse is passed through for the service to accept or reject rather than rejected locally. Covered by `test_include_domains_pins_the_version_that_supports_it`, `test_include_domains_with_an_older_pinned_version_raises` and `test_an_unrecognized_pinned_version_is_passed_through`.

tests/unit/gateway/test_gateway_web_search_targets.py
sundargthb: Changed with it. `test_include_domains_only` is now `test_include_domains_pins_the_version_that_supports_it` and asserts the version the helper fills in, so it no longer pins a request the service would reject. There is also `test_a_two_component_version_is_not_read_as_older`, because "1.2" has to compare equal to "1.2.0" and not older, or pinning it with an include list would raise for no reason.

tests/unit/gateway/test_gateway_web_search_targets.py
sundargthb: Dropped all four. `test_empty_exclude_domains_is_omitted` stayed. The one thing `test_parameter_values_always_present` was pinning is that `parameterValues` goes on the wire even when empty, since the service drops a configuration without it and then rejects the request as empty, so that reasoning moved into the `test_minimal` docstring where the exact equality assertion already covers the behaviour.

tests_integ/gateway/test_gateway_web_search_targets.py
sundargthb: us-east-1 is the right default here. Web search is not offered in us-west-2, so taking the ambient region would skip every CI run, which defeats the point of having the test. The docstring now says that explicitly. On this commit all three tests passed in the `Test (gateway)` job against us-east-1, so the account is entitled there and nothing is skipping.

tests_integ/gateway/test_gateway_web_search_targets.py
sundargthb: The region half is in. That check now runs in `setup_class` before `create_gateway_and_wait`, so an unsupported region skips the class without creating anything.

The account half I could not do the same way. No API reports whether a connector is available to an account: `bedrock-agentcore-control` in botocore 1.43.73 has no read operation for connectors at all, so entitlement can only be discovered by calling `CreateGatewayTarget` and reading the error, which is what the skip in `_create_target` does. If there is an internal API for this that is not in the public model, point me at it and I will move the check up. Leaving this one open.

tests_integ/gateway/test_gateway_web_search_targets.py
sundargthb: Reconciled. The module docstring now states the two reasons separately, one for the region and one for the account, and says both are checked before anything is created. The docstring on `_create_target` says which of the two it is handling, so the region explanation and the account explanation are no longer competing for the same skip.

tejaskash (changes requested):
Requesting changes for the `include_domains` version issue in `client.py` (see inline comment above). Without a default to 1.2.0 or client-side validation, the helper builds a request the service rejects. The other comments are non-blocking.

tejaskash (approved):
All comments addressed in f466439. Verified the version pinning logic, the trimmed unit tests, and that all three integ tests passed in the Test (gateway) job against us-east-1. Agree the account entitlement check has to stay in _create_target since there is no API to query it.

PR #658, opened 2026-09-03
feat(tools): add WebSearchClient for invoking Amazon Web Search

sundargthb:
## Description

The SDK can already create a web search connector target on a gateway (`GatewayClient.create_web_search_target`, #656), but there is no way to *call* the resulting tool from Python. Today anything that is not already an MCP client has to hand-roll the SigV4 signed MCP handshake against the gateway endpoint to use web search.

This adds `WebSearchClient`:

```python
from bedrock_agentcore.tools import WebSearchClient

client = WebSearchClient(region="us-east-1", gateway_id="my-gateway-abc123")
for result in client.search("what is amazon bedrock agentcore", max_results=5):
    print(result.title, result.url)
```

### Design notes

**No new dependency.** The transport sits behind a `WebSearchBackend` seam. `GatewayMcpBackend` speaks only the slice of MCP streamable HTTP that one tool call needs (`initialize`, the `notifications/initialized` ack, optionally `tools/list`, then `tools/call`) using `urllib3` and `botocore.auth.SigV4Auth`, both already core dependencies. `mcp-proxy-for-aws` would do the signing for us, but it is currently dev/test only here and it is async, so it would change the SDK's dependency surface and the calling style for one capability. If reviewers would rather take that dependency as a published extra, the backend is one file to swap and the public `search()` signature does not move.

**Two response framings.** A gateway may answer a POST with `application/json` or with `text/event-stream`. Both are parsed, since which one comes back for a single `tools/call` is not something the client can assume.

**Tool name resolution.** Gateway prefixes every tool with the name of the target it came from, delimited by three underscores, so the tool the agent sees is `<targetName>___WebSearch`, not `WebSearch`. Passing `target_name` derives the name directly and skips a round trip. Otherwise `tools/list` is walked (following `nextCursor`) and the WebSearch tool is picked, with a clear error when more than one target exposes one.

**Validation before the call.** `query` is required and capped at 200 characters, `max_results` at 1 to 25, per the documented tool schema. The domain and published-date filters require connector version 1.2.0 or later on the target, which the docstring says.

**Citations are preserved.** `WebSearchResult` keeps `url`, `title` and `published_date` alongside `text`, because the acceptable use terms require source citations and links to be retained and displayed in anything surfaced to an end user. Dropping them in the response type would make compliant use harder than it needs to be.

**Endpoint safety.** `get_gateway_mcp_endpoint` validates the gateway identifier as a DNS label before interpolating it into the hostname, matching the region validation already in `_utils/endpoints.py`, so a crafted identifier cannot redirect the request off AWS. The `connection` header is excluded from signing, which otherwise produces a signature mismatch server-side.

**Region handling.** The connector is offered in us-east-1, eu-west-1 and ap-northeast-1. Calling from another region logs a warning rather than raising, so a stale constant in the SDK never blocks a call to a region that was added after the release.

## Testing

`uv run pytest tests -q` passes: 3501 passed, 10 skipped, 4 xpassed. 97 new unit tests (81 for the client, 16 for the gateway identifier validation) cover the SigV4 header set, the request sequence, session reuse, both response framings, tool-name resolution including pagination and ambiguity, the `isError` and JSON-RPC error paths, ARN parsing, and the gateway identifier validation. `ruff check` and `ruff format --check` are clean. Statement coverage of the new module is 100%.

**Not verified against a live service.** The web search connector is enabled per account and this repo's integration test account is not entitled to it, so the invoke path has been exercised only against mocked HTTP responses. The request shapes come from the public tool schema and response format documentation, and the MCP sequence from the streamable HTTP transport spec, but nothing here has completed a real search. `tests_integ/tools/test_web_search_client.py` is included and will exercise the real path once run against an entitled account with `WEB_SEARCH_GATEWAY_ID` set; it skips rather than fails without one. Reviewers with an entitled account running that file would be the useful confirmation.

## Confirmed with the service team

The API surface this is built on has since been confirmed by the Web Search Tool team, and matches what is implemented here: SigV4 (AWS_IAM) with no search-specific API key or credential; availability in us-east-1, eu-west-1 and ap-northeast-1; access scoped through IAM policies on the gateway ARN rather than any key; and both `domainFilter` (include and exclude, 100 domains per list) and `publishedDateFilter` (`from`/`to`, ISO-8601 UTC, inclusive) at target and request level from connector version 1.2.0.

Two clarifications from that exchange are now in the docstrings: the caller's credentials need `bedrock-agentcore:InvokeGateway` on the gateway ARN while the gateway service role needs `bedrock-agentcore:InvokeWebSearch` on the connector, and request-level include lists intersect with the target-level include list rather than replacing it, so disjoint lists return no results with no error raised.

## Checklist

- [x] Unit tests added
- [x] Integration tests added (skip without an entitled account and a gateway)
- [x] Lint and format pass
- [x] No new dependencies

### DIFF
--- a/src/bedrock_agentcore/tools/__init__.py
+++ b/src/bedrock_agentcore/tools/__init__.py
@@ -26,6 +26,13 @@ from .config import (
     VpcConfig,
     create_browser_config,
 )
+from .web_search_client import (
+    WebSearchBackend,
+    WebSearchClient,
+    WebSearchError,
+    WebSearchResponse,
+    WebSearchResult,
+)
 
 __all__ = [
     "BasicAuth",
@@ -53,5 +60,10 @@ __all__ = [
     "SessionConfiguration",
     "ViewportConfiguration",
     "VpcConfig",
+    "WebSearchBackend",
+    "WebSearchClient",
+    "WebSearchError",
+    "WebSearchResponse",
+    "WebSearchResult",
     "create_browser_config",
 ]
--- a/src/bedrock_agentcore/tools/web_search_client.py
+++ b/src/bedrock_agentcore/tools/web_search_client.py
@@ -1,0 +1,648 @@
+"""Client for AgentCore Web Search.
+
+Web Search is reachable today as an AgentCore Gateway connector target, which the
+agent calls as an MCP tool. This module wraps that so callers get a plain
+``search()`` method and a normalized result type instead of MCP content blocks:
+
+    >>> from bedrock_agentcore.tools import WebSearchClient
+    >>>
+    >>> client = WebSearchClient(region="us-east-1", gateway_id="my-gateway-abc123")
+    >>> response = client.search("what shipped in python 3.13", max_results=5)
+    >>> for result in response.results:
+    ...     print(result.title, result.url)
+
+The transport lives behind :class:`WebSearchBackend` so the same ``search()``
+signature and the same :class:`WebSearchResult` can be served by a different
+backend later without changing callers.
+"""
+
+import json
+import logging
+import threading
+from dataclasses import dataclass, field
+from typing import Any, Dict, Iterator, List, Optional, Sequence
+
+import urllib3
+from botocore.auth import SigV4Auth
+from botocore.awsrequest import AWSRequest
+
+from bedrock_agentcore._utils.endpoints import get_gateway_mcp_endpoint
+from bedrock_agentcore._utils.user_agent import SDK_VERSION, build_user_agent_suffix
+
+logger = logging.getLogger(__name__)
+
+#: Name of the MCP tool the web search connector exposes. Fixed by the service.
+WEB_SEARCH_TOOL_NAME = "WebSearch"
+
+#: Gateway prefixes every tool it exposes with the name of the target it came from.
+GATEWAY_TOOL_NAME_DELIMITER = "___"
+
+#: Default target name used by ``GatewayClient.create_web_search_target``.
+DEFAULT_TARGET_NAME = "amazon-web-search"
+
+#: Service name the gateway data plane signs as.
+GATEWAY_SIGNING_SERVICE = "bedrock-agentcore"
+
+#: Documented input limits for the WebSearch tool.
+MAX_QUERY_LENGTH = 200
+MIN_MAX_RESULTS = 1
+MAX_MAX_RESULTS = 25
+
+#: Regions where the web search connector is offered. Used for a warning only, never
+#: to block a call, so that a newly added region does not require an SDK release.
+KNOWN_REGIONS = ("us-east-1", "eu-west-1", "ap-northeast-1")
+
+_MCP_PROTOCOL_VERSION = "2025-06-18"
+_DEFAULT_TIMEOUT = 30
+
+
+class WebSearchError(RuntimeError):
+    """Raised when a web search call fails."""
+
+
+@dataclass(frozen=True)
+class WebSearchResult:
+    """A single web search result.
+
+    Attributes:
+        text: The extracted snippet relevant to the query. Always present.
+        url: URL of the source page.
+        title: Title of the source page.
+        published_date: Publication date of the page, as reported by the index.
+    """
+
+    text: str
+    url: Optional[str] = None
+    title: Optional[str] = None
+    published_date: Optional[str] = None
+
+    @classmethod
+    def from_payload(cls, payload: Dict[str, Any]) -> "WebSearchResult":
+        """Build a result from one entry of a search response."""
+        return cls(
+            text=payload.get("text") or "",
+            url=payload.get("url"),
+            title=payload.get("title"),
+            published_date=payload.get("publishedDate"),
+        )
+
+
+@dataclass(frozen=True)
+class WebSearchResponse:
+    """The results of one web search.
+
+    Attributes:
+        results: The results, in the order the service returned them.
+        search_id: Service-assigned identifier for the search, when present.
+    """
+
+    results: List[WebSearchResult] = field(default_factory=list)
+    search_id: Optional[str] = None
+
+    def __len__(self) -> int:
+        """Number of results."""
+        return len(self.results)
+
+    def __iter__(self) -> Iterator[WebSearchResult]:
+        """Iterate over the results."""
+        return iter(self.results)
+
+    @classmethod
+    def from_payload(cls, payload: Dict[str, Any]) -> "WebSearchResponse":
+        """Build a response from the decoded search payload."""
+        raw_results = payload.get("results") or []
+        return cls(
+            results=[WebSearchResult.from_payload(item) for item in raw_results if isinstance(item, dict)],
+            search_id=payload.get("id"),
+        )
+
+
+def _build_arguments(
+    query: str,
+    max_results: Optional[int] = None,
+    include_domains: Optional[Sequence[str]] = None,
+    exclude_domains: Optional[Sequence[str]] = None,
+    published_after: Optional[str] = None,
+    published_before: Optional[str] = None,
+) -> Dict[str, Any]:
+    """Validate search inputs and shape them into the tool's argument object.
+
+    Raises:
+        ValueError: If the query is empty or over the documented length limit, or
+            if max_results falls outside the documented range.
+    """
+    if not query or not query.strip():
+        raise ValueError("query must be a non-empty string")
+    if len(query) > MAX_QUERY_LENGTH:
+        raise ValueError(f"query must be {MAX_QUERY_LENGTH} characters or fewer, got {len(query)}")
+
+    arguments: Dict[str, Any] = {"query": query}
+
+    if max_results is not None:
+        if not isinstance(max_results, int) or isinstance(max_results, bool):
+            raise ValueError(f"max_results must be an integer, got {type(max_results).__name__}")
+        if not MIN_MAX_RESULTS <= max_results <= MAX_MAX_RESULTS:
+            raise ValueError(f"max_results must be between {MIN_MAX_RESULTS} and {MAX_MAX_RESULTS}, got {max_results}")
+        arguments["maxResults"] = max_results
+
+    filters: Dict[str, Any] = {}
+    domain_filter: Dict[str, List[str]] = {}
+    if include_domains:
+        domain_filter["include"] = list(include_domains)
+    if exclude_domains:
+        domain_filter["exclude"] = list(exclude_domains)
+    if domain_filter:
+        filters["domainFilter"] = domain_filter
+
+    published_filter: Dict[str, str] = {}
+    if published_after:
+        published_filter["from"] = published_after
+    if published_before:
+        published_filter["to"] = published_before
+    if published_filter:
+        filters["publishedDateFilter"] = published_filter
+
+    if filters:
+        arguments["filters"] = filters
+
+    return arguments
+
+
+def _extract_search_payload(result: Dict[str, Any]) -> Dict[str, Any]:
+    """Pull the search payload out of an MCP ``tools/call`` result.
+
+    The connector returns the results as a JSON document inside a text content
+    block, so the text has to be decoded rather than read directly.
+
+    Raises:
+        WebSearchError: If the tool reported an error or returned no decodable
+            text content.
+    """
+    if result.get("isError"):
+        raise WebSearchError(f"Web search tool reported an error: {_first_text(result) or result}")
+
+    text = _first_text(result)
+    if text is None:
+        raise WebSearchError(f"Web search response contained no text content: {result}")
+
+    try:
+        payload = json.loads(text)
+    except json.JSONDecodeError as exc:
+        raise WebSearchError(f"Could not decode web search response as JSON: {text[:200]!r}") from exc
+
+    if not isinstance(payload, dict):
+        raise WebSearchError(f"Expected a JSON object in the web search response, got {type(payload).__name__}")
+    return payload
+
+
+def _first_text(result: Dict[str, Any]) -> Optional[str]:
+    """Return the first text content block of an MCP result, if any."""
+    for block in result.get("content") or []:
+        if isinstance(block, dict) and block.get("type") == "text" and isinstance(block.get("text"), str):
+            return block["text"]
+    return None
+
+
+class WebSearchBackend:
+    """How a :class:`WebSearchClient` reaches web search.
+
+    A backend takes the tool's argument object and returns the decoded search
+    payload, meaning a dict shaped ``{"id": ..., "results": [...]}``. Everything
+    above this line is transport independent.
+    """
+
+    def search(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
+        """Run one search and return the decoded payload."""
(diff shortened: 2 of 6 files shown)


src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+            if not line.startswith("data:"):
+                continue
+            chunk = line[len("data:") :].strip()
+            if not chunk:
+                continue
+            try:
+                message = json.loads(chunk)
+            except json.JSONDecodeError:
+                continue
+            if isinstance(message, dict) and "jsonrpc" in message:
```
tejaskash, line 459: This returns the first JSON-RPC message on the stream, not the response. If the gateway sends a notification (e.g. `notifications/progress`) before the `tools/call` result, the notification is returned and `search()` fails with "contained no text content".

`_post` knows the outgoing `id`, so matching on it fixes this. At minimum, require a `result` or `error` key.

src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+        """Send one JSON-RPC message and return the decoded reply, if there is one."""
+        body = json.dumps(message).encode("utf-8")
+
+        extra: Dict[str, str] = {}
+        if self._mcp_session_id:
+            extra["Mcp-Session-Id"] = self._mcp_session_id
+        if self._initialized:
+            extra["MCP-Protocol-Version"] = self._protocol_version
+
+        response = self._http.request(
```
tejaskash, line 314: Connection refused, DNS failure, TLS errors and read timeouts escape here as `urllib3` exceptions (`MaxRetryError`, `ReadTimeoutError`, `ProtocolError`). The `search()` docstring promises `WebSearchError`, so callers who follow it will miss the most common failure for a 30s HTTP call.

Suggest wrapping `urllib3.exceptions.HTTPError` into `WebSearchError` with `from exc`.

src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+            headers=self._signed_headers(body, extra),
+            timeout=urllib3.Timeout(total=self._timeout),
+            preload_content=True,
+        )
+
+        session_id = response.headers.get("Mcp-Session-Id")
+        if session_id:
+            self._mcp_session_id = session_id
+
+        if response.status >= 400:
```
tejaskash, line 327: If the gateway expires the session and returns 404 (the spec's re-initialize signal), this raises but leaves `_initialized=True` and `_mcp_session_id` set. Every later `search()` on this instance fails the same way. Only `close()` recovers, and that is not documented.

On 404 with a session id in flight, reset `_initialized` and `_mcp_session_id` before raising so the next call re-initializes.

src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+            }
+        )
+        if reply is None:
+            raise WebSearchError("Gateway did not answer the MCP initialize request")
+
+        negotiated = (reply.get("result") or {}).get("protocolVersion")
+        if isinstance(negotiated, str) and negotiated:
+            self._protocol_version = negotiated
+
+        self._initialized = True
```
tejaskash, line 367: Set after the notification POST on the next line succeeds. If that POST raises, the flag stays `True` and the next `search()` skips the handshake.

src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+def _decode_jsonrpc(content_type: str, data: bytes) -> Optional[Dict[str, Any]]:
+    """Decode a JSON-RPC reply from either a JSON body or an SSE stream.
+
+    Returns None when the body carries no JSON-RPC message, which is what a
+    notification acknowledgement looks like.
+    """
+    text = data.decode("utf-8", "replace")
+
+    if "text/event-stream" in content_type.lower():
+        for line in text.splitlines():
```
tejaskash, line 449: Per the SSE spec, consecutive `data:` lines in one event are joined with `\n`. Parsing each line alone means a message split across two lines is dropped and reported as "Gateway did not answer". Unlikely in practice but the parser claims to handle this framing.

src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+        if self.region not in KNOWN_REGIONS:
+            logger.warning(
+                "Web search is offered in %s. Calling %s may fail if the connector is not available there.",
+                ", ".join(KNOWN_REGIONS),
+                self.region,
+            )
+
+        if gateway_id:
+            gateway_endpoint = get_gateway_mcp_endpoint(gateway_id, self.region)
+        if not gateway_endpoint:
```
tejaskash, line 573: `gateway_id` goes through `_validate_endpoint_url` via `get_gateway_mcp_endpoint`, but a user-supplied `gateway_endpoint` is passed to the backend unchecked. The rest of `_utils/endpoints.py` validates even the env override. Signed requests (including `X-Amz-Security-Token`) then go to whatever host was given. One-line check for consistency.

src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+logger = logging.getLogger(__name__)
+
+#: Name of the MCP tool the web search connector exposes. Fixed by the service.
+WEB_SEARCH_TOOL_NAME = "WebSearch"
+
+#: Gateway prefixes every tool it exposes with the name of the target it came from.
+GATEWAY_TOOL_NAME_DELIMITER = "___"
+
+#: Default target name used by ``GatewayClient.create_web_search_target``.
+DEFAULT_TARGET_NAME = "amazon-web-search"
```
tejaskash, line 41: Defined and never referenced. `_ensure_tool_name` falls through to `tools/list` when `target_name` is absent instead of trying this default. Use it or remove it.

src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+            boto3_session: Session to take credentials and region from.
+            timeout: Per-request timeout in seconds.
+            integration_source: Optional framework identifier for the User-Agent.
+
+        Raises:
+            ValueError: If no gateway is identified, or more than one is.
+        """
+        import boto3
+
+        self._session = boto3_session or boto3.Session()
```
tejaskash, line 535: When `backend` is supplied this session is only used to read `region_name`, but it still does the full credential provider chain setup. Harmless, just wasted work.

src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+        so when the target already has an include list, passing ``include_domains``
+        narrows to the intersection of the two. If the two share no domains the
+        search returns nothing, which is a silent empty result rather than an
+        error, so check the target's configuration when a filtered search comes
+        back empty.
+
+        Args:
+            query: What to search for. 200 characters or fewer.
+            max_results: How many results to return, 1 to 25. Service default is 10.
+            include_domains: Restrict results to these domains. Up to 100. A root
```
tejaskash, line 613: "Up to 100" is documented for both domain lists but not enforced in `_build_arguments`, while `query` length and `max_results` are. Not wrong, just inconsistent.

src/bedrock_agentcore/tools/web_search_client.py
```diff
@@ -0,0 +1,666 @@
...
+    def __enter__(self) -> "WebSearchClient":
+        """Enter the context manager."""
+        return self
+
+    def __exit__(self, *exc_info: Any) -> None:
+        """Close the client on exit."""
+        self.close()
+
+
+def _parse_gateway_arn(arn: str) -> tuple:
```
tejaskash, line 652: `Tuple[str, str]` to match `browser_client.py`.

src/bedrock_agentcore/tools/__init__.py
```diff
@@ -26,6 +26,13 @@
     VpcConfig,
     create_browser_config,
 )
+from .web_search_client import (
+    WebSearchBackend,
```
tejaskash, line 30: `GatewayMcpBackend` is not exported though `WebSearchBackend` is. Anyone building the concrete backend directly (e.g. to share one across clients) has to import from the module path.

tejaskash (commented):
Reviewed the transport layer against the MCP streamable HTTP spec. Tests pass locally. Inline comments below, ordered roughly by severity. The first three matter once this runs against a real gateway.

tests/bedrock_agentcore/tools/test_web_search_client.py
```diff
@@ -0,0 +1,693 @@
...
+
+        headers = backend._http.request.call_args_list[2].kwargs["headers"]
+        assert headers["Authorization"].startswith("AWS4-HMAC-SHA256 Credential=AKIAEXAMPLE/")
+        assert "/us-east-1/bedrock-agentcore/aws4_request" in headers["Authorization"]
+        assert "X-Amz-Date" in headers
+        assert headers["Content-Type"] == "application/json"
+        assert headers["Accept"] == "application/json, text/event-stream"
+        assert headers["Content-Length"] == str(len(backend._http.request.call_args_list[2].kwargs["body"]))
+
+    def test_connection_header_is_never_signed(self):
```
tejaskash, line 270: This tests botocore, not this PR. `connection` is in botocore's `SIGNED_HEADERS_BLACKLIST` and no `connection` header is ever built here, so the test cannot fail. Suggest removing it.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Fixed by matching on the id. `_post` passes the outgoing id into `_decode_jsonrpc`, which skips any message that is not the reply to it, so a notification arriving first is ignored rather than returned. An error is still surfaced whatever id it carries, since JSON-RPC allows an error response with a null id. A reply that does carry our id but has neither a result nor an error now raises instead of being handed back. Tests: `test_a_reply_to_another_id_is_not_the_answer`, `test_an_error_is_returned_whatever_id_it_carries`, `test_event_stream_skips_a_notification_ahead_of_the_reply` and `test_reply_without_result_or_error_is_surfaced`.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Wrapped. `urllib3.exceptions.HTTPError` is caught and re-raised as a `WebSearchError` with `from exc`, naming the endpoint, so everything the docstring promises holds for connection, DNS, TLS and timeout failures. Covered by `test_transport_failure_becomes_a_web_search_error`.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Fixed. A 404 while a session id is held clears both `_initialized` and `_mcp_session_id` before raising, so the next call performs the handshake again instead of failing the same way until someone calls `close()`. `test_expired_session_is_dropped_and_the_next_search_reinitializes` asserts the state is cleared and that the following search initializes a second time and succeeds.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Fixed. The parser accumulates the `data` lines of an event and joins them with a newline, dispatches on a blank line, strips exactly one leading space after the colon, and ignores comment, `id` and `retry` lines. Tests cover a message split across two `data` lines, a final event with no trailing blank line, and an undecodable event followed by a good one.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Added. A supplied `gateway_endpoint` now goes through `_validate_endpoint_url`, the same check an endpoint built from a gateway id gets. `test_non_aws_gateway_endpoint_is_rejected` asserts a non-AWS host is refused before anything is signed.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Removed. Resolving the tool name through `tools/list` is the behaviour I want when no target name is given, since it works against a target created under any name, so the constant had no reader. The default target name now lives in the gateway helper in #656, where it is the name the target is actually created with.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Made lazy. The session is created on first use in `_resolve_session`, so passing a backend together with a region creates none at all. `test_a_backend_and_a_region_need_no_session` patches `boto3.Session` and asserts it is never called, and there are two more tests for the paths that do need one.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Enforced now, for both lists, in the same place as the query length and `max_results` checks. `test_domain_list_maximum` is parametrized over the two arguments.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Changed to `Tuple[str, str]`.

src/bedrock_agentcore/tools/__init__.py
sundargthb: Exported. `GatewayMcpBackend` is now imported in `bedrock_agentcore.tools` and listed in `__all__` next to `WebSearchBackend` and `WebSearchClient`.

tests/bedrock_agentcore/tools/test_web_search_client.py
sundargthb: Removed. You are right that it asserts botocore's own behaviour and cannot fail from anything in this PR.

src/bedrock_agentcore/tools/web_search_client.py
sundargthb: Done the other way round, because the notification itself has to carry `MCP-Protocol-Version` and that header is only sent once `_initialized` is true. So it is set before the POST and rolled back if the POST raises, which leaves the instance uninitialized with the session id cleared. `test_failed_notification_leaves_the_handshake_unset` covers it.

tejaskash (changes requested):
Requesting changes for the first three items in my inline review above (https://github.com/aws/bedrock-agentcore-sdk-python/pull/658#pullrequestreview-5118168164):

1. SSE decoder returns the first JSON-RPC message rather than the response matching the request id (line 459).
2. urllib3 transport exceptions escape `_post` despite the `WebSearchError` contract in the docstring (line 314).
3. An expired session (HTTP 404) leaves the backend permanently failing until `close()` (line 327).

The remaining comments are non-blocking.

tejaskash (approved):
All 12 items addressed at 0ed088e. Verified locally: 171 tests pass, and I exercised `_decode_jsonrpc` directly with a notification ahead of the reply, a reply to another id, and a message split across two `data:` lines. The reordered `_initialized` handling with rollback is the right call given the protocol version header dependency. Approving.

PR #664, opened 2026-09-16
fix(memory): use current Strands bidi session hooks

pgrayy:
*Issue #, if available:*

N/A

*Description of changes:*

Importing `AgentCoreMemorySessionManager` fails with Strands 1.56.0 because it references bidi hook events removed in [strands-agents/harness-sdk#4280](https://github.com/strands-agents/harness-sdk/pull/4280). Use the shared initialization and message hooks for both `Agent` and `BidiAgent`, and use `BidiAgentStopEvent` to save state when a bidi session stops. Keep initialization synchronous and offload persistence callbacks when `async_mode=True`.

Flush pending message and state batches after the final state sync for both `AfterInvocationEvent` and `BidiAgentStopEvent`. Skip automatic context retrieval for `BidiAgent`, because modifying local message history does not deliver that context to the live model.

Update the agent annotations to `LocalAgent`, raise the minimum `strands-agents` version to 1.56.0, and add regression coverage for callback ordering and bidi persistence without retrieval in both callback modes.

Validation:
- All pre-commit checks and the Bandit security scan passed.
- The full test suite with coverage passed in the locked Python 3.10 development environment (91% coverage). The focused session-manager and configuration suites passed (200 tests).
- Before the review fixes, five existing live AWS integration tests passed, covering initialization, conversation persistence, batching, and session restoration.
- Before the review fixes, two one-off bidi checks passed against live AgentCore storage with a scripted model and the default batch size of 1. These exercised public `start`, `send`, `receive`, and `stop` calls in both callback modes, including message restoration and state saved on stop.


jariy17 (dismissed):

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -990,21 +985,10 @@ async def _on_message_added_persist(event: MessageAddedEvent) -> None:
...
-        # invoke_callbacks_async, so async wrappers are safe.
-        registry.add_callback(BidiAgentInitializedEvent, lambda event: self.initialize_bidi_agent(event.agent))
-
-        async def _on_bidi_message_added(event: BidiMessageAddedEvent) -> None:
-            await asyncio.to_thread(self.append_bidi_message, event.message, event.agent)
-            await asyncio.to_thread(self.sync_bidi_agent, event.agent)
-
-        registry.add_callback(BidiMessageAddedEvent, _on_bidi_message_added)
-        registry.add_callback(BidiAfterInvocationEvent, _offload(self.sync_bidi_agent, lambda e: e.agent))
+        registry.add_callback(BidiAgentStopEvent, _offload(self.sync_agent, lambda e: e.agent))
```
notgitika, line 994: I might be missing something here, but with batching enabled, doesn’t `sync_agent()` just add the state to the buffer? 

since Bidi doesn’t fire `AfterInvocationEvent`, what flushes the pending messages and state after agent.stop()?

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
```diff
@@ -990,21 +985,10 @@ async def _on_message_added_persist(event: MessageAddedEvent) -> None:
...
-        # BidiAgentInitializedEvent dispatches through invoke_callbacks (sync),
-        # so its callback must stay sync; the other two dispatch through
-        # invoke_callbacks_async, so async wrappers are safe.
-        registry.add_callback(BidiAgentInitializedEvent, lambda event: self.initialize_bidi_agent(event.agent))
-
-        async def _on_bidi_message_added(event: BidiMessageAddedEvent) -> None:
-            await asyncio.to_thread(self.append_bidi_message, event.message, event.agent)
-            await asyncio.to_thread(self.sync_bidi_agent, event.agent)
-
-        registry.add_callback(BidiMessageAddedEvent, _on_bidi_message_added)
```
notgitika, line 1003: Now that bidi uses `MessageAddedEvent`, will `retrieve_customer_context` run for bidi too? It looks like it updates `agent.messages`, while the model still receives the original input event. could we skip retrieval for bidi or inject it into the outgoing event?

notgitika (commented):
Thanks for the fix, added 2 comments

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
pgrayy: Added a flush and also adjusted the ordering for AfterInvocationEvent. The inline comment explains under the batch_size > 1 blocks.

src/bedrock_agentcore/memory/integrations/strands/session_manager.py
pgrayy: Added a skip. This also highlights an important behavioral difference. It may actually make sense to go back to BidiMessageAddedEvent. I'll give this some thought. If I do update, I'll be sure to update here as well.

Another approach could be that Bidi just doesn't emit the event at all. And this is also something to consider for model providers that allow server side conversation management (e.g., OpenAI responses).

notgitika: Makes sense!

notgitika (approved):
LGTM

notgitika:
[Breaking Change Check / Detect Breaking Changes](https://github.com/aws/bedrock-agentcore-sdk-python/actions/runs/35104782829/job/104832746868?pr=664)
Failing due to a 403 and not a breaking change.

PR #667, opened 2026-09-16
fix(integ): adjust timeout on memory test. 

Hweinstock:
### Problem 
Tests are timing out on delete https://github.com/aws/bedrock-agentcore-sdk-python/actions/runs/35110378662/job/104856020087. 

Also noticed a race condition where update does not wait for the memory to be active before continuing. 

### Solution 
- extend timeout to 300s. 
- add wait_for_active. 

### Verification 
ran tests in dev account.


Hweinstock:
> Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

integ test failures unrelated, doesn't allow run from fork, but ran memory in dev account.

nborges-aws (approved):

PR #672, opened 2026-09-22
feat(tools): add AgentCore Web Search as a Strands Agents tool

sundargthb:
## Description

Adds `AgentCoreWebSearch`, exposing the AgentCore Web Search tool as a Strands Agents tool. Placement follows `memory/integrations/strands`, `payments/integrations/strands` and `gateway/integrations/strands`: a subpackage behind the existing `strands-agents` extra, reached only by explicit import. No `pyproject.toml` change, and a base install is unaffected.

```python
from strands import Agent

from bedrock_agentcore.tools.integrations.strands import AgentCoreWebSearch

with AgentCoreWebSearch(region="us-east-1", gateway_id="my-gateway-abc123") as search:
    agent = Agent(tools=[search.web_search])
    agent("What changed in the most recent boto3 release?")
```

**Draft for one reason: the tests in `tests_integ/` have not been run.** I don't currently have credentials for an account entitled to the web search connector, so I'm not claiming this is verified against a live gateway. Marking ready once they've run.

## Two behaviours worth a look in review

**`region` is defaulted only when no region and no gateway ARN were given.** `WebSearchClient` does `region = region or arn_region`, so filling a default in here unconditionally would silently send a `eu-west-1` gateway's traffic to `us-east-1`.

**A failed search raises rather than returning the message as a result.** The Strands executor converts the exception to a `status: "error"` tool result, so the agent keeps running and the model can react. Returning `"Web search failed: ..."` as a *successful* result is indistinguishable to the model from a search that found nothing.

I checked both by mutating the source and confirming exactly one test failed each time, rather than only that the suite passes.

One forward-compatibility note: which transport is used stays inside `WebSearchClient`. Every gateway argument here is optional and only the ones supplied are forwarded, so the direct web search API will work through the same call without a signature change in this module.

## Testing

```
$ uv run pytest tests/bedrock_agentcore/tools/integrations -q --cov=src/bedrock_agentcore/tools/integrations --cov-report=term-missing
24 passed
src/bedrock_agentcore/tools/integrations/strands/web_search.py   69   0   24   0   100%

$ uv run pytest tests/ -q
3568 passed, 10 skipped, 4 xpassed

$ uv run ruff check src tests tests_integ
All checks passed!

$ uv run ruff format --check src tests tests_integ
341 files already formatted

$ uv run mypy src/bedrock_agentcore/tools/integrations/
Success: no issues found in 3 source files
```

`tests_integ/tools/integrations/strands/test_web_search_integration.py` is **not** in that run. It needs `WEB_SEARCH_GATEWAY_ID` pointing at a gateway with a web search target, on an account entitled to the connector.

## Related

- The `WebSearchClient` this wraps, and its `tests_integ/tools/test_web_search_client.py`
- langchain-ai/langchain-aws#1275, the equivalent LangChain toolkit


sundargthb:
Superseded by #673, same commit pushed as a branch on this repo so the integration test jobs can actually run. The fork variant could not run them: `integration-testing.yml` triggers on `pull_request_target` and `actions/checkout@v6` refuses to check out fork code in that context, so all ten `Test (<group>)` jobs failed at the checkout step in under 11 seconds.

